PR #3, opened 2023-06-08
Fix markdown links

jeffweiss:
Prior to this commit several directory links in DEVELOPER.md were incorrectly formatted to render as links. This commit removes the errant `)`s to resolve this.


towynlin:
Thank you @jeffweiss — sorry so slow in merging!

PR #7, opened 2023-10-12
loosen guardrails and support no alignment

russelldeguzman:
This PR addresses a few things:
* Loosen the sampling guardrails so folks can experiment with the sampling behavior at the lower end. 
* Fix the units of the default parameters in BridgePowerController
* Support a zero-alignment configuration setting

### DIFF
--- a/src/apps/bridge/bridgePowerController.h
+++ b/src/apps/bridge/bridgePowerController.h
@@ -75,6 +75,7 @@ private:
   bool _rtcSet;
   bool _initDone;
   bool _subSamplingEnabled;
+  bool _configError;
   adin2111_DeviceHandle_t _adin_handle;
   EventGroupHandle_t _busPowerEventGroup;
   TaskHandle_t _task_handle;
--- a/src/apps/bridge/bridgePowerController.h
+++ b/src/apps/bridge/bridgePowerController.h
@@ -12,10 +12,10 @@ class BridgePowerController {
 public:
   explicit BridgePowerController(
       IOPinHandle_t &BusPowerPin,
-      uint32_t sampleIntervalMs = DEFAULT_SAMPLE_INTERVAL_S,
-      uint32_t sampleDurationMs = DEFAULT_SAMPLE_DURATION_S,
-      uint32_t subsampleIntervalMs = DEFAULT_SUBSAMPLE_INTERVAL_S,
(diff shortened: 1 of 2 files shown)


src/apps/bridge/bridgePowerController.h
```diff
@@ -45,16 +45,15 @@ class BridgePowerController {
...
-  static constexpr uint32_t MIN_SAMPLE_INTERVAL_S = (5 * 60);
-  static constexpr uint32_t MIN_SAMPLE_DURATION_S = (60);
+  static constexpr uint32_t MIN_SAMPLE_DURATION_S = (1);
+  static constexpr uint32_t MIN_SAMPLE_INTERVAL_S = (1);
   static constexpr uint32_t MAX_SAMPLE_INTERVAL_S = (24 * 60 * 60);
   static constexpr uint32_t MAX_SAMPLE_DURATION_S = (24 * 60 * 60);
-  static constexpr uint32_t MIN_SUBSAMPLE_INTERVAL_S = (12);
-  static constexpr uint32_t MIN_SUBSAMPLE_DURATION_S = (6);
+  static constexpr uint32_t MIN_SUBSAMPLE_INTERVAL_S = (1);
+  static constexpr uint32_t MIN_SUBSAMPLE_DURATION_S = (1);
```
victorsowa12: I believe I made the min sample/subsample durations 6s to make sure whenever the bus is turned on we have sufficient time do a topology sample. Have you tested that we are still able to get the topology with it down at 1?

victorsowa12: oh yeah in `topology_sampler_task` I wrote:
```
topology_sample();
```

The sample itself shouldn't take too long, at the time I was concerned about the devices booting up

russelldeguzman: Ah I see, we can make these all 6 and add a comment.

victorsowa12: Sounds good!

russelldeguzman: Our status quo for configs: we've been recommending settings, but giving folks flexibility to play with extremes. I don't expect people to set to the minimum, but if they do -- it is a devkit after all

victorsowa12 (approved):

towynlin (approved):

PR #9, opened 2023-10-13
Russ/use sys info msg

russelldeguzman:
Switch to bm_common message sys info.

### DIFF
--- a/src/lib/middleware/services/sys_info_service.cpp
+++ b/src/lib/middleware/services/sys_info_service.cpp
@@ -79,6 +79,7 @@ static bool sys_info_service_handler(size_t service_strlen, const char *service,
     d.node_id = getNodeId();
     d.sys_config_crc = _sys_config->getCRC32();
     size_t encoded_len;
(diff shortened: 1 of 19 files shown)


src/lib/middleware/services/sys_info_service.cpp
```diff
@@ -62,8 +63,6 @@ static bool sys_info_service_handler(size_t service_strlen, const char *service,
                                    uint8_t *reply_data) {
   (void)(req_data);
   bool rval = false;
-  size_t max_data_len = buffer_len;
-  buffer_len = 0;
```
towynlin, line 66: I had to read carefully around to understand that the message cpp file handles checking the buffer size. If the caller of this function passes a buffer that's too small, then this function will return false because encode will return CborErrorOutOfMemory. Yeah?

towynlin (approved):

src/lib/middleware/services/sys_info_service.cpp
russelldeguzman: Yes, exactly! Can add a comment for clarity

PR #11, opened 2023-10-16
Sc 194765/sensor watchdog

russelldeguzman:
Adding a generic sensor watchdog to the Aanderaa app.

Testing:
- Connected an Aanderaa sensor to confirm normal working. 
- Disconnected the sensor and eventually the watchdog kicks in
Screen Shot 2023-10-16 at 4 48 21 PM

### DIFF
--- a/src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
+++ b/src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
@@ -1,29 +1,38 @@
-#include "avgSampler.h"
-#include "debug.h"
-#include "util.h"
 #include "user_code.h"
+#include "OrderedKVPLineParser.h"
+#include "array_utils.h"
+#include "avgSampler.h"
+#include "bm_network.h"
+#include "bm_printf.h"
+#include "bm_pubsub.h"
 #include "bsp.h"
+#include "debug.h"
+#include "lwip/inet.h"
+#include "payload_uart.h"
+#include "sensorWatchdog.h"
+#include "sensors.h"
 #include "stm32_rtc.h"
 #include "task_priorities.h"
 #include "uptime.h"
-#include "bm_printf.h"
-#include "bm_pubsub.h"
-#include "lwip/inet.h"
-#include "bm_network.h"
 #include "usart.h"
-#include "payload_uart.h"
-#include "OrderedKVPLineParser.h"
-#include "array_utils.h"
-#include "sensors.h"
+#include "util.h"
 
 #define LED_ON_TIME_MS 20
 #define LED_PERIOD_MS 1000
+#define AANDERAA_WATCHDOG_TIMEOUT_MS (5 * 1000)
+#define AANDERAA_WATCHDOG_MAX_TRIGGERS (3)
+#define AANDERAA_WATCHDOG_ID "Aanderaa"
+#define AANDERAA_RAW_LOG "aanderaa_raw.log"
+#define AANDERAA_RESET_TIME_MS                                                                 \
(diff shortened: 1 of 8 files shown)


src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
```diff
@@ -249,12 +264,12 @@ void loop(void) {
 
   // If there is a line to read, read it ad then parse it
   if (PLUART::lineAvailable()) {
-
+    SensorWatchdog::SensorWatchdogPet(AANDERAA_WATCHDOG_ID);
```
russelldeguzman, line 262: Pet the watchdog every time we parse a line from the aanderaa

src/lib/common/sensorWatchdog.cpp
```diff
@@ -0,0 +1,131 @@
...
+    while(curr) {
+        if(curr->_timerHandle == xTimer){
+            configASSERT(timer_callback_handler_send_cb(_sensorWatchdogHandler, reinterpret_cast<void *>(curr), 0) == pdTRUE);
+        }
+        curr = curr->_next;
+    }
+    xSemaphoreGive(_sensor_watchdog_list_mutex);
+}
+
+static void _sensorWatchdogHandler(void *arg) {
```
russelldeguzman, line 98: Run in the timer_handler thread.

src/lib/common/sensorWatchdog.cpp
```diff
@@ -0,0 +1,131 @@
...
+        curr->_next = cfg;
+    } else {
+        _sensor_watchdog_list = cfg;
+    }
+    xSemaphoreGive(_sensor_watchdog_list_mutex);
+}
+
+static void _timerCallback(TimerHandle_t xTimer) {
+    configASSERT(xTimer);
+    configASSERT(xSemaphoreTake(_sensor_watchdog_list_mutex, 0) == pdTRUE);
```
russelldeguzman, line 87: We block for zero in `_timerCallback` because:
```
Timer callback functions execute in the context of the timer service task. It is therefore essential that timer callback functions never attempt to block. For example, a timer callback function must not call vTaskDelay(), vTaskDelayUntil(), or specify a non zero block time when accessing a queue or a semaphore.
```
And instead we attempt to run in the callback task: `timer_handler`

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
```diff
@@ -117,6 +122,14 @@ currentData_t aggregateStats(AveragingSampler &sampler) {
   return ret_stats;
 }
 
+static bool aanderaaSensorWatchdogHandler(void * arg) {
+  (void) (arg);
+  IOWrite(&BB_PL_BUCK_EN, 1);
+  vTaskDelay(pdMS_TO_TICKS(100));
```
towynlin: Where does the 100 ms come from? Datasheet? Our testing? Other?

towynlin (approved):
Great job! Thanks for making it easily reusable for other sensors. Had one question about the 100ms power cycle.

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
russelldeguzman: Good question! This came from testing but I actually found a Note in the user manual:
https://www.aanderaa.com/media/pdfs/td266-zpulse-dcs-4420-4830-4520-4930.pdf

```(electronics require up to 500ms start up time).```

So I'll bump this up as a precaution, thanks for the flag!

PR #12, opened 2023-10-17
circ mean + std

russelldeguzman:
Adding circular mean and std.

Testing:
Unit test - 
circ mean verified against https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.circmean.html
circ std verified against custom python impl:
```
The wave std (in radians) is:
np.sqrt(2 - 2 * np.sqrt(a1 ** 2 + b1 ** 2)
where a1 is the mean of the cosine of the angles, and b1 is the mean of the sine of the angles.
```

### DIFF
--- a/src/lib/common/avgSampler.cpp
+++ b/src/lib/common/avgSampler.cpp
@@ -290,6 +290,7 @@ double AveragingSampler::getCircularMean() {
 
 /*!
   Compute circular standard deviation of all the samples
+Per Kuik et al 1988: https://doi.org/10.1175/1520-0485(1988)018%3C1020:AMFTRA%3E2.0.CO;2
 
   \return circular standard deviation of all the samples
  */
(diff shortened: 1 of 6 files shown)


src/lib/common/avgSampler.cpp
```diff
@@ -228,3 +228,76 @@ double AveragingSampler::getVariance(double mean, bool useKahan) {
...
+double AveragingSampler::getCircularMean() {
+  configASSERT(_samples != NULL);
+  double sin_sum = getTrigSum(TRIG_MEAN_TYPE_SIN);
+  double cos_sum = getTrigSum(TRIG_MEAN_TYPE_COS);
+  return atan2(sin_sum, cos_sum);
+}
+
+/*!
+  Compute circular standard deviation of all the samples
+
```
towynlin, line 294:
```suggestion
Per Kuik et al 1988: https://doi.org/10.1175/1520-0485(1988)018%3C1020:AMFTRA%3E2.0.CO;2

```

towynlin (approved):
Excellent. Thanks — added comment referencing the source of the algorithm.

PR #13, opened 2023-10-18
Sc 194765/aandera app

russelldeguzman:
This PR has two parts:

1. Bridge Aanderaa controller - responsible for querying the topology for all "aanderaa" apps and subscribing to <node_id>/aanderaa
2. Aanderaa app - responsible for pushing all parsed data to the <node_id>/aanderaa app

The bridge processing aanderaa data!
Screen Shot 2023-10-19 at 3 56 56 PM


src/lib/common/topology_sampler.h
```diff
@@ -7,4 +7,7 @@
 
 #include <stdint.h>
 
+#define TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE (24)
```
russelldeguzman: Fairly arbitrary! Don't think we'll be hitting this any time soon!

src/apps/bridge/aanderaaController.cpp
```diff
@@ -0,0 +1,233 @@
...
+  _ctx._subbed_aanderaas = NULL;
+  _ctx._num_subbed_aanderaas = 0;
+  _ctx._sample_timer = xTimerCreate("AanderaaCtlTim", pdMS_TO_TICKS(SAMPLE_TIMER_MS), pdTRUE,
+                                    NULL, sampleTimerCallback);
+  configASSERT(_ctx._sample_timer);
+  configASSERT(xTimerStart(_ctx._sample_timer, 10) == pdTRUE);
+  _ctx._initialized = true;
+  while (true) {
+    ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
+    if(_ctx._bridge_power_controller->waitForSignal(true,pdMS_TO_TICKS(TOPO_TIMEOUT_MS))) {
```
russelldeguzman: Wait's for BM Bus on to do a sample

src/apps/bridge/aanderaaController.cpp
```diff
@@ -0,0 +1,233 @@
...
+            }
+        }
+        }
+    }
+  }
+}
+
+static void createAanderaaSub(uint64_t node_id) {
+  Aanderaa_t *new_sub = reinterpret_cast<Aanderaa_t *>(pvPortMalloc(sizeof(Aanderaa_t)));
+  new_sub = new (new_sub) Aanderaa_t();
```
russelldeguzman, line 199: First time using a placement new constructor 😅. I think this is the proper use case though - We want to allocate a class (struct) on the heap and do the proper construction, but because we're using an RTOS, we can't use `new Aanderaa_t` because of our dependancy on `pvPortMalloc`.

src/apps/bridge/aanderaaController.cpp
```diff
@@ -0,0 +1,233 @@
...
+  }
+  _ctx._subbed_aanderaas = NULL;
+  _ctx._num_subbed_aanderaas = 0;
+  _ctx._sample_timer = xTimerCreate("AanderaaCtlTim", pdMS_TO_TICKS(SAMPLE_TIMER_MS), pdTRUE,
+                                    NULL, sampleTimerCallback);
+  configASSERT(_ctx._sample_timer);
+  configASSERT(xTimerStart(_ctx._sample_timer, 10) == pdTRUE);
+  _ctx._initialized = true;
+  while (true) {
+    ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
```
russelldeguzman, line 176: Driven by `SAMPLE_TIMER_MS`, which is set to a 30s period. Which means the bridge won't know about the Aanderaa until 30s into boot. Not strongly attached to this number, but we want to sample rather infrequently.

src/apps/bridge/aanderaaController.cpp
```diff
@@ -0,0 +1,233 @@
...
+                                                            "abs_speed_std_cm_s: %.3f "
+                                                            "direction_circ_mean_rad: %.3f "
+                                                            "direction_circ_std_rad: %.3f "
+                                                            "temp_mean_deg_c: %.3f\n",
+                                node_id, aanderaa->abs_speed_mean_cm_s.getMean(),
+                                aanderaa->abs_speed_std_cm_s.getStd(),
+                                aanderaa->direction_circ_mean_rad.getCircularMean(),
+                                aanderaa->direction_circ_std_rad.getCircularStd(),
+                                aanderaa->temp_mean_deg_c.getMean());
+        BRIDGE_LOG_PRINTN(log_buf, log_buflen);
```
russelldeguzman: Right now just log the avg sample to the spotter rotated bridge log as a placeholder. Do fancier stuff later.

src/lib/common/topology_sampler.h
towynlin: Another source for a fairly arbitrary number — we're defining this field in the satellite message to the Sofar backend as a 4-bit integer representing 1-16 where the bridge is always present and there may be up to 15 more.

src/apps/bridge/aanderaaController.cpp
towynlin: TIL!

src/apps/bridge/aanderaaController.cpp
```diff
@@ -0,0 +1,233 @@
...
+                }
+            }
+        }
+        }
+    }
+  }
+}
+
+static void createAanderaaSub(uint64_t node_id) {
+  Aanderaa_t *new_sub = reinterpret_cast<Aanderaa_t *>(pvPortMalloc(sizeof(Aanderaa_t)));
```
victorsowa12: nit: we should be able to use static_cast here since we are dealing with a void * casted to some other type. I think we would only need a reinterpret_cast when going from one struct to another struct (or some other complex, potentially unsafe cast).

victorsowa12 (approved):
👍, minor comment on casting but LGTM

src/apps/bridge/aanderaaController.cpp
victorsowa12: TIL too! 😄

src/apps/bridge/aanderaaController.cpp
russelldeguzman: Oops good catch!

PR #15, opened 2023-10-19
feat: calculate system config crc using the cbor map representation

towynlin:
For the same underlying config, this will change the calculated CRC value, since we're using a different representation that is more appropriate for serialization and communication across systems.

### DIFF
--- a/src/lib/sys/configuration.cpp
+++ b/src/lib/sys/configuration.cpp
@@ -31,8 +31,14 @@ bool Configuration::loadAndVerifyNvmConfig(void) {
         if(!_flash_partition.read(CONFIG_START_OFFSET_IN_BYTES, reinterpret_cast<uint8_t *>(_ram_partition), sizeof(ConfigPartition_t), CONFIG_LOAD_TIMEOUT_MS)){
             break;
         }
-        uint32_t computed_crc32 = crc32_ieee(reinterpret_cast<const uint8_t *>(&_ram_partition->header.version), (sizeof(ConfigPartition_t)-sizeof(_ram_partition->header.crc32)));
-        if(_ram_partition->header.crc32 != computed_crc32) {
+        size_t buffer_len = sizeof(ConfigPartition_t);
+        uint8_t buffer[buffer_len];
+        bool encodeSuccess = asCborMap(buffer, buffer_len);
+        if (!encodeSuccess) {
+            break;
+        }
+        uint32_t computed_crc32 = crc32_ieee(buffer, buffer_len);
+        if (_ram_partition->header.crc32 != computed_crc32) {
             break;
         }
         rval = true;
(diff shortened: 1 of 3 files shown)


src/lib/sys/configuration.cpp
```diff
@@ -31,8 +31,14 @@ bool Configuration::loadAndVerifyNvmConfig(void) {
         if(!_flash_partition.read(CONFIG_START_OFFSET_IN_BYTES, reinterpret_cast<uint8_t *>(_ram_partition), sizeof(ConfigPartition_t), CONFIG_LOAD_TIMEOUT_MS)){
             break;
         }
-        uint32_t computed_crc32 = crc32_ieee(reinterpret_cast<const uint8_t *>(&_ram_partition->header.version), (sizeof(ConfigPartition_t)-sizeof(_ram_partition->header.crc32)));
-        if(_ram_partition->header.crc32 != computed_crc32) {
+        size_t buffer_len = sizeof(ConfigPartition_t);
```
towynlin: We probably want to make this smaller. For any realistic short-term use case 256 bytes is probably more than sufficient.

russelldeguzman: Yeah makes sense to me - I think as long as we add debugging provision for the configuration system to panic loudly and obviously when we're hitting encoding buffer limits we're good. 

I don't have a good sense of the encoding efficiency of cbor - Our configurations are on average a [char[20] : uint32]. I think 256 is fine as long as we can do all of the bridge configs (7) plus probably 8 more in the short term?

src/lib/sys/configuration.cpp
```diff
@@ -31,8 +31,14 @@ bool Configuration::loadAndVerifyNvmConfig(void) {
...
+        uint8_t buffer[buffer_len];
```
russelldeguzman: Maybe the heap (or just `.bss`) is a better place to put this while we're still deciding on the size of `buffer_len`?
We have the potential to call configurations from a few different tasks atm, and all the FreeRTOS tasks have their own specified stack. A little concerned that this may cause an intermittent stack overflow that may be hard to track down in certain cases.

src/lib/sys/configuration.cpp
```diff
@@ -671,8 +677,16 @@ const char* Configuration::dataTypeEnumToStr(ConfigDataTypes_e type) {
 bool Configuration::saveConfig(bool restart) {
     bool rval = false;
     do {
-        _ram_partition->header.crc32 = crc32_ieee(reinterpret_cast<const uint8_t *>(&_ram_partition->header.version), (sizeof(ConfigPartition_t)-sizeof(_ram_partition->header.crc32)));
-        if(!_flash_partition.write(CONFIG_START_OFFSET_IN_BYTES, reinterpret_cast<uint8_t *>(_ram_partition), sizeof(ConfigPartition_t), CONFIG_LOAD_TIMEOUT_MS)){
+        size_t buffer_len = sizeof(ConfigPartition_t);
+        uint8_t buffer[buffer_len];
```
russelldeguzman: Same question as above

src/lib/sys/configuration.cpp
towynlin: Good call. I'll change that.

src/lib/sys/configuration.cpp
towynlin: I'm thinking I'll go ahead and make `asCborMap` more robust, including allocating more and retrying if we get `CborErrorOutOfMemory`. I don't want to put our future selves in pain when some future user adds a ton of configs and can't figure out why their hardware doesn't behave as expected.

russelldeguzman (approved):
Nice! 👍

PR #17, opened 2023-10-22
Feature/serial payload example

evanShap:
This is the demo code for the upcoming "Bristlemouth Dev Kit Guide: Integrating an RS232 Serial Sensor" guide [internal draft](https://www.notion.so/bristlemouth/Internal-Draft-Bristlemouth-Dev-Kit-Guide-Integrating-an-RS232-Serial-Sensor-abb7d4c9dd38454f941cbec92182142b#1c3921dea2664214b314735443b5af2e).

Ideally this is merged in and released as part of 0.6.0 prior to BristleCon on November 9.


DEVELOPER.md
```diff
@@ -4,6 +4,10 @@ This document (hopefully) details the where/why/how of bristlemouth firmware. Fo
 
 If you're developing on the Bristlemouth Development Kit, see the [BMDK_README.md](./src/apps/bm_devkit/BMDK_README.md).
 
+# Running Tests
+### Build Tests
+To run build tests locally, run `python3 tools/scripts/test/did_i_break_something.py tools/scripts/test/configs`.  This script will verify all apps compile successfully.
```
victorsowa12, line 9: do we list anywhere how to do just `make test` in the cmake directory or how to do `ctest`? Might be

evanShap: not sure, let's add in separate PR if needed?

src/apps/bm_devkit/serial_payload_example/user_code/user_code.cpp
```diff
@@ -81,8 +100,40 @@ void loop(void) {
...
+    printf("%02X ", byte_read);
+  }
+  if (readingBytesTimer > -1 &&
+      (u_int32_t)uptimeGetMs() - (u_int32_t)readingBytesTimer >= BYTES_CLUSTER_MS) {
+    printf("\n");
+    readingBytesTimer = -1;
+  }
+
   // Read a line if it is available
+  // Note - PLUART::setUseLineBuffer must be set true in setup to enable lines.
```
victorsowa12, line 130: what happens if you enable setUseLiineBuffer and setUserByteStreamBuffer? Do they work simultaneously?

victorsowa12 (approved):
Got a question about both the byte stream buffer and line reading working simultaneously, but looks good otherwise 👍

src/apps/bm_devkit/serial_payload_example/user_code/user_code.cpp
victorsowa12: i see that they are both enabled by default, but I just want to confirm that they can work at the same time

evanShap: Yep - they don't stomp on each other. PLUART will process either or both depending on the configurations.

PR #19, opened 2023-10-25
get sys info crc from nodes

russelldeguzman:
This PR aggregates all nodes sys-info crc's into the network's crc.
1. Topology sampler samples the nodes.
2. Open a top level cbor array
3. For each node discovered, request the sys info and add it to subarray
4. As you process each sys info, then send a request for the cbor map configuration
5. On receipt of the cbor map configuration, add it to the subarray and then close
6. When all nodes have been added, close the top level cbor array
7. Now compute the CRC32 over the entire encoded 2D cbor network configuration

TODO - we print the cbor array now for debug, but in another PR, send all this to Spotter.

Also clang-format

Testing:

A bridge, and two motes network config

```
83                                      # array(3)
   85                                   # array(5)
      1B C00F1C4D07116874               # unsigned(13839311297089464436)
      6A                                # text(10)
         6272696467652D646267           # "bridge-dbg"
      1A A64F9410                       # unsigned(2790233104)
      1A A89496A1                       # unsigned(2828310177)
      A1                                # map(1)
         78 1C                          # text(28)
            627269646765506F776572436F6E74726F6C6C6572456E61626C6564 # "bridgePowerControllerEnabled"
         00                             # unsigned(0)
   85                                   # array(5)
      1B D158D85954E1F6AD               # unsigned(15085044830065260205)
      6F                                # text(15)
         68656C6C6F5F776F726C642D646267 # "hello_world-dbg"
      1A 1C4B8AEA                       # unsigned(474712810)
      1A 04D44C65                       # unsigned(81022053)
      A0                                # map(0)
   85                                   # array(5)
      1B 2F58E75E9F6554B5               # unsigned(3411731111320310965)
      6F                                # text(15)
         68656C6C6F5F776F726C642D646267 # "hello_world-dbg"
      1A 1C4B8AEA                       # unsigned(474712810)
      1A 04D44C65                       # unsigned(81022053)
      A0                                # map(0)

```

Turns into:

```
[[13839311297089464436, "bridge-dbg", 2790233104, 2828310177, {"bridgePowerControllerEnabled": 0}], [15085044830065260205, "hello_world-dbg", 474712810, 81022053, {}], [3411731111320310965, "hello_world-dbg", 474712810, 81022053, {}]]
```

### DIFF
--- a/src/lib/common/topology_sampler.cpp
+++ b/src/lib/common/topology_sampler.cpp
@@ -1,23 +1,29 @@
-#include <string.h>
 #include "FreeRTOS.h"
-#include "task.h"
 #include "queue.h"
-#include "timers.h"
+#include "semphr.h"
+#include "task.h"
 #include "task_priorities.h"
 #include "timer_callback_handler.h"
-#include "semphr.h"
+#include "timers.h"
+#include <string.h>
 
-#include "bm_l2.h"
 #include "bcmp.h"
 #include "bcmp_info.h"
 #include "bcmp_messages.h"
 #include "bcmp_neighbors.h"
 #include "bcmp_topology.h"
 #include "bm_common_structs.h"
-#include "bm_util.h"
+#include "bm_l2.h"
 #include "bm_serial.h"
+#include "bm_util.h"
+#include "cbor.h"
+#include "config_cbor_map_service.h"
+#include "config_cbor_map_srv_reply_msg.h"
+#include "config_cbor_map_srv_request_msg.h"
 #include "crc.h"
 #include "device_info.h"
+#include "sys_info_service.h"
+#include "sys_info_svc_reply_msg.h"
 #include "topology_sampler.h"
 #include "util.h"
 
@@ -27,30 +33,48 @@
 #define TOPOLOGY_BEGIN_TIMEOUT_MS 30
 
 #define BUS_POWER_ON_DELAY 5000
+#define NODE_SYS_INFO_REQUEST_TIMEOUT_S (3)
+#define NETWORK_SYS_INFO_REQUEST_TIMEOUT_MS (NODE_SYS_INFO_REQUEST_TIMEOUT_S * 1000)
+#define NODE_CONFIG_CBOR_MAP_REQUEST_TIMEOUT_S (3)
+#define NODE_CONFIG_CBOR_MAP_REQUEST_TIMEOUT_MS (NODE_CONFIG_CBOR_MAP_REQUEST_TIMEOUT_S * 1000)
+#define NODE_CONFIG_PADDING                                                                    \
+  (512) // Accounts for name of app + cbor config map + encoding inefficiencies
+#define NUM_CONFIG_FIELDS_PER_NODE (5)
 
 typedef struct node_list {
   uint64_t nodes[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
   uint16_t num_nodes;
   SemaphoreHandle_t node_list_mutex;
-} node_list_s; 
+} node_list_s;
 
 static BridgePowerController *_bridge_power_controller;
-static cfg::Configuration* _hw_cfg;
-static cfg::Configuration* _sys_cfg;
+static cfg::Configuration *_hw_cfg;
+static cfg::Configuration *_sys_cfg;
 static TimerHandle_t topology_timer;
 static bool _sampling_enabled;
 static bool _send_on_boot;
 static node_list_s _node_list;
+static QueueHandle_t _sys_info_queue;
+static QueueHandle_t _config_cbor_map_queue;
 
 static void topology_timer_handler(TimerHandle_t tmr);
 static void topology_sampler_task(void *parameters);
+static bool sys_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
+                              const char *service, size_t reply_len, uint8_t *reply_data);
+static bool cbor_config_map_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
+                                     const char *service, size_t reply_len,
+                                     uint8_t *reply_data);
+static bool encode_sys_info(CborEncoder &array_encoder,
+                                SysInfoSvcReplyMsg::Data &sys_info);
+static bool encode_cbor_configuration(CborEncoder &array_encoder,
+                                      ConfigCborMapSrvReplyMsg::Data &cbor_map_reply);
+static bool create_network_info_cbor_array(uint8_t *cbor_buffer, size_t &cbor_bufsize);
+static void topology_sample_cb(networkTopology_t *networkTopology) {
 
-static void topology_sample_cb(networkTopology_t* networkTopology) {
-
-  bm_common_network_info_t* network_info = NULL;
+  bm_common_network_info_t *network_info = NULL;
   xSemaphoreTake(_node_list.node_list_mutex, portMAX_DELAY);
   do {
-    if (!networkTopology){
+    if (!networkTopology) {
       printf("networkTopology NULL, task must be busy\n");
       break;
     }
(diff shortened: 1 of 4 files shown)


src/lib/common/topology_sampler.cpp
```diff
@@ -64,53 +71,98 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
...
+    xQueueReset(sys_info_crc_queue);
+    for (cursor = networkTopology->front, counter = 0;
+         (cursor != NULL) && (counter < _node_list.num_nodes);
+         cursor = cursor->nextNode, counter++) {
       _node_list.nodes[counter] = cursor->neighbor_table_reply->node_id;
-      network_crc32_calc = crc32_ieee_update(network_crc32_calc, reinterpret_cast<uint8_t *>(&_node_list.nodes[counter]), sizeof(uint64_t));
+      network_crc32_calc = crc32_ieee_update(
+          network_crc32_calc, reinterpret_cast<uint8_t *>(&_node_list.nodes[counter]),
+          sizeof(uint64_t));
+      if (!sys_info_service_request(_node_list.nodes[counter], sys_info_reply_cb,
```
russelldeguzman, line 108: After we have our list of nodes, now we can request the sys info from each node, which will come back async from `sys_info_reply_cb`

src/lib/common/topology_sampler.cpp
```diff
@@ -64,53 +71,98 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
...
+        break;
+      }
+    }
+
+    // Wait for all the sys info replies to come in
+    uint32_t node_crc_rx_count = 0;
+    while (node_crc_rx_count < _node_list.num_nodes) {
+      uint32_t node_crc = 0;
+      // Update the network crc with all of the node crcs
+      if (xQueueReceive(sys_info_crc_queue, &node_crc,
```
russelldeguzman: We wait on all of the CRC's from the nodes to come back and aggregate them on the network crc.

src/lib/common/topology_sampler.cpp
```diff
@@ -149,12 +199,60 @@ static void topology_timer_cb(void *arg) {
...
+ * 
+ * @param[in] ack True if the message was acknowledged, false otherwise.
+ * @param[in] msg_id The message id.
+ * @param[in] service_strlen The length of the service string.
+ * @param[in] service The service string.
+ * @param[in] reply_len The length of the reply data.
+ * @param[in] reply_data The reply data.
+ * @return True if the request was handled, false otherwise.
+ */
+static bool sys_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
```
russelldeguzman, line 236: Sys info callback for each of the nodes - prints out the info, computes the crc based on the info, and sends it back to the topology sampler task

src/lib/common/topology_sampler.cpp
```diff
@@ -64,53 +71,100 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
 
     // compile all additional info here
     _node_list.num_nodes = networkTopology->length;
-    configASSERT(sizeof(_node_list.nodes) >= _node_list.num_nodes * sizeof(uint64_t)); // if we hit this, expand the node list size
+    configASSERT(sizeof(_node_list.nodes) >=
+                 _node_list.num_nodes *
+                     sizeof(uint64_t)); // if we hit this, expand the node list size
```
towynlin: Maybe move this comment to the line before the assertion. Feel free to override clang-format when it's ugly. This one is a silly suggestion, but I find that when it does this, it's usually trying to tell us put comments on their own line.

src/lib/common/topology_sampler.cpp
```diff
@@ -64,53 +71,100 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
...
       _node_list.nodes[counter] = cursor->neighbor_table_reply->node_id;
-      network_crc32_calc = crc32_ieee_update(network_crc32_calc, reinterpret_cast<uint8_t *>(&_node_list.nodes[counter]), sizeof(uint64_t));
+      network_crc32_calc = crc32_ieee_update(
+          network_crc32_calc, reinterpret_cast<uint8_t *>(&_node_list.nodes[counter]),
+          sizeof(uint64_t));
+      if (!sys_info_service_request(_node_list.nodes[counter], sys_info_reply_cb,
+                                    NODE_SYS_INFO_REQUEST_TIMEOUT_S)) {
+        printf("Failed to send sys info request to node: %" PRIu64 "\n",
+               _node_list.nodes[counter]);
+        break;
```
towynlin, line 113: Hm, I'm not sure we want to break here, and if we do it's only breaking out of the for loop, not the top-level do-while. I don't know the reasons this could fail. If it's recoverable, we could try again later. If not, then we probably want to abort the whole calculation or maybe even crash. What do you think? We *need* this full calculation to have a consistent outcome; if it doesn't, continuing to run and send data with an incorrect CRC will cause bad data on the wavefleet side.

src/lib/common/topology_sampler.cpp
```diff
@@ -64,53 +71,100 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
...
-      .partition = BM_COMMON_CFG_PARTITION_SYSTEM,
-      .crc32 = _sys_cfg->getCRC32(),
+        .partition = BM_COMMON_CFG_PARTITION_SYSTEM,
+        .crc32 = _sys_cfg->getCRC32(),
     };
 
-    network_crc32_calc = crc32_ieee_update(network_crc32_calc, reinterpret_cast<uint8_t *>(&config_crc), sizeof(bm_common_config_crc_t));
+    network_crc32_calc =
+        crc32_ieee_update(network_crc32_calc, reinterpret_cast<uint8_t *>(&config_crc),
+                          sizeof(bm_common_config_crc_t));
```
towynlin: is this one now duplicating what's received from the sysinfo request?

towynlin (changes requested):
OK, thank you so much. You're so fast! This is a solid start — great job using the queue to receive the async replies. 👏 

I think we need all the info returned by the sysinfo requests built into a cbor array, and then the crc for the network needs to be calculated over that whole cbor array. I don't think we need some of the previous info that was a part of the network crc, or at least we don't need it duplicated in the same format as before. I'll add more detail to shortcut.

Sorry for the lack of clarity in the story, and thank you for getting it to this point so quickly!

src/lib/common/topology_sampler.cpp
russelldeguzman: Ah I guess so - since we ask ourselves what our own CRC32 from sys config is.

src/lib/common/topology_sampler.cpp
russelldeguzman: I think returning is sufficient here.

src/lib/common/topology_sampler.cpp
```diff
@@ -64,53 +88,101 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
...
+               _node_list.nodes[counter]);
+        exit = true;
+        break;
+      }
+    }
+    if (exit) {
+      break;
     }
 
-    bm_common_config_crc_t config_crc = {
```
russelldeguzman, line 80: Narrowing scope for these variables below.

src/lib/middleware/bm_service_request.cpp
```diff
@@ -141,7 +151,9 @@ static bool _request_list_remove_request(bm_service_request_node_t * node) {
 static bm_service_request_node_t * _create_node(size_t service_strlen, const char * service, bm_service_reply_cb reply_cb, uint32_t timeout_ms) {
     bm_service_request_node_t * node = static_cast<bm_service_request_node_t*>(pvPortMalloc(sizeof(bm_service_request_node_t)));
     configASSERT(node);
-    node->service = service;
+    node->service = static_cast<char*>(pvPortMalloc(service_strlen));
```
russelldeguzman, line 154: Adding support so our nodes always have reference to their service name (in case the caller frees the name buffer)

src/lib/middleware/bm_service_request.cpp
```diff
@@ -132,6 +139,9 @@ static bool _request_list_remove_request(bm_service_request_node_t * node) {
         }
     }
     if(node_to_delete){
+        if(node_to_delete->service) {
```
russelldeguzman, line 142: Make sure to free the name now that we're allocating it!

src/lib/middleware/services/config_cbor_map_service.cpp
```diff
@@ -83,10 +83,13 @@ static bool config_map_service_handler(size_t service_strlen, const char *servic
             printf("Invalid partition id\n");
         }
         size_t buffer_size;
+        if(config) {
+            cbor_map = config->asCborMap(buffer_size);
```
russelldeguzman, line 87: Fixing a memory leak 🙀 
We free `cbor_map` at the end of the function.

src/lib/middleware/bm_service_request.h
```diff
@@ -5,6 +5,7 @@
 
 /*!
  * @brief Callback function for bm_service_request
+ * @note Should not call bm_service_request from within this callback.
```
russelldeguzman, line 8: Calling a request nested in the callback can cause a deadlock.  

We can add support for this when we need it

src/lib/common/topology_sampler.cpp
```diff
@@ -27,30 +33,48 @@
 #define TOPOLOGY_BEGIN_TIMEOUT_MS 30
 
 #define BUS_POWER_ON_DELAY 5000
+#define NODE_SYS_INFO_REQUEST_TIMEOUT_S (3)
+#define NETWORK_SYS_INFO_REQUEST_TIMEOUT_MS (NODE_SYS_INFO_REQUEST_TIMEOUT_S * 1000)
```
towynlin: Any objection to changing this to start with `NODE_` for consistency?

src/lib/common/topology_sampler.cpp
```diff
@@ -27,30 +33,48 @@
...
+#define NODE_CONFIG_CBOR_MAP_REQUEST_TIMEOUT_S (3)
+#define NODE_CONFIG_CBOR_MAP_REQUEST_TIMEOUT_MS (NODE_CONFIG_CBOR_MAP_REQUEST_TIMEOUT_S * 1000)
+#define NODE_CONFIG_PADDING                                                                    \
+  (512) // Accounts for name of app + cbor config map + encoding inefficiencies
```
towynlin: Please move the comment to its own line.

src/lib/common/topology_sampler.cpp
```diff
@@ -64,53 +88,102 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
...
-      .revision = 0,
-      .gitSHA = 0,
-    };
-
-    getFWVersion(&fw_info.major, &fw_info.minor, &fw_info.revision);
-    fw_info.gitSHA = getGitSHA();
+    // Create the network info cbor array and calculate the crc32
+    size_t cbor_bufsize =
+        _node_list.num_nodes * (sizeof(SysInfoSvcReplyMsg::Data) +
+                                sizeof(ConfigCborMapSrvReplyMsg::Data) + NODE_CONFIG_PADDING);
```
towynlin, line 123: Right now, for our current use cases, `NODE_CONFIG_PADDING` seems too big. However, future use cases will definitely require more space as Bristlemouth networks have more nodes and more system config values set. The example in the PR description only needed 142 bytes but this would have allocated 1680. (I believe each of the msg data types would be 24 bytes.)

My main question is how much heap do we have available? If we left this as is, allocating 560 for each node, how many nodes would use up our heap? Any idea? No big deal for DVT2 since 3 is the max we expect in this phase. Just wanted to share my 🤔 .

src/lib/common/topology_sampler.cpp
```diff
@@ -64,53 +88,102 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
...
+      // TODO: Printing the buffer for now, but we should send this to spotter before freeing. 
+      for(uint32_t i = 0; i < cbor_bufsize; i++) {
+        printf("%02x ", cbor_buffer[i]);
+        if(i % 16 == 15) { // print a newline every 16 bytes (for print pretty-ness)
+          printf("\n");
+        }
+      }
+      printf("\n");
+      vPortFree(cbor_buffer);
+    }
```
towynlin: I think the `vPortFree(cbor_buffer)` needs to move down around line 193, outside the top-level do-while.

Also, the printing/sending currently happens even if `encoding_success == false`. It needs to move either into the above nested do-while, around line 133, or somewhere else down below.

src/lib/common/topology_sampler.cpp
```diff
@@ -149,12 +220,255 @@ static void topology_timer_cb(void *arg) {
...
+    printf("Reply: %" PRIu32 "\n", msg_id);
+    printf(" * Node id: %" PRIx64 "\n", reply.node_id);
+    printf(" * Git SHA: 0x%08" PRIx32 "\n", reply.git_sha);
+    printf(" * Sys config CRC: 0x%08" PRIx32 "\n", reply.sys_config_crc);
+    printf(" * App name: %.*s\n", reply.app_name_strlen, reply.app_name);
+    if (xQueueSend(_sys_info_queue, &reply, 0) != pdTRUE) {
+      printf("Failed to send sys info\n");
+    }
+  } else {
+    printf("NACK\n");
```
towynlin: Since we have multiple requests that might nack, print more info here to aid debugging.

src/lib/common/topology_sampler.cpp
```diff
@@ -149,12 +220,255 @@ static void topology_timer_cb(void *arg) {
...
+          printf("\n");
+        }
+      }
+    }
+    printf("\n");
+    if (!xQueueSend(_config_cbor_map_queue, &reply, 0)) {
+      printf("Failed to send cbor map\n");
+    }
+  } else {
+    printf("NACK\n");
```
towynlin: Since we have multiple requests that might nack, print more info here to aid debugging.

src/lib/common/topology_sampler.cpp
```diff
@@ -149,12 +220,255 @@ static void topology_timer_cb(void *arg) {
...
+    }
+    printf("Service: %.*s\n", service_strlen, service);
+    printf("Reply: %" PRIu32 "\n", msg_id);
+    printf(" * Node id: %" PRIx64 "\n", reply.node_id);
+    printf(" * Git SHA: 0x%08" PRIx32 "\n", reply.git_sha);
+    printf(" * Sys config CRC: 0x%08" PRIx32 "\n", reply.sys_config_crc);
+    printf(" * App name: %.*s\n", reply.app_name_strlen, reply.app_name);
+    if (xQueueSend(_sys_info_queue, &reply, 0) != pdTRUE) {
+      printf("Failed to send sys info\n");
+    }
```
towynlin: The call to `decode` here (as well as on line 277 where I won't repeat the same comment) allocates memory for `app_name`. It took me a while to follow these through the queues and functions to determine that they were being freed.

So, first, some comments could be helpful there, like on the line before the `xQueueSend`, comment that `app_name` will be freed on the receiving side, and on the line before the subsequent encode call, say that the encoding function frees `app_name`.

Second, allocated memory needs to be freed in case of errors, for example around line 301 if the queue send fails, and probably more cases like that.

src/lib/common/topology_sampler.cpp
```diff
@@ -149,12 +220,255 @@ static void topology_timer_cb(void *arg) {
...
+ * @brief Encodes the cbor configuration map into the cbor array.
+ *
+ * @param[in/out] array_encoder The encoder to encode into.
+ * @param[in] cbor_map_reply The cbor map reply to encode.
+ * @return True if the cbor map reply was encoded, false otherwise.
+ */
+static bool encode_cbor_configuration(CborEncoder &array_encoder,
+                                      ConfigCborMapSrvReplyMsg::Data &cbor_map_reply) {
+  CborError err = cbor_encode_byte_string(&array_encoder, cbor_map_reply.cbor_data,
+                                  cbor_map_reply.cbor_encoded_map_len);
```
towynlin: Please encode the actual cbor map here rather than a byte string.

towynlin (changes requested):
Phew! Solid. I'm learning a lot by seeing how you're doing this. Thank you! Still some more changes required, but it's close.

src/lib/common/topology_sampler.cpp
russelldeguzman: On the order of 126k
```Available:            126216```

So plenty for now, I can reduce the padding to 128 if that's a more reasonable ceiling.

Comment (towynlin) [reply src/lib/common/topology_sampler.cpp:123]: Oh wow, that's more than I expected. Just leave it. 👍

File: src/lib/common/topology_sampler.cpp (hunk 16 above)
Comment (russelldeguzman) [reply src/lib/common/topology_sampler.cpp]: Added the service + msg id at the top of the function.

File: src/lib/common/topology_sampler.cpp (hunk 17 above)
Comment (russelldeguzman) [reply src/lib/common/topology_sampler.cpp]: Added the service + msg id at the top of the function.

File: src/lib/common/topology_sampler.cpp [hunk 20]
```diff
@@ -64,68 +89,112 @@ static void topology_sample_cb(networkTopology_t* networkTopology) {
...
+    configASSERT(cbor_buffer);
 
-    getFWVersion(&fw_info.major, &fw_info.minor, &fw_info.revision);
-    fw_info.gitSHA = getGitSHA();
+    if (create_network_info_cbor_array(cbor_buffer, cbor_bufsize)) {
+      network_crc32_calc = crc32_ieee(cbor_buffer, cbor_bufsize);
+    } else {
+      printf("Failed to create network info cbor array\n");
+      break;
+    }
```
towynlin, line 132: Made a small, quick refactor in commit 863ca97 rather than asking you to do it @russelldeguzman.

- Cuts 13 lines with 2 levels of nesting to 6 lines with one level of nesting
- Removes a loop within a loop, making break behavior easier to understand at a glance
- Removes the need for the `encoding_success` variable

towynlin (approved):
Solid work. Thank you!

PR #20, opened 2023-10-26
use kahan summation for aanderaa

russelldeguzman:


test/src/common/avgSampler_ut.cpp
```diff
@@ -212,3 +212,26 @@ TEST_F(AvgSamplerTest, circ) {
...
+                         3235, 64, 243, 234, 234, 234, 234, 234, 234, 234,
+                         54, 54, 54, 54, 54, 54, 54, 54, 54, 54};
+  const uint32_t num_samples = sizeof(samplesDeg) / sizeof(double);
+
+  for (uint16_t sample = 0; sample < num_samples; sample++) {
+    // Use sample+1 as timestamp for now
+    EXPECT_TRUE(sampler.addSampleTimestamped(degToRad(samplesDeg[sample]), (sample + 1)));
+  }
+
+  EXPECT_NEAR(sampler.getRunningWelfordVariance(), sampler.getVariance(), 0.5);
```
russelldeguzman, line 235: There's a bit of mathematical delta here, but seems to converge as the number of samples grows.

PR #21, opened 2023-10-27
Transmit aggregations config

victorsowa12:
Adding a config to the bridge to control when to transmit aggregations.

Testing:
When not set, by default we transmit agg:
Screen Shot 2023-10-27 at 3 04 27 PM

When set to 0, we do not transmit agg:
Screen Shot 2023-10-27 at 3 04 14 PM

When set to 1, we transmit agg:
Screen Shot 2023-10-27 at 3 04 00 PM

### DIFF
--- a/src/apps/bridge/aanderaaController.cpp
+++ b/src/apps/bridge/aanderaaController.cpp
@@ -1,5 +1,6 @@
 #include "aanderaaController.h"
 #include "aanderaa_data_msg.h"
+#include "app_config.h"
 #include "avgSampler.h"
 #include "bm_network.h"
 #include "bm_pubsub.h"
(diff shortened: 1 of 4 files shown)


src/apps/bridge/aanderaaController.cpp
```diff
@@ -46,6 +47,7 @@ typedef struct aanderaaControllerCtx {
   TimerHandle_t _sample_timer;
   BridgePowerController *_bridge_power_controller;
   cfg::Configuration *_usr_cfg;
+  uint32_t _transmit_aggregations;
```
towynlin, line 50: Why uint32 instead of bool? Are there more possible values? Does the config system not have a notion of bool?

towynlin (approved):
Looks good — just one question about why a conceptual bool is encoded as an int.

russelldeguzman (approved):

src/apps/bridge/aanderaaController.cpp
victorsowa12: Let me check, I just based this on the bridge_power_controller's `bridgePowerControllerEnabled` which uses a uint32.

victorsowa12: No bool in the config data types
```
typedef enum ConfigDataTypes{
    UINT32,
    INT32,
    FLOAT,
    STR,
    BYTES,
} ConfigDataTypes_e;
```

PR #22, opened 2023-10-28
Feature/aanderaa app configs

evanShap:
This PR includes:
- :bug: the rtc value from the on-board aggregation is basically always 0, because the system starts the aggregation before the UTC fix from Spotter comes in. -> use rtc_end
- replace the hardcoded `CURRENT_SAMPLE_PERIOD_MS` with a config named "`CURRENT_READING_INTERVAL_MS`, which defaults to 60s - the intended reading interval for Aanderaa DVT2.
- implement a baud rate configuration
- configify nSkipReadings
- configify readingIntervalMs
- configify payloadWdToS

These changes have been tested in DVT1 systems and are dry-soaking through October 30

sc-194765

### DIFF
--- a/src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
+++ b/src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
@@ -89,7 +89,7 @@ static RTCTimeAndDate_t statsEndRtc =
     {}; // Timestamp that tracks the end of aggregation periods.
 
 // app_main passes a handle to the user config partition in NVM.
-extern cfg::Configuration *userConfigurationPartition;
+extern cfg::Configuration *systemConfigurationPartition;
 
 /*
  * Setup a LineParser to turn the ASCII serial data from the Aanderaa into numbers
@@ -186,16 +186,16 @@ static int createAanderaaDataTopic(void) {
 void setup(void) {
   /* USER ONE-TIME SETUP CODE GOES HERE */
   // Retrieve user-set config values out of NVM.
-  configASSERT(userConfigurationPartition);
-  userConfigurationPartition->getConfig("currentAggPeriodMin", strlen("currentAggPeriodMin"),
+  configASSERT(systemConfigurationPartition);
+  systemConfigurationPartition->getConfig("currentAggPeriodMin", strlen("currentAggPeriodMin"),
                                         current_agg_period_min);
-  userConfigurationPartition->getConfig("nSkipReadings", strlen("nSkipReadings"),
+  systemConfigurationPartition->getConfig("nSkipReadings", strlen("nSkipReadings"),
                                         n_skip_readings);
-  userConfigurationPartition->getConfig("readingIntervalMs", strlen("readingIntervalMs"),
+  systemConfigurationPartition->getConfig("readingIntervalMs", strlen("readingIntervalMs"),
                                         reading_interval_ms);
-  userConfigurationPartition->getConfig("payloadWdToS", strlen("payloadWdToS"),
+  systemConfigurationPartition->getConfig("payloadWdToS", strlen("payloadWdToS"),
                                         payload_wd_to_s);
-  userConfigurationPartition->getConfig("plUartBaudRate", strlen("plUartBaudRate"),
+  systemConfigurationPartition->getConfig("plUartBaudRate", strlen("plUartBaudRate"),
                                         baud_rate);
 
   max_readings_in_agg = (((uint64_t)CURRENT_AGG_PERIOD_MS / reading_interval_ms) + N_SAMPLES_PAD);
@@ -220,8 +220,8 @@ void setup(void) {
   // Enable the input to the Vout power supply and ensure PL buck is disabled.
   IOWrite(&BB_VBUS_EN, 0);
   IOWrite(&BB_PL_BUCK_EN, 1);
-  // ensure Vbus stable before enable Vout with a 1.5s delay.
-  vTaskDelay(pdMS_TO_TICKS(1500));
+  // ensure Aanderaa is off for sufficient time to fully reset.
+  vTaskDelay(pdMS_TO_TICKS(AANDERAA_RESET_TIME_MS));
   // enable Vout, 12V by default.
   IOWrite(&BB_PL_BUCK_EN, 0);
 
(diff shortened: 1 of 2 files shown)


evanShap:
sc-194765

evanShap:
SC-194765

evanShap:
https://app.shortcut.com/sofar/story/194765/implement-bristlemouth-current-meter-module-firmware

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
```diff
@@ -21,13 +21,51 @@
...
+// - If using reading intervals <5 seconds, recommend 2, otherwise 0.
+// It takes time for some internal Aanderaa init, and we're not parsing the "*" characters to infer errors yet.
+#define DEFAULT_N_SKIP_READINGS (0)
+static uint32_t n_skip_readings = DEFAULT_N_SKIP_READINGS;
+static uint8_t readings_skipped = 0;
+
+// The reading interval in ms the Aanderaa is configured for.
+// - Currently, must take care to manually set Aanderaa and this setting to the same values.
+#define DEFAULT_CURRENT_READING_INTERVAL_MS (60*1000)
+static uint32_t reading_interval_ms = DEFAULT_CURRENT_READING_INTERVAL_MS;
```
towynlin, line 48: The combination of "must take care" and 60s here vs. 2s and 1s mentioned in PR description feels like a yellow flag. What have we currently decided is actually the best reading interval? Can we make it the default? Can we set the interval in the Aanderaa sensor from firmware, or does it require a different tool?

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
```diff
@@ -162,36 +189,52 @@ void setup(void) {
...
   userConfigurationPartition->getConfig("currentAggPeriodMin", strlen("currentAggPeriodMin"),
                                         current_agg_period_min);
+  userConfigurationPartition->getConfig("nSkipReadings", strlen("nSkipReadings"),
+                                        n_skip_readings);
+  userConfigurationPartition->getConfig("readingIntervalMs", strlen("readingIntervalMs"),
+                                        reading_interval_ms);
+  userConfigurationPartition->getConfig("payloadWdToS", strlen("payloadWdToS"),
+                                        payload_wd_to_s);
+  userConfigurationPartition->getConfig("plUartBaudRate", strlen("plUartBaudRate"),
+                                        baud_rate);
```
towynlin: Just to confirm — since these are user configs, they won't be sent to the Sofar backend through Spotter. Is that the behavior we want? If we want to have remote visibility into these values, they need to be system configs.

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
```diff
@@ -162,36 +189,52 @@ void setup(void) {
...
   // Initialize our LineParser, which will allocated any needed memory for parsing.
   parser.init();
   // Setup the UART – the on-board serial driver that talks to the RS232 transceiver.
   PLUART::init(USER_TASK_PRIORITY);
-  // Baud set per expected baud rate of the sensor.
-#ifdef AANDERAA_BLUE
-  PLUART::setBaud(115200);
-#else
-  PLUART::setBaud(9600);
-#endif
```
towynlin, line 182: Does 9600 baud work for both models or can we get rid of all the code referencing the blue?

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
```diff
@@ -162,36 +189,52 @@ void setup(void) {
...
   // Turn on the UART.
   PLUART::enable();
-  // Enable the input to the Vout power supply.
+  // Enable the input to the Vout power supply and ensure PL buck is disabled.
   IOWrite(&BB_VBUS_EN, 0);
-  // ensure Vbus stable before enable Vout with a 5ms delay.
-  vTaskDelay(pdMS_TO_TICKS(5));
+  IOWrite(&BB_PL_BUCK_EN, 1);
+  // ensure Vbus stable before enable Vout with a 1.5s delay.
+  vTaskDelay(pdMS_TO_TICKS(1500));
```
towynlin: I'm curious about the change from a 5ms delay to 1500ms. In September, Michael and I were fixing I2C bugs, and he did some oscilloscope validations to see how long of a delay we needed in order for the 3v3 rail to stabilize. Slack conversation [here](https://sofarocean.slack.com/archives/C05J96Q6QLX/p1695754718673349?thread_ts=1695750119.795059&cid=C05J96Q6QLX). Does it take longer with the Aanderaa sensor attached? Or did you find you needed to increase it this much based on experience testing the sensor in real conditions?

towynlin (commented):
I've got a bunch of review nits that aren't important given our current deadlines, but I had 4 questions that seem important.

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
evanShap: Sorry, PR comments were auto-generated from commit tags - the 1s and 2s are no longer relevant. 60s is the decided default for DVT2.
In principal we can eventually set the reading interval on the Aanderaa in firmware, but this will require a more comprehensive driver to be developed.

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
evanShap: Good catch - I'll move these to sys partition and retest.

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
evanShap: 9600 baud is actually the default for both models - 115200 ended up being captured in the `#ifdef AANDERAA_BLUE` because the baud on the 1 Blue sensor we had had been adjusted to 115200 by someone before it made it to Russ.

src/apps/bristleback_apps/aanderaa/user_code/user_code.cpp
evanShap: I had originally changed the delay to `AANDERAA_RESET_TIME_MS` (500ms) per the note Russ added in the definition of that constant. Intent is to ensure that whatever was going on before we got here, we turned the thing off long enough to ensure it fully power cycled.
I then increased the delay to a hard-coded 1.5s because 500ms was too fast for my serial terminal to reliably capture the debug prints a few lines down. 😬 
I'll revert to `AANDERAA_RESET_TIME_MS`, which is the best-informed duration here imo.

towynlin: Ah! Right, yes, awesome. I remember him looking up that duration. Perfect.

towynlin (approved):

PR #24, opened 2023-10-30
Sc 196416/config map service

russelldeguzman:
Adding a cbor config map service 

Note that right now, our config system will return a stringz decoded key:
```{"hello unint 1" : 1}```  vs ```{"hello":1}```
This will be handled in an upcoming PR since it's unrelated to the changes in this PR.

### DIFF
--- a/src/lib/debug/debug_bm_service.cpp
+++ b/src/lib/debug/debug_bm_service.cpp
@@ -106,7 +106,7 @@ static bool config_map_callback(bool ack, uint32_t msg_id, size_t service_strlen
             break;
           }
           printf("%02x ", reply.cbor_data[i]);
-          if(i%16 == 15) {
+          if(i%16 == 15) { // print a newline every 16 bytes (for print pretty-ness)
             printf("\n");
           }
         }
--- a/src/lib/debug/debug_bm_service.cpp
+++ b/src/lib/debug/debug_bm_service.cpp
@@ -13,6 +13,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include "sys_info_svc_reply_msg.h"
+#include "config_cbor_map_service.h"
+#include "config_cbor_map_srv_reply_msg.h"
+#include "config_cbor_map_srv_request_msg.h"
 
 static BaseType_t bmServiceCmd(char *writeBuffer, size_t writeBufferLen,
                                const char *commandString);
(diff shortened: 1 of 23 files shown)

### CODE REVIEW

File: src/lib/debug/debug_bm_service.cpp [hunk 1]
```diff
@@ -78,6 +82,47 @@ static bool sys_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
...
+      if(reply.success) {
+        for(uint32_t i = 0; i<reply.cbor_encoded_map_len; i++) {
+          if(!reply.cbor_data) {
+            printf("NULL map\n");
+            break;
+          }
+          printf("%02x ", reply.cbor_data[i]);
+          if(i%16 == 15) {
+            printf("\n");
+          }
```
victorsowa12: 🪄 could you add a comment explaining why 16/15, or a #define for these magical numbers please?

src/lib/middleware/services/config_cbor_map_service.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+bool config_cbor_map_service_request(uint64_t target_node_id, uint32_t partition_id, bm_service_reply_cb reply_cb, uint32_t timeout_s) { 
+  bool rval = false;
+  char *target_service_str = static_cast<char *>(pvPortMalloc(BM_SERVICE_MAX_SERVICE_STRLEN));
+  configASSERT(target_service_str);
+  size_t target_service_strlen =
+      snprintf(target_service_str, BM_SERVICE_MAX_SERVICE_STRLEN, "%" PRIx64 "%s",
+               target_node_id, CONFIG_MAP_SUFFIX);
+  ConfigCborMapSrvRequestMsg::Data req = {partition_id};
+  static constexpr size_t cbor_buflen = sizeof(ConfigCborMapSrvRequestMsg::Data) + 25; // Abitrary padding.
```
victorsowa12, line 41: why do we need padding?

victorsowa12 (approved):

src/lib/middleware/services/config_cbor_map_service.cpp
russelldeguzman: The padding accounts for Cbor's encoding inefficiency. 25 as padding seems like a reasonable balance between "it-works", without over-allocating for this static length piece of data.

victorsowa12: Gotcha!

PR #26, opened 2023-10-30
Sc 196635 - refactor aanderaaController.cpp to use one timer for all aggregations

victorsowa12:
Refactored aanderaaController.cpp to use one timer to mark the end of an aggregation period for all of the aanderaa's instead of one timer **per** aanderaa. 

This entailed creating a new software timer that will fire every `currentAggPeriodMin`. Originally I was going to process the aggregated readings in this timers callback, but realizzed this may involve blocking code(a big no-no in timer callbacks). Instead I modified the existing task notifications to use bits, such that the timer callback will set a bit when it is time to end the aggregation. Now the aandderaa controller task will be notified when the aggregation period is complete and it will do the computations on the buffers, send them to the "report builder" task (just a todo comment now), and then clear the buffers. 

Since the buffers get new readings added to their buffers in the middleware when receive a message we subscribed to, I added a mutex lock so that we can safely access the buffers from the aanderaa controller app to calculate the mean/stdev + clear them for the next aggregation period.

I also added a small check to the buffers to make sure they have a minimum amount of readings within them and if not we will send NAN for the means/stdevs. This will guarantee that the "report builder" will get a struct for every aanderaa, even if the aanderaa node stops working part way through a sample, or between samples.

### DIFF
--- a/src/apps/bridge/aanderaaController.h
+++ b/src/apps/bridge/aanderaaController.h
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 
 #define DEFAULT_TRANSMIT_AGGREGATIONS 1
+#define DEFAULT_SAMPLES_PER_REPORT 2
 
 void aanderaControllerInit(BridgePowerController *power_controller,
                            cfg::Configuration *usr_cfg,
--- a/src/apps/bridge/aanderaaController.h
+++ b/src/apps/bridge/aanderaaController.h
@@ -9,7 +9,6 @@
 #include <stdlib.h>
 
 #define DEFAULT_TRANSMIT_AGGREGATIONS 1
-#define DEFAULT_SAMPLES_PER_REPORT 2
 
 void aanderaControllerInit(BridgePowerController *power_controller,
                            cfg::Configuration *usr_cfg,
--- a/src/apps/bridge/aanderaaController.h
+++ b/src/apps/bridge/aanderaaController.h
@@ -10,6 +10,15 @@
 
 #define DEFAULT_TRANSMIT_AGGREGATIONS 1
 
+typedef struct aanderaa_aggregations_s {
+  uint64_t node_id;
+  double abs_speed_mean_cm_s;
+  double abs_speed_std_cm_s;
+  double direction_circ_mean_rad;
+  double direction_circ_std_rad;
+  double temp_mean_deg_c;
+} aanderaa_aggregations_t;
+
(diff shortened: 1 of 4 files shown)


src/lib/middleware/bm_network.cpp
```diff
@@ -19,7 +19,7 @@ static constexpr uint8_t networkTopicType = 1;
 bool spotter_tx_data(const void* data, uint16_t data_len, bm_serial_network_type_e type) {
     bool rval = false;
     size_t msg_len = sizeof(bm_serial_network_data_header_t) + data_len;
-    uint8_t * data_buf = reinterpret_cast<uint8_t*>(pvPortMalloc(msg_len));
+    uint8_t * data_buf = static_cast<uint8_t*>(pvPortMalloc(msg_len));
```
victorsowa12, line 22: small fix I found

src/apps/bridge/app_config.h
```diff
@@ -12,6 +12,7 @@ constexpr const char *SUBSAMPLE_ENABLED = "subsampleEnabled";
 constexpr const char *BRIDGE_POWER_CONTROLLER_ENABLED = "bridgePowerControllerEnabled";
 constexpr const char *ALIGNMENT_INTERVAL_5MIN = "alignmentInterval5Min";
 constexpr const char *TRANSMIT_AGGREGATIONS = "transmitAggregations";
+constexpr const char *SAMPLES_PER_REPORT = "samplesPerReport";
```
victorsowa12, line 15: This isn't used by this PR , but will be used by the "report builder" which will come next. (This PR originally implemented this but took a drastic turn. Since it will eventually be needed I decided it was okay to leave it)

src/apps/bridge/aanderaaController.h
```diff
@@ -10,6 +10,15 @@
...
 #define DEFAULT_TRANSMIT_AGGREGATIONS 1
 
+typedef struct aanderaa_aggregations_s {
+  uint64_t node_id;
+  double abs_speed_mean_cm_s;
+  double abs_speed_std_cm_s;
+  double direction_circ_mean_rad;
+  double direction_circ_std_rad;
+  double temp_mean_deg_c;
+} aanderaa_aggregations_t;
```
victorsowa12, line 20: struct to send to the "report builder"

victorsowa12:
Need to test with a simulated aanderaa app - then I will mark it ready for review

towynlin (approved):
LGTM 🏃

PR #27, opened 2023-10-30
Calculate mean and std from same buf, no need to dup data

towynlin:


src/apps/bridge/aanderaaController.cpp
```diff
@@ -131,11 +125,11 @@ void Aanderaa::aanderaSubCallback(uint64_t node_id, const char *topic, uint16_t
...
-                              node_id, aanderaa->abs_speed_mean_cm_s.getMean(),
-                              aanderaa->abs_speed_std_cm_s.getStd(),
-                              aanderaa->direction_circ_mean_rad.getCircularMean(),
-                              aanderaa->direction_circ_std_rad.getCircularStd(),
-                              aanderaa->temp_mean_deg_c.getMean());
+                              node_id, aanderaa->abs_speed_cm_s.getMean(true),
+                              aanderaa->abs_speed_cm_s.getStd(true),
+                              aanderaa->direction_rad.getCircularMean(),
+                              aanderaa->direction_rad.getCircularStd(),
+                              aanderaa->temp_deg_c.getMean(true));
```
towynlin, line 132: Use Kahan summation per discussion in product spec

victorsowa12 (approved):

PR #28, opened 2023-11-02
Bridge saves SM config CRC list

towynlin:


src/lib/sys/configuration.cpp
```diff
@@ -741,6 +755,13 @@ bool Configuration::saveConfig(bool restart) {
                 if(cbor_value_get_string_length(&it,&size) != CborNoError){
                     return false;
                 }
+                break;
+            }
+            case ARRAY: {
```
russelldeguzman, line 760: We should also update `topology_sampler.cpp` : `encode_cbor_configuration` to support array types!

src/lib/sys/configuration.h
```diff
@@ -40,7 +42,7 @@ typedef struct ConfigPartition {
     ConfigValue_t values[MAX_NUM_KV];
 }__attribute__((packed, aligned(1))) ConfigPartition_t;
 
-class Configuration {
+class Configuration : public AbstractConfiguration {
```
russelldeguzman, line 45: 👍

src/lib/sys/configuration.cpp
```diff
@@ -337,6 +337,15 @@ uint8_t *Configuration::asCborMap(size_t &buffer_size) {
           err = cbor_encode_byte_string(&map, tmpB, tmpBSize);
         }
         break;
+      case ARRAY:
+        tmpBSize = MAX_STR_LEN_BYTES;
+        memset(tmpB, 0, MAX_STR_LEN_BYTES);
+        internalSuccess = getConfigCbor(key.keyBuffer, key.keyLen, tmpB, tmpBSize);
+        if (internalSuccess) {
+            memcpy(map.data.ptr, tmpB, tmpBSize);
+            map.data.ptr += tmpBSize;
```
towynlin, line 346: @russelldeguzman I wanted you to see this — figured out last night how to copy an already-encoded container in place without parsing while keeping the encoder state valid — it's really just [these two lines](https://github.com/sofarocean/tinycbor/blob/3cba6b11aaa0f6f674cd56ebaa573c4b65f71ee7/src/cborencoder.c#L298-L299), adapted from `append_to_buffer` in cborencoder.c!

russelldeguzman: Ah this is great!!

src/lib/sys/configuration.cpp
towynlin: Thanks for the tip! It didn't break the build with a "switch must handle all enum values" so I didn't see it before. Done in commit d7d621d. It was a little tricky since, in addition to knowing how to calculate on the internal pointer of the encoder like last night, I also had to find the internal pointer of the parser.

src/lib/common/topology_sampler.cpp
```diff
@@ -142,21 +143,14 @@ static void topology_sample_cb(networkTopology_t *networkTopology) {
...
-          printf("Failed to set crc in hwcfg\n");
-        }
+    SMConfigCRCList sm_config_crc_list(_hw_cfg);
+    bool known = sm_config_crc_list.contains(network_crc32_calc);
+    if (!known || _send_on_boot) {
+      if (!known) {
+        printf("The smConfigurationCrc is not in the known list! calc: 0x%" PRIx32
+               " Adding it.\n",
+               network_crc32_calc);
+        sm_config_crc_list.add(network_crc32_calc);
```
towynlin, line 153: Here's the actual usage of the new class changing in the topology sampler. Instead of storing a single CRC, we now use the list. This behavior is not final since we only want to add CRCs to the list when reporting to wavefleet, but it's the next step adaptation.

src/lib/common/topology_sampler.cpp
```diff
@@ -611,6 +607,23 @@ static bool encode_cbor_configuration(CborEncoder &array_encoder,
...
+          already_advanced = true;
+          err = cbor_value_advance(&it);
+          if (err != CborNoError) {
+            break;
+          }
+          size_t encoded_array_byte_len = it.source.ptr - array_start;
+          memcpy(map_encoder.data.ptr, array_start, encoded_array_byte_len);
+          map_encoder.data.ptr += encoded_array_byte_len;
+          break;
+        }
```
towynlin, line 626: Per your feedback @russelldeguzman here's the array copy in the topology sampler.

src/lib/sys/abstract_configuration.h
```diff
@@ -0,0 +1,17 @@
...
+namespace cfg {
+
+// Allows for mocking in unit tests
+class AbstractConfiguration {
+public:
+  virtual bool getConfigCbor(const char *key, size_t key_len, uint8_t *value,
+                             size_t &value_len) = 0;
+  virtual bool setConfigCbor(const char *key, size_t key_len, uint8_t *value,
+                             size_t value_len) = 0;
+};
```
towynlin, line 15: Obviously the real configuration class has many more methods. These two were all I needed for the test mocks though. If in the future we write more tests using the MockConfiguration and need more of the methods mocked, then they just need to be added both in the mock class and here.

russelldeguzman (approved):

PR #30, opened 2023-11-06
Sc 196701/cbor sensor encoder

russelldeguzman:
Adding a generic cbor sensor report library for a 3d array.
See example usage at the top of 
`src/apps/bridge/cbor_sensor_report_encoder.cpp`

Test:
Builds - but need an end-to-end data test once the aggregator is built


towynlin (changes requested):
I *love* this! 👏  Bravo! One language change please — within each sample are a number of "sample members". We should not call these "readings" because readings are the individual numbers before aggregation. I've also heard @evanShap refer to "data channels", but that's different too — we take both a mean and a standard deviation from one "data channel". Evan and I brainstormed a bit and think "sample member" is the best we can come up with. Please change every place in this PR that currently refers to "readings" to "member" or "sample member".

towynlin (approved):

PR #31, opened 2023-11-07
Sc 196635/report builder

victorsowa12:
The report builder gathers samples from the aanderaa controller, creates a corresponding element in a linked list, adds the data to element in an array of length `samplesPerReport`. Then after we reach samplesPerReport samples, the report builder searches through its linked list and looks for elements that correspond to the "max topology" during the "report period". The report builder tracks "max topology" by checking the network CRC periodically and updating the CRC and topology it will use to build the report if its current topology is smaller than the new one. If there is an element in the linked list that is also in the topology the array of samples will be copied into the report. If there is no element in the linked list then the report gets filled with NAN's. Once the entire report is built the report builder publishes the Cbor array to spotter via bm_serial. After publishing the report, the linked list is cleared and all of the network tracking (crc, "max topology") are cleared in the report builders local context to be re-created during the next report window.

Example data being sent out of Spotter:
```
343a0f3044f7c28,27.582,30.562,2.127,1.307,26.942
Current bit offset: 166
Rounded bit offset: 168
New bit offset: 552
10858606t [MS] [INFO] Added message(id: 102 len: 69) to queue MS_Q_LEGACY: (1)!
Message: E6 00 00 00 00 D5 AA A0 12 1D 16 D8 00 06 D4 00 01 00 F5 F8 80 81 81 85 FB 40 3B 95 04 54 EB 05 8F FB 40 3E 8F C7 68 0E 66 E9 FB 40 01 04 33 45 30 3F 0E FB 3F F4 E9 1E 81 A3 DA 3F FB 40 3A F1 2B A1 6E 7A 31
Submitted sensor data to sat/cell queue
```
the CBOR buffer is `81 81 85 FB 40 3B 95 04 54 EB 05 8F FB 40 3E 8F C7 68 0E 66 E9 FB 40 01 04 33 45 30 3F 0E FB 3F F4 E9 1E 81 A3 DA 3F FB 40 3A F1 2B A1 6E 7A 31` and can be
decoded on cbor.me to:
```
[[[27.582097346666668, 30.56163645125972, 2.127050915274487, 1.3069138588013087, 26.942072]]]
from the printout on spotter the means/stdevs:
343a0f3044f7c28, 27.582, 30.562, 2.127, 1.307, 26.942
```

### DIFF
--- a/src/apps/bridge/reportBuilder.cpp
+++ b/src/apps/bridge/reportBuilder.cpp
@@ -452,6 +452,8 @@ static void report_builder_task(void *parameters) {
               size_t temp_node_list_size = sizeof(temp_node_list);
               uint32_t temp_num_nodes;
               if (topology_sampler_get_node_list(temp_node_list, temp_node_list_size, temp_num_nodes, TOPO_TIMEOUT_MS)) {
+                // TODO - we need to use the cbor_map and build a report builder topology map that only
+                // includes aanderaa nodes
                 printf("Got topology in report builder!\n");
                 if (temp_num_nodes >= _ctx._report_period_num_nodes) {
                   printf("Updating CRC and topology in report builder!\n");
--- a/src/apps/bridge/reportBuilder.cpp
+++ b/src/apps/bridge/reportBuilder.cpp
@@ -19,8 +19,8 @@
 
 typedef struct report_builder_element_s {
   uint64_t node_id;
-  uint8_t sensor_type; // TODO - actually use it, it's ignored for now
-  aanderaa_aggregations_t *sensor_data; // TODO - make generic
+  uint8_t sensor_type;
+  void *sensor_data;
   report_builder_element_s *next;
   report_builder_element_s *prev;
 } report_builder_element_t;
@@ -34,9 +34,10 @@ static aanderaa_aggregations_t NAN_AGG = {.abs_speed_mean_cm_s = NAN,
 class ReportBuilderLinkedList {
   private:
 
-    report_builder_element_t* newElement(uint64_t node_id, uint8_t sensor_type, aanderaa_aggregations_t *sensor_data, uint32_t samples_per_report, uint32_t sample_counter);
+    report_builder_element_t* newElement(uint64_t node_id, uint8_t sensor_type, void *sensor_data, uint32_t sensor_data_size, uint32_t samples_per_report, uint32_t sample_counter);
     void freeElement(report_builder_element_s **element_pointer);
-    void append(report_builder_element_t *curr, uint64_t node_id, uint8_t sensor_type, aanderaa_aggregations_t *sensor_data, uint32_t samples_per_report, uint32_t sample_counter);
+    void append(report_builder_element_t *curr, uint64_t node_id, uint8_t sensor_type, void *sensor_data, uint32_t sensor_data_size, uint32_t samples_per_report, uint32_t sample_counter);
+    void addSensorDataToElement(report_builder_element_t *element, uint8_t sensor_type, void *sensor_data, uint32_t sample_counter);
 
     report_builder_element_t *head;
     size_t size;
@@ -47,7 +48,7 @@ class ReportBuilderLinkedList {
     ~ReportBuilderLinkedList();
 
     bool removeElement(report_builder_element_t *element);
-    bool addSample(uint64_t node_id, uint8_t sensor_type, aanderaa_aggregations_t *sensor_data, uint32_t samples_per_report, uint32_t sample_counter);
+    bool addSample(uint64_t node_id, uint8_t sensor_type, void *sensor_data, uint32_t sensor_data_size, uint32_t samples_per_report, uint32_t sample_counter);
     report_builder_element_t* findElement(uint64_t node_id);
     void clear();
     size_t getSize();
@@ -82,27 +83,18 @@ ReportBuilderLinkedList::~ReportBuilderLinkedList() {
   }
 }
 
-report_builder_element_t* ReportBuilderLinkedList::newElement(uint64_t node_id, uint8_t sensor_type, aanderaa_aggregations_t *sensor_data, uint32_t samples_per_report, uint32_t sample_counter) {
+report_builder_element_t* ReportBuilderLinkedList::newElement(uint64_t node_id, uint8_t sensor_type, void *sensor_data, uint32_t sensor_data_size, uint32_t samples_per_report, uint32_t sample_counter) {
   report_builder_element_t *element = static_cast<report_builder_element_t *>(pvPortMalloc(sizeof(report_builder_element_t)));
   configASSERT(element != NULL);
   element->node_id = node_id;
   element->sensor_type = sensor_type; // Ignored for now
   // TODO - use the sensor type to determine the size of the sensor data
-  element->sensor_data = static_cast<aanderaa_aggregations_t *>(pvPortMalloc(samples_per_report * sizeof(aanderaa_aggregations_t)));
+  element->sensor_data = static_cast<void *>(pvPortMalloc(samples_per_report * sensor_data_size));
   configASSERT(element->sensor_data != NULL);
-  memset(element->sensor_data, 0, samples_per_report * sizeof(aanderaa_aggregations_t));
+  memset(element->sensor_data, 0, samples_per_report * sensor_data_size);
 
-  // Back fill the sensor_data with NANs if we are not on sample_count 0.
-  uint32_t i;
-  for (i = 0; i < sample_counter; i++) {
-    memcpy(&element->sensor_data[i], &NAN_AGG, sizeof(aanderaa_aggregations_t));
-  }
-  // Copy the sensor data into the elements array in the correct location within the buffer
-  if (sensor_data != NULL) {
-    memcpy(&element->sensor_data[i], sensor_data, sizeof(aanderaa_aggregations_t));
-  } else {
-    memcpy(&element->sensor_data[i], &NAN_AGG, sizeof(aanderaa_aggregations_t));
-  }
+  addSensorDataToElement(element, sensor_type, sensor_data, sample_counter);
+
   element->next = NULL;
   element->prev = NULL;
   return element;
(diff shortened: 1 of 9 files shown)


src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,490 @@
...
+            if (temp_network_crc32 != _ctx._report_period_max_network_crc32) {
+              printf("Getting topology in report builder!\n");
+              uint64_t temp_node_list[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
+              size_t temp_node_list_size = sizeof(temp_node_list);
+              uint32_t temp_num_nodes;
+              if (topology_sampler_get_node_list(temp_node_list, temp_node_list_size, temp_num_nodes, TOPO_TIMEOUT_MS)) {
+                // TODO - we need to use the cbor_map and build a report builder topology map that only
+                // includes aanderaa nodes
+                printf("Got topology in report builder!\n");
+                if (temp_num_nodes >= _ctx._report_period_num_nodes) {
```
victorsowa12, line 522: I did a >= here since I thought we should change the report builders topology list reference if we swap out a node... although perhaps in most cases that should also be accompanied by someone turning on/off spotter?

src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,490 @@
...
+          uint8_t *last_network_config = topology_sampler_alloc_last_network_config(temp_network_crc32, temp_cbor_config_size);
+          if (last_network_config != NULL) {
+            printf("Got CRC %" PRIx32 " OLD CRC %" PRIx32 "\n", temp_network_crc32, _ctx._report_period_max_network_crc32);
+            if (temp_network_crc32 != _ctx._report_period_max_network_crc32) {
+              printf("Getting topology in report builder!\n");
+              uint64_t temp_node_list[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
+              size_t temp_node_list_size = sizeof(temp_node_list);
+              uint32_t temp_num_nodes;
+              if (topology_sampler_get_node_list(temp_node_list, temp_node_list_size, temp_num_nodes, TOPO_TIMEOUT_MS)) {
+                // TODO - we need to use the cbor_map and build a report builder topology map that only
```
russelldeguzman, line 519: Why do we need an aanderaa specific report builder topology?

src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,490 @@
...
+        case REPORT_BUILDER_INCREMENT_SAMPLE_COUNT: {
+          _ctx._sample_counter++;
+          printf("Incrementing sample counter to %d\n", _ctx._sample_counter);
+          if (_ctx._sample_counter >= _ctx._samplesPerReport) {
+            if (_ctx._samplesPerReport > 0 && _ctx._transmitAggregations) {
+              // Need to have more than one node to report anything (1 node would be just the bridge)
+              if (_ctx._report_period_num_nodes > 1) {
+                uint32_t num_sensors = _ctx._report_period_num_nodes - 1; // Minus one since topo includes the bridge
+                sensor_report_encoder_context_t context;
+                uint8_t cbor_buffer[315];
```
russelldeguzman: Please #define or constexpr `315`

russelldeguzman: This is also a fairly large buffer to pop onto the task stack - it looks like the stack depth is 1024 so this buffer consumes almost 1/3 of the stack here. 

Watch the watermark carefully or consider dynamic allocation or `static & memset(0)` so we don't get a stack overflow scenario.

src/apps/bridge/reportBuilder.cpp
victorsowa12: The behavior rn is that if there is a node we will add it to the report, even if it is not an aanderaa node. This means that a unit with a devkit will be sending extra data. I asked Evan and he said there will be some DVT2 customers using additional devkit nodes, or for our own testing the loadcell. 

I suppose we don't need an aanderaa specific topology, but more a map that associates a node_id in the topo list with it being an aanderaa, that way when the report builder loops throught the topology, we know when to look for/add a sensors samples to a report.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,490 @@
...
+                  }
+                  sensor_report_encoder_open_sensor(context, _ctx._samplesPerReport);
+                  for (uint32_t j = 0; j < _ctx._samplesPerReport; j++) {
+                    addSamplesToReport(context, element->sensor_type, element->sensor_data, j);
+                  }
+                  sensor_report_encoder_close_sensor(context);
+                }
+                sensor_report_encoder_close_report(context);
+                size_t cbor_buffer_len = sensor_report_encoder_get_report_size_bytes(context);
+                app_pub_sub_bm_bridge_sensor_report_data_t *message_buff = static_cast<app_pub_sub_bm_bridge_sensor_report_data_t *>(pvPortMalloc(sizeof(uint32_t) + sizeof(size_t) + cbor_buffer_len));
```
russelldeguzman: `sizeof(app_pub_sub_bm_bridge_sensor_report_data_t) + cbor_buffer_len`
instead of summing individual elements.

src/apps/bridge/reportBuilder.cpp
russelldeguzman: Ah I see, so we need something like a "isSupportedSensorSchema" or something where we can first add aanderaa and then eventually the RBR and soft?

victorsowa12: Thats one aspect. I think we also need report builder to save a dict with "node_id : isSupportedSensorSchema, ...." such that if a sensor did not generate any samples during a report period, but is a part of the "max topology" the report builder will know if it can add NAN's to the report, or if it should not add anything.

src/apps/bridge/reportBuilder.cpp
victorsowa12: Good point! I'll go with dynamic allocation since we may want to support multipart messages in the future.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,497 @@
...
+ * @param sensor_data Pointer to the sensor data for the report.
+ * @param sample_counter The current sample counter, indicating the position at which to add the sensor data.
+ */
+void ReportBuilderLinkedList::addSensorDataToElement(report_builder_element_t *element, uint8_t sensor_type, void *sensor_data, uint32_t sample_counter) {
+    switch (sensor_type) {
+      case AANDERAA_SENSOR_TYPE: {
+        // Back fill the sensor_data with NANs if we are not on the right sample count
+        uint32_t i;
+        for (i = 0; i < sample_counter; i++) {
+          memcpy(&(static_cast<aanderaa_aggregations_t *>(element->sensor_data))[i], &NAN_AGG, sizeof(aanderaa_aggregations_t));
```
russelldeguzman: Nit: Not super necessary now, but maybe makes sense to have size_t sensor_data_size as an argument. Maybe just a  TODO comment is fine for now so you don't have to change all the calls.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,497 @@
...
+  } else {
+    if (head == NULL && size == 0) {
+      append(NULL, node_id, sensor_type, sensor_data, sensor_data_size, samples_per_report, sample_counter);
+      rval = true;
+    } else {
+      report_builder_element_t *element = head;
+      while (element->next != NULL) {
+        element = element->next;
+      }
+      append(element, node_id, sensor_type, sensor_data, sensor_data_size, samples_per_report, sample_counter);
```
russelldeguzman, line 206: do we need an `rval = true` after this append? is there ever a scenario where this is false?

src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+static void report_builder_task(void *parameters);
+
+ReportBuilderLinkedList::ReportBuilderLinkedList() {
+  head = NULL;
+  size = 0;
+}
+
+ReportBuilderLinkedList::~ReportBuilderLinkedList() {
+  report_builder_element_t *element = head;
```
russelldeguzman: Nit: could we just call `ReportBuilderLinkedList::clear()` here?

src/apps/bridge/reportBuilder.cpp
victorsowa12: Talked IRL. Originally the `sensor_data_size` was an input, but then I realized that this function needs to know how to cast the void * sensor data and thus needs to already know the varying sensor data structs, so then it can extract the size of the structs from those instead of needing an input.

src/apps/bridge/reportBuilder.cpp
victorsowa12: yup!

src/apps/bridge/reportBuilder.cpp
victorsowa12: Nope, also the functions that `findElementAndAddSampleToElement` call don't return anything so I'll also make this function a void return type.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,497 @@
...
+          printf("Incrementing sample counter to %d\n", _ctx._sample_counter);
+          if (_ctx._sample_counter >= _ctx._samplesPerReport) {
+            if (_ctx._samplesPerReport > 0 && _ctx._transmitAggregations) {
+              // Need to have more than one node to report anything (1 node would be just the bridge)
+              if (_ctx._report_period_num_nodes > 1) {
+                uint32_t num_sensors = _ctx._report_period_num_nodes - 1; // Minus one since topo includes the bridge
+                sensor_report_encoder_context_t context;
+                uint8_t *cbor_buffer = static_cast<uint8_t *>(pvPortMalloc(MAX_SENSOR_REPORT_CBOR_LEN));
+                configASSERT(cbor_buffer != NULL);
+                sensor_report_encoder_open_report(cbor_buffer, sizeof(cbor_buffer), num_sensors, context);
```
russelldeguzman: We should probably structure this function a bit differently (maybe do/while?) to check the return codes of `sensor_report_encoder_xxx`. This function unlikely fail to right now, but once we start adding more functionality / sensors / etc, we have a chance of failing and will try to operate on a borked cbor report.

victorsowa12: Great idea! If we fail to encode part of the message should we still try to send what we were able to encode? (in the case we run out of space in our buffer)

russelldeguzman: Good question, depending on where we fail to encode we could -  but it also may be invalid cbor at that point. I think for now it's probably fine to abort the message on an encoding failure for DVT2, but maybe @towynlin or Evan has more insight on how much of an effort we should make to send things out.

towynlin: Yeah, for DVT2 let's just abort when we're not sure what's happening. With more data gathered in the field, we'll learn what we should handle differently.

towynlin (approved):

src/apps/bridge/reportBuilder.cpp
```diff
@@ -0,0 +1,547 @@
...
+ * @param sensor_data Pointer to the sensor data for the report.
+ * @param sample_index The index of the sensor data to be copied into the sensor report.
+ */
+static bool addSamplesToReport(sensor_report_encoder_context_t &context, uint8_t sensor_type, void *sensor_data, uint32_t sample_index) {
+  bool rval = false;
+  switch (sensor_type){
+    case AANDERAA_SENSOR_TYPE: {
+      aanderaa_aggregations_t aanderaa_sample = (static_cast<aanderaa_aggregations_t *>(sensor_data))[sample_index];
+      if (sensor_report_encoder_open_sample(context, AANDERAA_NUM_SAMPLE_MEMBERS) != CborNoError) {
+        BRIDGE_LOG_PRINT("Failed to open sample in addSamplesToReport\n");
```
russelldeguzman, line 308: I think we need to break in these statements if the encoding fails.

russelldeguzman (approved):
One more comment, but otherwise it looks good!

src/apps/bridge/reportBuilder.cpp
victorsowa12: good catch yup!

PR #32, opened 2023-11-07
set and get the last known config info

russelldeguzman:
Store a  "last-known" crc32 and config for retrieval. 

Protected by the topo sampler mutex.


towynlin (approved):
Great!

Per the architecture doc, we need to specifically control when the last network config is updated rather than updating every time the topology changes, but that can be in a different PR.

victorsowa12 (approved):

PR #35, opened 2023-11-10
fix: bm cfg of non-neighbors with new link-local forwarding

towynlin:
Prior to this PR, given a topology of 3 nodes A | B | C, running on node A any `bm cfg` command that targeted node C would fail. The reason was that config commands are sent to the link-local multicast address, and node B, upon finding that it was not the target of the message, simply dropped the packet.

This PR adds the ability in BCMP to forward messages in that scenario, using a feature described in section 5.4.4.2 of the Bristlemouth spec that had not previously been implemented.

> One additional requirement relating to the link-local neighbors-only address, which can help cut down on unnecessary transmissions, is that granular messaging limited to a specific Port can be achieved by setting a destination address with the pattern:
> 
> ff02::000X:1
> 
> Wherein X represents the egress port on which to direct the packet (as opposed to all available ports by default). As with the equivalent global all-nodes address pattern, these four bits shall be cleared before transmission on the relevant port.
> 
> This addressing feature can help to support more efficient execution of higher-level Bristlemouth functions relating to neighbor discovery, capability negotiation, and resource discovery, as these exchanges carry information associated with specific neighbors and network segments, which may not be useful or relevant to nodes on other segments.

---

Other minor changes:

- Renaming `src_port` to `egress_port` and `dst_port` to `ingress_port` for clarity using the unambiguous official terminology from the standard
- Noting in TODO comments some places where the current implementation does not conform to the published standard
- Noting in TODO comments that BCMP assumes a 2-port PHY in this forwarding implementation and that future effort should generalize
- Some clang-format cleanup

### DIFF
--- a/src/lib/bcmp/bcmp.cpp
+++ b/src/lib/bcmp/bcmp.cpp
@@ -99,16 +99,17 @@ static void dfu_copy_and_process_message(struct pbuf *pbuf) {
 */
 int32_t bmcp_process_packet(struct pbuf *pbuf, ip_addr_t *src, ip_addr_t *dst) {
   int32_t rval = 0;
-  // uint8_t src_port;
-  uint8_t dst_port;
+  // uint8_t egress_port;
+  uint8_t ingress_port;
 
   configASSERT(pbuf);
   configASSERT(src);
 
-   // Ingress and Egress ports are mapped to the 5th and 6th byte of the IPv6 src address
-   // as per the bristlemouth protocol spec */
-  // src_port = ((src->addr[1]) & 0xFF);
-  dst_port = ((src->addr[1] >> 8) & 0xFF);
+  // Ingress and Egress ports are mapped to the 5th and 6th byte of the IPv6 src address.
+  // This is in conflict with the v1.0.0, June 2023 version of the Bristlemouth spec.
+  // TODO: align with spec
+  // egress_port = ((src->addr[1]) & 0xFF);
+  ingress_port = ((src->addr[1] >> 8) & 0xFF);
 
   CLEAR_PORTS(src->addr);
 
@@ -133,7 +134,7 @@ int32_t bmcp_process_packet(struct pbuf *pbuf, ip_addr_t *src, ip_addr_t *dst) {
     switch(header->type) {
       case BCMP_HEARTBEAT: {
         // Send out heartbeats
-        bcmp_process_heartbeat(reinterpret_cast<bcmp_heartbeat_t *>(header->payload), src, dst_port);
+        bcmp_process_heartbeat(reinterpret_cast<bcmp_heartbeat_t *>(header->payload), src, ingress_port);
         break;
       }
 
@@ -182,7 +183,12 @@ int32_t bmcp_process_packet(struct pbuf *pbuf, ip_addr_t *src, ip_addr_t *dst) {
       case BCMP_CONFIG_STATUS_RESPONSE:
       case BCMP_CONFIG_DELETE_REQUEST:
       case BCMP_CONFIG_DELETE_RESPONSE: {
-        bcmp_process_config_message(static_cast<bcmp_message_type_t>(header->type), header->payload);
+        bool should_forward = bcmp_process_config_message(
+            static_cast<bcmp_message_type_t>(header->type), header->payload);
+        if (should_forward) {
+          // Forward the message to all ports other than the ingress port.
+          bcmp_ll_forward(pbuf, ingress_port);
+        }
         break;
       }
 
@@ -398,6 +404,37 @@ err_t bcmp_tx(const ip_addr_t *dst, bcmp_message_type_t type, uint8_t *buff, uin
     }
   } while(0);
 
+  return rval;
+}
+
+/*! \brief Forward the payload to all ports other than the ingress port.
+
+    See section 5.4.4.2 of the Bristlemouth spec for details.
+
+    \param[in] pbuf Packet buffer to forward.
+    \param[in] ingress_port Port on which the packet was received.
+    \return ERR_OK on success, or an error that occurred.
+*/
+err_t bcmp_ll_forward(struct pbuf *pbuf, uint8_t ingress_port) {
+  ip_addr_t port_specific_dst;
+  ip6_addr_set(&port_specific_dst, &multicast_ll_addr);
+
+  // TODO: Make more generic. This is specifically for a 2-port device.
+  uint8_t egress_port = ingress_port == 1 ? 2 : 1;
+  port_specific_dst.addr[3] = 0x1000000 | (egress_port << 8);
+
+  const ip_addr_t *src = netif_ip_addr6(_ctx.netif, 0);
+
+  // L2 will clear the egress port from the destination address,
+  // so calculate the checksum on the link-local multicast address.
+  bcmp_header_t *header = static_cast<bcmp_header_t *>(pbuf->payload);
+  header->checksum = 0;
+  header->checksum = ip6_chksum_pseudo(pbuf, IP_PROTO_BCMP, pbuf->len, src, &multicast_ll_addr);
+
+  err_t rval = raw_sendto_if_src(_ctx.pcb, pbuf, &port_specific_dst, _ctx.netif, src);
+  if (rval != ERR_OK) {
+    printf("Error forwarding BMCP packet link-locally: %d\n", rval);
+  }
   return rval;
 }
 
--- a/src/lib/bcmp/bcmp_config.cpp
+++ b/src/lib/bcmp/bcmp_config.cpp
@@ -389,10 +389,15 @@ static void bcmp_process_del_response_message(bm_common_config_delete_key_respon
     vPortFree(keyprintbuf);
 }
 
-void bcmp_process_config_message(bcmp_message_type_t bcmp_msg_type, uint8_t* payload) {
+/*!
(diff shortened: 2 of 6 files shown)


src/lib/bcmp/bcmp.cpp
```diff
@@ -99,16 +99,17 @@ static void dfu_copy_and_process_message(struct pbuf *pbuf) {
...
 
-   // Ingress and Egress ports are mapped to the 5th and 6th byte of the IPv6 src address
-   // as per the bristlemouth protocol spec */
-  // src_port = ((src->addr[1]) & 0xFF);
-  dst_port = ((src->addr[1] >> 8) & 0xFF);
+  // Ingress and Egress ports are mapped to the 5th and 6th byte of the IPv6 src address.
+  // This is in conflict with the v1.0.0, June 2023 version of the Bristlemouth spec.
+  // TODO: align with spec
+  // egress_port = ((src->addr[1]) & 0xFF);
+  ingress_port = ((src->addr[1] >> 8) & 0xFF);
```
towynlin, line 112: The spec says bits 108:111 of the source IP address indicates the ingress port. This implementation is using too many bits (8 instead of 4) and in the wrong place (bits 80:87). Similarly, the implementation uses bits 88-95 for the egress port instead of 104:107 in the spec.

src/lib/bcmp/bcmp_config.cpp
```diff
@@ -389,10 +389,15 @@ static void bcmp_process_del_response_message(bm_common_config_delete_key_respon
...
+*/
+bool bcmp_process_config_message(bcmp_message_type_t bcmp_msg_type, uint8_t *payload) {
+    bool should_forward = false;
     do {
         bm_common_config_header_t * msg_header = reinterpret_cast<bm_common_config_header_t *>(payload);
-        if(msg_header->target_node_id != getNodeId()){
+        if (msg_header->target_node_id != getNodeId()) {
+            should_forward = true;
             break;
         }
```
towynlin, line 402: Instead of just dropping the packet, let's return true to indicate BCMP should forward the message.

src/lib/bcmp/bcmp.cpp
```diff
@@ -182,7 +183,12 @@ int32_t bmcp_process_packet(struct pbuf *pbuf, ip_addr_t *src, ip_addr_t *dst) {
       case BCMP_CONFIG_STATUS_RESPONSE:
       case BCMP_CONFIG_DELETE_REQUEST:
       case BCMP_CONFIG_DELETE_RESPONSE: {
-        bcmp_process_config_message(static_cast<bcmp_message_type_t>(header->type), header->payload);
+        bool should_forward = bcmp_process_config_message(
+            static_cast<bcmp_message_type_t>(header->type), header->payload);
+        if (should_forward) {
+          // Forward the message to all ports other than the ingress port.
+          bcmp_ll_forward(pbuf, ingress_port);
+        }
```
towynlin, line 191: If the message can't be processed at this node, forward it, calling the new `bcmp_ll_forward` function.

src/lib/bcmp/bcmp.cpp
```diff
@@ -401,6 +407,37 @@ err_t bcmp_tx(const ip_addr_t *dst, bcmp_message_type_t type, uint8_t *buff, uin
...
 
+/*! \brief Forward the payload to all ports other than the ingress port.
+
+    See section 5.4.4.2 of the Bristlemouth spec for details.
+
+    \param[in] pbuf Packet buffer to forward.
+    \param[in] ingress_port Port on which the packet was received.
+    \return ERR_OK on success, or an error that occurred.
+*/
+err_t bcmp_ll_forward(struct pbuf *pbuf, uint8_t ingress_port) {
```
towynlin, line 418: New function callable by the BCMP layer

src/lib/bcmp/bm/bm_l2.cpp
```diff
@@ -93,8 +93,11 @@ static void bm_l2_process_tx_evt(l2_queue_element_t *tx_evt) {
     for (uint32_t idx=0; idx < BM_NETDEV_TYPE_MAX; idx++) {
         switch (bm_l2_ctx.devices[idx].type) {
             case BM_NETDEV_TYPE_ADIN2111: {
-                err_t retv = adin2111_tx((adin2111_DeviceHandle_t) bm_l2_ctx.devices[idx].device_handle, static_cast<uint8_t *>(tx_evt->pbuf->payload), tx_evt->pbuf->len,
-                                   (tx_evt->port_mask >> mask_idx) & ADIN2111_PORT_MASK, bm_l2_ctx.devices[idx].start_port_idx);
+                err_t retv =
+                    adin2111_tx((adin2111_DeviceHandle_t)bm_l2_ctx.devices[idx].device_handle,
+                                static_cast<uint8_t *>(tx_evt->pbuf->payload), tx_evt->pbuf->len,
+                                (tx_evt->port_mask >> mask_idx) & ADIN2111_PORT_MASK,
+                                bm_l2_ctx.devices[idx].start_port_idx);
```
towynlin, line 100: Only a formatting change

src/lib/bcmp/bm/bm_l2.cpp
```diff
@@ -354,8 +357,24 @@ err_t bm_l2_rx(void* device_handle, uint8_t* payload, uint16_t payload_len, uint
...
+    constexpr uint8_t num_adin_ports = 2; // TODO generalize
+    if (egress_port > 0 && egress_port <= num_adin_ports) {
+      port_mask = 1 << (egress_port - 1);
+    }
+
+    // clear the egress port set by the application
+    eth_frame[egress_port_idx] = 0;
+
+    return bm_l2_tx(pbuf, port_mask);
 }
```
towynlin, line 378: This is the callback at layer 2, called by LWIP after the ethernet frame header has been added. This:

1. inspects the egress port added to the destination IP address by `bcmp_ll_forward`
2. uses that to build a port mask
3. clears those bits in the destination IP address so that we send to the actual multicast group that other nodes are subscribed to
4. and sends the packet with the port mask to the ADIN2111 via `bm_l2_tx`.

src/lib/bcmp/bm/bm_l2.cpp
```diff
@@ -354,8 +357,24 @@ err_t bm_l2_rx(void* device_handle, uint8_t* payload, uint16_t payload_len, uint
...
+    // if the application set an egress port, send only to that port
+    constexpr size_t bcmp_egress_port_offset_in_dest_addr = 13;
+    const size_t egress_port_idx = sizeof(struct eth_hdr) + offsetof(struct ip6_hdr, dest) +
+                                   bcmp_egress_port_offset_in_dest_addr;
+    uint8_t *eth_frame = static_cast<uint8_t *>(pbuf->payload);
+    uint8_t egress_port = eth_frame[egress_port_idx];
+    constexpr uint8_t num_adin_ports = 2; // TODO generalize
+    if (egress_port > 0 && egress_port <= num_adin_ports) {
+      port_mask = 1 << (egress_port - 1);
+    }
```
victorsowa12, line 372: Question for my own understanding, here we will set the `port_mask = 1 << (egress_port - 1)` , where egress_port will be 1 or 2 meaning `port_mask` will equal  0b01 or 0b10 and these correspond to the two different ports?

victorsowa12 (approved):
Looks good!

src/lib/bcmp/bm/bm_l2.cpp
towynlin: You've got it exactly. The field that comes in via the IP address is a number (1 or 2), and the ADIN tx function needs a bit mask of which ports to send on, where port 1 is the LSB (0b01), port 2 is the next bit (0b10). If both bits are set (0b11, the default) then the ADIN transmits on both ports.

src/lib/bcmp/bcmp.cpp
russelldeguzman: When we do update this, sounds like it'll be a breaking release? i.e. folks will need to update all BM nodes in their network?

src/lib/bcmp/bcmp.cpp
russelldeguzman: I wonder if we need to do this for all places we have `multicast_ll_addr`

I see references currently in: 
`bcmp_heartbeat`
`bcmp_neighbors`
`bcmp_resource_discovery`
`bcmp_time` 

Maybe doesn't need to be this PR but assuming that they suffer from the same behavior?

russelldeguzman (approved):
👍 

Looks great!
As per above, we should probably cut a ticket for 
`bcmp_resource_discovery`
`bcmp_time`
if not addressed in this PR.

src/lib/bcmp/bcmp.cpp
russelldeguzman: Actually I don't think heartbeat or neighbors need to change?
But time and resource_discovery likely have the same issue.

src/lib/bcmp/bcmp.cpp
towynlin: Yep that's true, so definitely before v1.0 and preferably sooner rather than later. I'd advocate for making this change at the same time as the planned release next year of generalized sensor networks, since everyone will have to update at that time anyway.

src/lib/bcmp/bcmp.cpp
towynlin: Good call. I'll create a new story for resource discovery and time now and can code them up today actually in a separate PR.

PR #39, opened 2023-11-13
Sc 197050/remove current agg period min from bridge

victorsowa12:
Remove the current agg period min config from the bridge and instead us the sample duration to calculate buffer sizes. Also use the sample period end in the bridgePowerController to send a task notification to the aanderaaController to end and aggregation period and send samples to the report builder.

### DIFF
--- a/src/apps/bridge/aanderaaController.cpp
+++ b/src/apps/bridge/aanderaaController.cpp
@@ -19,10 +19,7 @@
 #define MIN_READINGS_FOR_AGGREGATION 3
 #define DEFAULT_CURRENT_READING_PERIOD_MS 60 * 1000 // default is 1 minute: 60,000 ms
 
-typedef enum {
-  SAMPLER_TIMER_BITS = 0x01,
-  AGGREGATION_TIMER_BITS = 0x02,
-} aanderaaControllerBits_t;
+TaskHandle_t aanderaa_controller_task_handle = NULL;
 
 typedef struct Aanderaa {
   uint64_t node_id;
@@ -84,7 +81,7 @@ static void createAanderaaSub(uint64_t node_id);
 static void runController(void *param);
 static Aanderaa_t *findAanderaaById(uint64_t node_id);
 static void sampleTimerCallback(TimerHandle_t timer);
-static void aggregationTimerCallback(TimerHandle_t timer);
+// static void aggregationTimerCallback(TimerHandle_t timer);
 static bool node_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
                                const char *service, size_t reply_len, uint8_t *reply_data);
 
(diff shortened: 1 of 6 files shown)


towynlin:
I think this looks good, but I'd love it if you could teach us what's going on by adding a few PR comments.

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -227,6 +228,7 @@ void BridgePowerController::_update(
         if (nextSampleEpochS > currentCycleS) {
           powerBusAndSetSignal(false);
         }
+        xTaskNotify(aanderaa_controller_task_handle, AGGREGATION_TIMER_BITS, eSetBits);
```
victorsowa12, line 231: Whenever a sample ends we will tell the aanderaaController to compute mean/stdev.

src/apps/bridge/aanderaaController.cpp
```diff
@@ -298,11 +272,10 @@ static void createAanderaaSub(uint64_t node_id) {
 
   new_sub->node_id = node_id;
   new_sub->next = NULL;
-  new_sub->current_agg_period_ms = (DEFAULT_CURRENT_AGG_PERIOD_MIN * 60 * 1000);
-  uint32_t agg_period_min;
-  if (_ctx._usr_cfg->getConfig("currentAggPeriodMin", strlen("currentAggPeriodMin"),
-                               agg_period_min)) {
-    new_sub->current_agg_period_ms = (agg_period_min * 60 * 1000);
```
victorsowa12, line 339: this config "currentAggPeriodMin" controlled when the aanderaa controller would compute the mean/stdev. This isn't connected at all to the sample duration and thus could be completely different. (i.e. multiple aggregations per sample duration, or multiple sample durations per aggregation. From "[The Bridge shouldn’t have its own currentAggPeriodMin setting - it should just be using the sample duration / sample interval to drive aggregation timing. For now and foreseeable future wavefleet and other consumers of the data will be expecting 1::1 aggregation::sample period.](https://sofarocean.slack.com/archives/C01B91FGDR8/p1699404547576279)" this config needed to be removed and instead the mean/ stdev calculation trigger needs to be at the end of a sample duration.

src/apps/bridge/aanderaaController.cpp
```diff
@@ -298,11 +272,10 @@ static void createAanderaaSub(uint64_t node_id) {
...
+  new_sub->current_agg_period_ms = (BridgePowerController::DEFAULT_SAMPLE_DURATION_S * 1000);
+  if (_ctx._sys_cfg->getConfig(AppConfig::SAMPLE_DURATION_MS, strlen(AppConfig::SAMPLE_DURATION_MS),
+                               new_sub->current_agg_period_ms)) {
+
```
victorsowa12, line 312: since we are now using the sample duration we need to pull that config to calculate the proper buffer size.

src/apps/bridge/aanderaaController.h
```diff
@@ -18,8 +18,14 @@ typedef struct aanderaa_aggregations_s {
   double temp_mean_deg_c;
 } aanderaa_aggregations_t;
 
+typedef enum {
+  SAMPLER_TIMER_BITS = 0x01,
+  AGGREGATION_TIMER_BITS = 0x02,
+} aanderaaControllerBits_t;
```
victorsowa12, line 24: needed to move this to the header so that the bridgePowerController can use it.

src/apps/bridge/bridgePowerController.cpp
victorsowa12: This used to be done by a software timer with a period of "currentAggPeriodMin", but the time that the aanderaa controller computes its means/stdev needs to be tied to the end of a sample, which is done here by notifying the aanderaaController app that it is time to compute its means/stdevs.

victorsowa12:
@towynlin I added some comments. LMK if you have more questions!

towynlin (approved):
Thank you for the explanatory comments! Very helpful!

PR #40, opened 2023-11-14
Sc 197055/min max agg checks

victorsowa12:
Add min/max checks for the current meter sample_members. The initial sample_members will be NAN, and if they are within the valid range, they will be set, otherwise they will stay NAN.

### DIFF
--- a/src/apps/bridge/aanderaaController.cpp
+++ b/src/apps/bridge/aanderaaController.cpp
@@ -70,7 +70,7 @@ static constexpr uint32_t SAMPLE_TIMER_MS = 30 * 1000;
 static constexpr uint32_t TOPO_TIMEOUT_MS = 10 * 1000;
 static constexpr uint32_t NODE_INFO_TIMEOUT_MS = 1000;
 
-#define PI 3.14159265358979323846
+#define PI  3.14159265358979323846
 
 static constexpr double DIRECTION_SAMPLE_MIN = 0.0;
 static constexpr double DIRECTION_SAMPLE_MAX = 2*PI;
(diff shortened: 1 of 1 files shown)


src/apps/bridge/aanderaaController.cpp
```diff
@@ -69,6 +69,16 @@ static constexpr uint32_t SAMPLE_TIMER_MS = 30 * 1000;
...
 static constexpr uint32_t NODE_INFO_TIMEOUT_MS = 1000;
 
+#define PI  3.14159265358979323846
+
+static constexpr double DIRECTION_SAMPLE_MIN = 0.0;
+static constexpr double DIRECTION_SAMPLE_MAX = 2*PI;
+static constexpr double ABS_SPEED_SAMPLE_MIN = 0.0;
+static constexpr double ABS_SPEED_SAMPLE_MAX = 300.0;
+static constexpr double TEMP_SAMPLE_MIN = -5.0;
+static constexpr double TEMP_SAMPLE_MAX = 40.0;
```
towynlin: Change these from SAMPLE to READING.

src/apps/bridge/aanderaaController.cpp
```diff
@@ -240,11 +250,21 @@ static void runController(void *param) {
...
             // TODO - verify that we can assume if one sampler is below the min then all of them are.
             if(curr->abs_speed_cm_s.getNumSamples() >= MIN_READINGS_FOR_AGGREGATION) {
-              agg.abs_speed_mean_cm_s = curr->abs_speed_cm_s.getMean(true);;
-              agg.abs_speed_std_cm_s = curr->abs_speed_cm_s.getStd(true);;
-              agg.direction_circ_mean_rad = curr->direction_rad.getCircularMean();;
-              agg.direction_circ_std_rad = curr->direction_rad.getCircularStd();
-              agg.temp_mean_deg_c = curr->temp_deg_c.getMean(true);;
+              if (curr->abs_speed_cm_s.getMean(true) > ABS_SPEED_SAMPLE_MIN && curr->abs_speed_cm_s.getMean(true) < ABS_SPEED_SAMPLE_MAX) {
+                agg.abs_speed_mean_cm_s = curr->abs_speed_cm_s.getMean(true);
+              }
```
towynlin: The functions like `getMean` do a potentially lengthy calculation. We don't want to call it 3 times, tripling the time complexity of this part of the code. Instead, call it once, saving the value in a variable, like the previous removed code would actually be fine. And then check that saved variable against the max and min, setting to NaN if outside.

towynlin (changes requested):

src/apps/bridge/aanderaaController.cpp
victorsowa12: Good call! I will change it!

src/apps/bridge/aanderaaController.cpp
victorsowa12: OOH yeah, these shouldn't be SAMPLE. SAMPLE_MEMBER may be more appropriate since these are being used to check the values that are going into the sample and they are not being used to check each individual reading?

towynlin: It's debatable since ideally these ought to be the ranges for the readings as well as the sample members... but since we saw that the aanderaa was giving negative direction readings and can translate that into the sample, I take it back — let's just leave these as-is. "Sample member" here just feels unnecessarily verbose.

towynlin: Oh I see you already changed it. Cool cool, thanks, that works. 👍

towynlin (approved):

PR #45, opened 2023-11-15
Add an example app using the TSYS01

victorsowa12:
Added an example app for the TSYS01 temperature sensor on the BM_SPI lines. These lines overlap with the USART3 lines, which are enabled as USART3 in bm_mote_v1.0 bsp. So I made a new BSP that enables these lines as SPI1.

The new app is similar to the bm_mote serial payload example app, except it just uses the TSYS driver to init, check PROM and then loop read the sensor. I added a config to change the delay of the loop so EE team can configure for their needs.

Most files are auto generated and can be skipped, the majority of the apps files were copied over, the new file worth looking at is the new user_code.cpp, where I added some comments. Also the TSYS01.cpp file changed a little bit from the Spotter repo, I added a comment explaining the difference.

Note: all the other peripherals are still configured (ex: lpuart) but most are not used, so if it makes sense to turn them off/remove them from the BSP that can be done.  (extra special note: since lpuart is still configured in the bsp, but we no longer have a usart in this configuration, we need to manually prevent CubeMX from deleting the usart hal files. otherwise serial.c will complain)


src/apps/bm_soft_module/user_code/user_code.cpp
```diff
@@ -0,0 +1,74 @@
+#include "user_code.h"
+#include "bm_network.h"
+#include "bm_printf.h"
+#include "bm_pubsub.h"
```
victorsowa12, line 4: new file to look at. This is a modified serial payload example, with all the serial stuff removed and just the soft sensor sampling and printing.

src/apps/bm_soft_module/user_code/user_code.cpp
```diff
@@ -0,0 +1,74 @@
...
+  } else {
+    printf("SOFT Serial Number: Get SN Failed\n");
+  }
+  uint32_t soft_delay_s = 0;
+  if (userConfigurationPartition->getConfig("softReadingPeriodS", strlen("softReadingPeriodS"), soft_delay_s)) {
+    soft_delay = soft_delay_s * 1000;
+    printf("SOFT Delay: %" PRIu32 "ms\n", soft_delay);
+  } else {
+    printf("SOFT Delay: Using default % " PRIu32 "ms\n", soft_delay);
+  }
```
victorsowa12: grab a config to set the period at which we read the sensor

src/lib/drivers/TSYS01.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "task.h"
+#include "TSYS01.h"
+#include "debug.h"
+
+TSYS01::TSYS01(SPIInterface_t* spiInterface, IOPinHandle_t *cspin):
+_SPIInterface(spiInterface),_CSPin(cspin)
+{
+  IOWrite(_CSPin, 1);
+}
+
```
victorsowa12: Copied over from spotter repo, all "add errors to queues" were replaced with printfs. Nothing else was changed

src/lib/drivers/TSYS01.h
```diff
@@ -0,0 +1,61 @@
+//
+// Created by Genton Mo on 2019-07-19.
+// Ported by Alvaro
+//
+
+#ifndef WS_APPS_TSYS01_H
+#define WS_APPS_TSYS01_H
+
+#include "stm32u5xx_hal.h"
+#include "io.h"
```
victorsowa12: nothing changed in this header

towynlin (approved):

russelldeguzman (approved):

PR #47, opened 2023-11-16
fix: add "N Readings" as a column to AAND_AGG.csv

towynlin:


src/apps/bridge/aanderaaController.cpp
```diff
@@ -252,13 +252,15 @@ static void runController(void *param) {
             log_buflen = snprintf(log_buf, SENSOR_LOG_BUF_SIZE,
                               "%s,"
```
russelldeguzman: Mind adding a comment for "timemstamp" here? I forgot 😛

russelldeguzman (approved):

src/apps/bridge/aanderaaController.cpp
towynlin: Oh yeah, good call, totally, will do.

PR #55, opened 2023-11-28
Sc 195977/add bringup uart configurability

victorsowa12:
Adding the ability to change the baud of the USARTs/LPUART when using debug_uart.c. This is to allow the testing of SDI-12 with the bringup_mote app. Furthermore, to support this I added the ability to set the rx pin level high/low, as it needs to be low(inverted) for SDI-12. Lastly, I got rid of the configAssert when the RX line buffer overflows. Instead I added a print statement, clear the memory and reset the index to 0. 

### Example back and forth via SDI-12 on two bringup_mote's

Also note the gpio pins that need to be set!
Sender gpios:
 - `gpio clr vbus_5v_en`
 - `gpio clr bf_5v_en`
 
Receiver Gpios:
 - `gpio clr vbus_5v_en`
 - `gpio clr bf_5v_en`
 - `gpio set bf_sdi12_oe`

Serial settings:
- `serial rxLevel lpuart low`
- `serial baud lpuart 1200`
- `serial enable lpuart`
  
Screen Shot 2023-11-28 at 2 51 55 PM

### Trace from SDI-12 transaction (Also traced the Tx/RX lines of the UART)
Screen Shot 2023-11-28 at 2 54 10 PM
### Showing the settings of the logic analyzer
Screen Shot 2023-11-28 at 2 54 28 PM

## Example buffer overflow:
Screen Shot 2023-11-28 at 3 05 19 PM

### DIFF
--- a/src/lib/debug/debug_uart.c
+++ b/src/lib/debug/debug_uart.c
@@ -306,6 +306,8 @@ static BaseType_t debugSerialCommand(char *writeBuffer,
                       3, // Get the rest of the parameters
                       &parameterStringLength);
 
+      LL_USART_Disable((USART_TypeDef *)handle->device);
+
       if (strncmp("high", parameter, parameterStringLength) == 0) {
         LL_USART_SetRXPinLevel((USART_TypeDef *)handle->device, LL_USART_RXPIN_LEVEL_STANDARD);
       } else if (strncmp("low", parameter, parameterStringLength) == 0) {
@@ -314,6 +316,8 @@ static BaseType_t debugSerialCommand(char *writeBuffer,
         printf("ERR Invalid paramters\n");
         break;
       }
+
+      LL_USART_Enable((USART_TypeDef *)handle->device);
       printf("OK\n");
     } else if (strncmp("baud", parameter, parameterStringLength) == 0) {
       BaseType_t handle_string_len;
--- a/src/lib/debug/debug_uart.c
+++ b/src/lib/debug/debug_uart.c
@@ -172,7 +172,10 @@ static const CLI_Command_Definition_t cmdSerial = {
   "serial:\n"
   " * serial list\n"
   " * serial enable <interface>\n"
-  " * serial tx <interface> <message>\n",
+  " * serial tx <interface> <message>\n"
+  " * serial baud <baudrate>\n"
+  " * serial rxLevel <interface> <high/low>\n"
+  " * serial disable <interface>\n",
   // Command function
   debugSerialCommand,
   // Number of parameters (variable)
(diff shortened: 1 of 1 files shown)


src/lib/debug/debug_uart.c
```diff
@@ -285,6 +286,92 @@ static BaseType_t debugSerialCommand(char *writeBuffer,
...
+                      &parameterStringLength);
+
+      LL_USART_Disable((USART_TypeDef *)handle->device);
+
+      if (strncmp("high", parameter, parameterStringLength) == 0) {
+        LL_USART_SetRXPinLevel((USART_TypeDef *)handle->device, LL_USART_RXPIN_LEVEL_STANDARD);
+      } else if (strncmp("low", parameter, parameterStringLength) == 0) {
+        LL_USART_SetRXPinLevel((USART_TypeDef *)handle->device, LL_USART_RXPIN_LEVEL_INVERTED);
+      } else {
+        printf("ERR Invalid paramters\n");
```
russelldeguzman, line 315: If we break here, will the USART be disabled indefinitely until someone tries again?
Maybe we should just re-enable with the current config if we have a invalid CLI params

src/lib/debug/debug_uart.c
```diff
@@ -285,6 +286,92 @@ static BaseType_t debugSerialCommand(char *writeBuffer,
...
+        printf("ERR Invalid interface\n");
+        break;
+      }
+
+      parameter = FreeRTOS_CLIGetParameter(
+                      commandString,
+                      3, // Get the rest of the parameters
+                      &parameterStringLength);
+
+      LL_USART_Disable((USART_TypeDef *)handle->device);
```
victorsowa12, line 307: Interesting thing to note here, the reference manual says the usart/lpuart needs to be disabled (aka the bit this func sets = 0) in order to change the CR2 register. This register tells the peripheral the poloarity of the tx/rx pins. That is why I am doing this here. The interesting part is that the reference manual also says the usart/lpuart needs to be disabled when changing the baud register BRR, however, from experience... it dosen't. 🤷‍♂️

src/lib/debug/debug_uart.c
victorsowa12: oops yeah  - i'll fix that

src/lib/debug/debug_uart.c
towynlin: 🤔 🤷‍♂️

src/lib/debug/debug_uart.c
```diff
@@ -409,10 +497,11 @@ static void processLineBufferedRxByte(void *serialHandle, uint8_t byte) {
...
 
-    // Heavy handed way of dealing with overflow for now
-    // Later we can just purge the buffer
-    // TODO - log error and clear buffer instead
-    configASSERT((lineBuffer->idx + 1) < lineBuffer->len);
+    if (lineBuffer->idx >= lineBuffer->len) {
+      printf("[%s] Line buffer overflowed, clearing!\n", handle->name);
+      memset(lineBuffer->buffer, 0, lineBuffer->len);
+      lineBuffer->idx = 0;
+    }
```
towynlin, line 504: 👏

towynlin (approved):

russelldeguzman (approved):

PR #58, opened 2023-11-30
Bug/bridge mem leak

victorsowa12:
Fixes heap corruption bug with the bridges reportBuilder. When backfilling for an element that was not part of the report builder yet the element->sample_counter was erroneously incremented in a memcpy statement, thus putting the last element beyond the allocated buffer into unallocated heap. So I removed the erroneous increment. In addition, I increased the report builders max cbor buffer size to allow for up-to a max samples in bit-packed message(26 samples).


src/apps/bridge/memfault_platform_config.h
```diff
@@ -19,3 +19,8 @@
 #define MEMFAULT_METRICS_HEARTBEAT_INTERVAL_SECS (6 * 60 * 60)
 
 #define MEMFAULT_WATCHDOG_SW_TIMEOUT_SECS 30
+
+// If you want to enable heap stats collection while debugging, uncomment the following lines
+// #define MEMFAULT_COREDUMP_COLLECT_HEAP_STATS 1
+// #define MEMFAULT_FREERTOS_PORT_HEAP_STATS_ENABLE 1
+// #define MEMFAULT_COREDUMP_HEAP_STATS_LOCK_ENABLE 0
```
victorsowa12, line 26: left commented out since we can't get bridge coredumps in the field yet

victorsowa12: but helpful when you have a debugger attached

src/apps/bridge/reportBuilder.cpp
```diff
@@ -18,10 +18,23 @@
...
+  Each sample bitpacked is ~12 bytes so we can fit 26 samples in a message
+
+  26 * (sizeof(aanderaa_aggregations_t)) = 26 * (7 * sizeof(double) + sizeof(uint32_t)) =
+  26 * (7 * 8 + 4) = 26 * 60 = 1560 bytes
+
+  Then we need to add some overhead for the cbor array, lets just use 128 bytes.
+
+  1560 + 128 = 1688 bytes
+*/
+#define MAX_SENSOR_REPORT_CBOR_LEN 1688
```
victorsowa12, line 37: I don't think we will ever get this big, but just in case!

src/apps/bridge/reportBuilder.cpp
```diff
@@ -156,7 +169,7 @@ void ReportBuilderLinkedList::addSampleToElement(report_builder_element_t *eleme
           // We use the element->sample_counter to track within each element how many samples
           // the element has received.
           for (; element->sample_counter < sample_counter; element->sample_counter++) {
-            memcpy(&(static_cast<aanderaa_aggregations_t *>(element->sensor_data))[element->sample_counter++], &NAN_AGG, sizeof(aanderaa_aggregations_t));
+            memcpy(&(static_cast<aanderaa_aggregations_t *>(element->sensor_data))[element->sample_counter], &NAN_AGG, sizeof(aanderaa_aggregations_t));
```
victorsowa12, line 175: this is the bug fix! 

the behavior with the erroneous increment is that we would have an array of 2 samples and accidentally append the second sample (when backfilling) beyond the end of the array like `[NAN sample, NULL(skipped)] NAN (beyond end of the allocated array)`

russelldeguzman (approved):

towynlin (approved):
Bravo!

PR #69, opened 2023-12-19
Add sensor sampler configs to mote_bringup

victorsowa12:
This app is used by the golden mote for bed of nails testing and for that purpose it is better if we are able to turn off the sensor sampling. So I copied over the same configs from the bmdk common apps.

https://app.shortcut.com/sofar/story/197760/add-sensor-sampler-configs-to-mote-bristlefin


towynlin (commented):
For this app, would it be better to make "sensor sampling off" the default instead of requiring that configs be set to have our intended default behavior?

victorsowa12:
Yup that is a good idea! changed to default off

towynlin (approved):

PR #72, opened 2023-12-21
ci: optionally sign each config

towynlin:


tools/scripts/release/auto_build.py
```diff
@@ -113,15 +111,18 @@ def setup_cmake_dir(self, config):
...
+        if not config["name"].startswith("bootloader") and not config["encrypt"]:
+            bootloader_cfg_key = (
+                "bootloader_signing_required"
+                if config["sign"]
+                else "bootloader_development"
+            )
+            if bootloader_cfg_key in self._configs:
+                bootloader_cfg = self._configs[bootloader_cfg_key]
+                if "elf_fullpath" in bootloader_cfg:
+                    cmd += [f'-DBOOTLOADER_PATH={bootloader_cfg["elf_fullpath"]}']
```
towynlin, line 125: This was the main slightly tricky thing — include the correct bootloader in the unified image based on whether we're signing the app. Also took care of the TODO comment.

tools/scripts/release/auto_build.py
```diff
@@ -343,8 +337,6 @@ def get_name_hash(name, numchars=None):
 parser.add_argument(
     "--prod", action="store_true", help="This is a production (non-eng) build"
 )
-parser.add_argument("--sign", action="store_true", help="Use signed images")
-parser.add_argument("--encrypt", action="store_true", help="Use encrypted images")
```
towynlin, line 347: Mostly this PR is just removing these arguments from the top level script, instead using the `sign` and `encrypt` keys from each yml config.

tools/scripts/release/auto_build.py
```diff
@@ -299,20 +300,13 @@ def build_all(self):
...
-
-                # Don't use signature support for non-prod bootloaders so they'll
-                # accept any image
-                if self.sign and (name == "bootloader") and not self.is_prod:
-                    print("Disabling signature support for bootloader")
-                    self.sign = False
+                if "sign" not in config:
+                    config["sign"] = False
+                if "encrypt" not in config:
+                    config["encrypt"] = False
```
towynlin, line 306: The `sign` and `encrypt` keys are optional in the yml config. In our current version of default.yml, `sign` is always explicitly set to true or false, and we're not using `encrypt` at all.

victorsowa12 (approved):

PR #73, opened 2024-01-04
fix: bridge power utc alignment when rtc jumps

towynlin:
Also add more logging about next state target times

### DIFF
--- a/src/apps/bridge/bridgePowerController.cpp
+++ b/src/apps/bridge/bridgePowerController.cpp
@@ -81,8 +81,12 @@ BridgePowerController::BridgePowerController(IOPinHandle_t &BusPowerPin,
 void BridgePowerController::powerControlEnable(bool enable) {
   _powerControlEnabled = enable;
   if (enable) {
-    _sampleIntervalStartS = getEpochS();
-    _subsampleIntervalStartS = _sampleIntervalStartS;
+    uint32_t currentCycleS = getEpochS();
+    if (_sampleIntervalStartS < currentCycleS) {
+      _sampleIntervalStartS =
+          _alignNextInterval(currentCycleS, _sampleIntervalStartS, _sampleIntervalS);
+      _subsampleIntervalStartS = _sampleIntervalStartS;
+    }
     printf("Bridge power controller enabled\n");
   } else {
     printf("Bridge power controller disabled\n");
@@ -166,18 +170,16 @@ static void stateLogPrintTarget(const char *state, uint32_t target) {
   vPortFree(printbuf);
 }
 
-void BridgePowerController::_update(
-    void) { // FIXME: Refactor this function to libStateMachine: https://github.com/wavespotter/bristlemouth/issues/379
+void BridgePowerController::_update(void) {
   uint32_t time_to_sleep_ms = MIN_TASK_SLEEP_MS;
   do {
     if (!_initDone) { // Initializing
       BRIDGE_LOG_PRINT("Bridge State Init\n");
-      powerBusAndSetSignal(
-          true,
-          false); // We start Bus on, no need to signal an eth up / power up event to l2 & adin
-      vTaskDelay(
-          INIT_POWER_ON_TIMEOUT_MS); // Set bus on for two minutes for init.
-      if(_configError) {
+      // We start Bus on, no need to signal an eth up / power up event to l2 & adin
+      powerBusAndSetSignal(true, false);
+      // Set bus on for two minutes for init.
+      vTaskDelay(INIT_POWER_ON_TIMEOUT_MS);
+      if (_configError) {
         BRIDGE_LOG_PRINT("Bridge configuration error! Please check configs, using default.\n");
       }
       static constexpr size_t printBufSize = 200;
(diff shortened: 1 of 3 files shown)


src/apps/bridge/bridgePowerController.cpp
```diff
@@ -298,12 +343,48 @@ uint32_t BridgePowerController::getEpochS() {
...
+ */
+uint32_t BridgePowerController::_alignNextInterval(uint32_t nowEpochS,
+                                                   uint32_t lastIntervalStartS,
+                                                   uint32_t sampleIntervalS) {
+  uint32_t alignedEpoch = lastIntervalStartS + sampleIntervalS;
+  while (alignedEpoch < nowEpochS) {
+    // If the aligned epoch is in the past, the RTC must have just jumped forward.
+    // We need to add sample intervals until we reach the future.
+    alignedEpoch += sampleIntervalS;
+  }
```
towynlin, line 368: The crux of the issue in sc-197129 was that when the RTC would jump significantly forward, the "next interval start time" would still be in the past, because we were only adding exactly one interval to the last start time. This would lead to the bus being powered off and never being powered back on again. The solution is to keep adding intervals until we actually have a "next interval start time" that is in the future. @evanShap FYI

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -298,12 +343,48 @@ uint32_t BridgePowerController::getEpochS() {
...
+
+  // If an alignment is configured, we need to align sampling intervals to UTC.
+  if (_alignmentS != 0) {
+    uint32_t remainder = alignedEpoch % _alignmentS;
+    if (remainder != 0) {
+      uint32_t adjustment = _alignmentS - remainder;
+      // We only align forward because subtracting could take us into the past.
+      // It would be possible to handle that situation,
+      // but the code would get much more complicated.
+      alignedEpoch += adjustment;
```
towynlin, line 378: I actually started down the road of making the smallest adjustment, dropping back if we're only a little ahead or jumping forward if we're a little behind. At some point however, I realized that there was an edge case where we go back into the past, and handling it made this function super long, unwieldy, and hard to understand.

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -298,12 +343,48 @@ uint32_t BridgePowerController::getEpochS() {
...
+
+      constexpr size_t bufsize = 128;
+      char buffer[bufsize];
+      int len =
+          snprintf(buffer, bufsize,
+                   "Aligning next sample interval to UTC by delaying an additional %" PRIu32
+                   " seconds to %" PRIu32 "\n",
```
towynlin, line 385: New log line about UTC alignment as requested @evanShap

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -245,16 +272,33 @@ void BridgePowerController::_update(
...
     }
 
   } while (0);
+
+  if (time_to_sleep_ms > MIN_TASK_SLEEP_MS) {
+    constexpr size_t printBufSize = 48;
+    char *printbuf = static_cast<char *>(pvPortMalloc(printBufSize));
+    configASSERT(printbuf);
+    int len = snprintf(printbuf, printBufSize, "Controller task will wait %" PRIu32 " ms\n",
+                       time_to_sleep_ms);
```
towynlin, line 296: New log line about exactly how long this task will wait before controlling the bus again

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -152,18 +156,30 @@ void BridgePowerController::subsampleEnable(bool enable) {
...
 bool BridgePowerController::isSubsampleEnabled() { return _subsamplingEnabled; }
 
-void BridgePowerController::_update(
-    void) { // FIXME: Refactor this function to libStateMachine: https://github.com/wavespotter/bristlemouth/issues/379
+static void stateLogPrintTarget(const char *state, uint32_t target) {
+  constexpr int printBufSize = 96;
+  char *printbuf = static_cast<char *>(pvPortMalloc(printBufSize));
+  configASSERT(printbuf);
+  int len = snprintf(printbuf, printBufSize,
+                     "Bridge State %s until %" PRIu32 " epoch seconds\n", state, target);
```
towynlin, line 164: Every state change log in {sample, sample off, subsample, subsample off} now also prints the target time for the subsequent state @evanShap

test/src/common/bridgePowerController_ut.cpp
```diff
@@ -28,6 +28,18 @@ BaseType_t (*custom_fakes[])(RTCTimeAndDate_t *timeAndDate) = {
...
+  (void)_;
+  *pinval = persistent_fake_pin_val;
+  return true;
+}
+bool io_write_custom_fake(const void *_, uint8_t pinval) {
+  (void)_;
+  persistent_fake_pin_val = pinval;
+  return true;
+}
+
```
towynlin, line 42: Without these custom fakes to track the persistent pin state, the "golden path" test below was actually just always succeeding! It was completely wrong and not helping us detect regressions. That test definitely should've started failing at least when we removed the bus power cycle after the init period, but it didn't. I don't know why it started failing in CI while I worked on this PR, but I'm glad it did! @russelldeguzman @victorsowa12 I'd like to verbally walk through these tests in a review to make sure we all understand them.

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -245,16 +272,33 @@ void BridgePowerController::_update(
...
           powerBusAndSetSignal(false);
         }
+
+        // Align the first sample to UTC when the RTC finally gets set
+        checkAndUpdateRTC();
+        uint32_t currentCycleS = getEpochS();
+        if (_rtcSet && _sampleIntervalStartS > currentCycleS) {
+          time_to_sleep_ms =
+              MAX((_sampleIntervalStartS - currentCycleS) * 1000, MIN_TASK_SLEEP_MS);
+        }
```
towynlin, line 284: Prior to this PR, we would not bother aligning the first sample, so for scientific research purposes, customers would have one data point that arguably they should throw out, though they probably didn't know it. This code aligns the first sample when the bridge power controller was already configured enabled but the RTC wasn't set, and then the RTC gets set the first time. There are 2 other similar places — (1) init and (2) when the RTC was already set but power control was disabled and then gets enabled.

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -185,46 +201,59 @@ void BridgePowerController::_update(
...
       BRIDGE_LOG_PRINT("Bridge State Init Complete\n");
       checkAndUpdateRTC();
+      uint32_t currentCycleS = getEpochS();
+      if (_rtcSet && _sampleIntervalStartS > currentCycleS) {
+        time_to_sleep_ms =
+            MAX((_sampleIntervalStartS - currentCycleS) * 1000, MIN_TASK_SLEEP_MS);
+        // The default state until first sample interval depends
+        // on whether bus power control is enabled.
+        powerBusAndSetSignal(!_powerControlEnabled);
+      }
```
towynlin, line 211: This handles aligning the first sample to UTC in the case where the RTC has already been set by the time the init period ends.

russelldeguzman (approved):
👍

victorsowa12 (approved):

PR #75, opened 2024-01-12
Fix/mcuboot reverts

victorsowa12:
Fixes a flash write alignment bug in the bootloader that makes MCUBoot vulnerable to power downs mid image swap.

### DIFF
--- a/src/apps/bridge/app_main.cpp
+++ b/src/apps/bridge/app_main.cpp
@@ -336,7 +336,8 @@ static void defaultTask( void *parameters ) {
 
     debugSysInit();
     debugMemfaultInit(&usbCLI);
-    mcubootCliInit();
+    // uncomment for the `update sec`, `update clr`, `update confirm` commands
+    // mcubootCliInit();
 
     debugGpioInit(debugGpioPins, sizeof(debugGpioPins)/sizeof(DebugGpio_t));
     debugSpotterInit();
--- a/src/apps/bridge/app_main.cpp
+++ b/src/apps/bridge/app_main.cpp
@@ -30,7 +30,6 @@
 #include "debug_rtc.h"
 #include "debug_sys.h"
 #include "debug_w25.h"
-#include "mcuboot_cli.h"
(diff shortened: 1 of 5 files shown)


src/apps/bridge/app_main.cpp
```diff
@@ -335,6 +336,7 @@ static void defaultTask( void *parameters ) {
 
     debugSysInit();
     debugMemfaultInit(&usbCLI);
+    mcubootCliInit();
```
victorsowa12: Needed to access the `update sec` and `update clr` commands for testing testing image swapping on an isolated board. Will be removed before merge (unless we want to keep it)

src/lib/mcuboot/port_flash.c
```diff
@@ -172,7 +172,7 @@ int     flash_area_erase(const struct flash_area *area, uint32_t off, uint32_t l
 /*< Returns this `flash_area`s alignment */
 size_t flash_area_align(const struct flash_area *area) {
   (void)area;
-  return 8;
+  return MCUBOOT_BOOT_MAX_ALIGN;
```
victorsowa12, line 175: The fix! It needs to be 16 since that is the min-write-size of the STM32U5's flash. This is defined as 16 in src/CMakeLists.txt

src/lib/mcuboot/include/mcuboot_config/mcuboot_config.h
```diff
@@ -95,7 +95,7 @@
 
 /* Default maximum number of flash sectors per image slot; change
  * as desirable. */
-#define MCUBOOT_MAX_IMG_SECTORS 256
+#define MCUBOOT_MAX_IMG_SECTORS 121
```
victorsowa12, line 98: This is the number of sectors, which should be 968kB(app_size)/8kB(flash page size)

russelldeguzman (approved):
🎉

towynlin (approved):
👏👏👏

PR #77, opened 2024-01-25
adding support for sample type in the report encoder

russelldeguzman:
To support additional sensors (i.e. BM soft) when we need to know what type of sample to decode/bitpack on the Spotter. 
Therefore, I added a string at the top of each sample array to define what kind of samples are contained in the array. 

On the Spotter side - I'll generalize a report builder decode library and account for this extra string member in the sample array.

### DIFF
--- a/src/apps/bridge/cbor_sensor_report_encoder.cpp
+++ b/src/apps/bridge/cbor_sensor_report_encoder.cpp
@@ -87,14 +87,25 @@ CborError sensor_report_encoder_close_sensor(sensor_report_encoder_context_t &co
 }
 
 /*!
- * @brief Open a sample array for encoding.
+ * @brief Open a sample array for encoding. The first member in the sampe array is the sample type.
  * @param[in] context The encoding context.
  * @param[in] num_sample_members The number of samples in the sensor.
+ * @param[in] sample_type A string defining type of sample.
  * @return Cbor error code.
  */
 CborError sensor_report_encoder_open_sample(sensor_report_encoder_context_t &context,
-                                            size_t num_sample_members) {
-  return cbor_encoder_create_array(&context.sensor_array, &context.sample_array, num_sample_members);
+                                            size_t num_sample_members,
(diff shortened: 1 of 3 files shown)


src/apps/bridge/reportBuilder.cpp
```diff
@@ -323,70 +349,90 @@ void reportBuilderAddToQueue(uint64_t node_id, uint8_t sensor_type, void *sensor
...
-        BRIDGE_LOG_PRINT("Failed to close sample in addSamplesToReport\n");
-        break;
-      }
-      rval = true;
+  switch (sensor_type) {
+  case AANDERAA_SENSOR_TYPE: {
+    aanderaa_aggregations_t aanderaa_sample =
+        (static_cast<aanderaa_aggregations_t *>(sensor_data))[sample_index];
+    if (sensor_report_encoder_open_sample(context, AANDERAA_NUM_SAMPLE_MEMBERS,
+                                          "aandera_current_v0") != CborNoError) {
```
russelldeguzman: This is the only "real" change in the file. 
The rest is `clang-format`

src/apps/bridge/cbor_sensor_report_encoder.cpp
```diff
@@ -87,14 +87,25 @@ CborError sensor_report_encoder_close_sensor(sensor_report_encoder_context_t &co
 }
 
 /*!
- * @brief Open a sample array for encoding.
+ * @brief Open a sample array for encoding. The first member in the sampe array is the sample type.
```
victorsowa12: nit: sampe -> sample

victorsowa12 (approved):
Small nit but looks good!

src/apps/bridge/reportBuilder.cpp
towynlin: nit: aandera -> aanderaa

towynlin (approved):
similarly, looks good — found one more typo

PR #79, opened 2024-01-26
docs: add important doxygen comments to spotter_tx_data

towynlin:
Multiple folks in the community forum have hit these limitations. See for example https://bristlemouth.discourse.group/t/more-info-on-growing-set-of-abstract-parsing-tools/247/8?u=zachary


victorsowa12 (approved):
👏

PR #80, opened 2024-02-01
Sc 198151/bridge soft support

victorsowa12:
Adding support for SOFT on the bridge. The bridge will by default assume a 2 HZ sampling frequency(there is a config to change this expected frequency if needed) for the Soft modules and upon subscribing to the Soft node it will create buffers large enough to take in that many samples. It will then aggregate those buffers at the end of a sample duration and send the aggregated sample to the reportBuilder. After receiving enough aggregations the reportBuilder will cbor encode the SOFT aggregations and send them to Spotter.

Right not I implemented logging of individual and aggregate readings in the same way the Aanderaa logs are done, however, there has been a decision to merge these into shared logs. There are still some details that need to be spec'd out, but those changes will come in a separate PR.

### DIFF
--- a/src/apps/bridge/app_pub_sub.h
+++ b/src/apps/bridge/app_pub_sub.h
@@ -38,6 +38,14 @@ extern "C" {
 #define APP_PUB_SUB_BM_BRIDGE_AANDERAA_AGG_TYPE        1
 #define APP_PUB_SUB_BM_BRIDGE_AANDERAA_AGG_VERSION     1
 
+#define APP_PUB_SUB_BM_BRIDGE_SOFT_IND_TOPIC       "bridge/soft_ind"
+#define APP_PUB_SUB_BM_BRIDGE_SOFT_IND_TYPE        1
+#define APP_PUB_SUB_BM_BRIDGE_SOFT_IND_VERSION     1
+
+#define APP_PUB_SUB_BM_BRIDGE_SOFT_AGG_TOPIC       "bridge/soft_agg"
+#define APP_PUB_SUB_BM_BRIDGE_SOFT_AGG_TYPE        1
+#define APP_PUB_SUB_BM_BRIDGE_SOFT_AGG_VERSION     1
+
 typedef struct app_pub_sub_bm_bridge_sensor_report_data {
     uint32_t bm_config_crc32;
     size_t cbor_buffer_len;
--- a/src/apps/bridge/app_pub_sub.h
+++ b/src/apps/bridge/app_pub_sub.h
@@ -30,6 +30,7 @@ extern "C" {
 #define APP_PUB_SUB_BM_BRIDGE_PRINTF_TYPE        1
 #define APP_PUB_SUB_BM_BRIDGE_PRINTF_VERSION     1
 
+// TODO - keep as sensor specific logs or switch to the bm_sensor common log
 #define APP_PUB_SUB_BM_BRIDGE_AANDERAA_IND_TOPIC       "bridge/aanderaa_ind"
 #define APP_PUB_SUB_BM_BRIDGE_AANDERAA_IND_TYPE        1
 #define APP_PUB_SUB_BM_BRIDGE_AANDERAA_IND_VERSION     1
@@ -45,6 +46,7 @@ extern "C" {
 #define APP_PUB_SUB_BM_BRIDGE_SOFT_AGG_TOPIC       "bridge/soft_agg"
 #define APP_PUB_SUB_BM_BRIDGE_SOFT_AGG_TYPE        1
 #define APP_PUB_SUB_BM_BRIDGE_SOFT_AGG_VERSION     1
+// END TODO
 
 typedef struct app_pub_sub_bm_bridge_sensor_report_data {
     uint32_t bm_config_crc32;
--- a/src/apps/bridge/bridgeLog.cpp
+++ b/src/apps/bridge/bridgeLog.cpp
@@ -21,6 +21,14 @@ void bridgeSensorLogPrintf(bridgeSensorLogType_e type, const char *str, size_t l
                 printf("[%s] %.*s", "AANDERAA_AGG", len, str);
                 bm_serial_pub(getNodeId(), APP_PUB_SUB_BM_BRIDGE_AANDERAA_AGG_TOPIC, sizeof(APP_PUB_SUB_BM_BRIDGE_AANDERAA_AGG_TOPIC)-1, reinterpret_cast<const uint8_t*>(str),len, APP_PUB_SUB_BM_BRIDGE_AANDERAA_AGG_TYPE, APP_PUB_SUB_BM_BRIDGE_AANDERAA_AGG_VERSION);
                 break;
+            case SOFT_IND:
+                printf("[%s] %.*s", "SOFT_IND", len, str);
+                bm_serial_pub(getNodeId(), APP_PUB_SUB_BM_BRIDGE_SOFT_IND_TOPIC, sizeof(APP_PUB_SUB_BM_BRIDGE_SOFT_IND_TOPIC)-1, reinterpret_cast<const uint8_t*>(str),len, APP_PUB_SUB_BM_BRIDGE_SOFT_IND_TYPE, APP_PUB_SUB_BM_BRIDGE_SOFT_IND_VERSION);
+                break;
+            case SOFT_AGG:
+                printf("[%s] %.*s", "SOFT_AGG", len, str);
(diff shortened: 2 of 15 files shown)


src/apps/bridge/app_pub_sub.h
```diff
@@ -30,6 +30,7 @@ extern "C" {
 #define APP_PUB_SUB_BM_BRIDGE_PRINTF_TYPE        1
 #define APP_PUB_SUB_BM_BRIDGE_PRINTF_VERSION     1
 
+// TODO - keep as sensor specific logs or switch to the bm_sensor common log
```
victorsowa12, line 33: will change to shared logs in another PR

src/apps/bridge/bridgeLog.cpp
```diff
@@ -13,6 +13,7 @@ void bridgeLogPrintf(const char *str, size_t len) {
 void bridgeSensorLogPrintf(bridgeSensorLogType_e type, const char *str, size_t len) {
     if(len > 0){
         switch(type) {
+            // TODO - keep as sensor specific logs or switch to the bm_sensor common log
```
victorsowa12, line 16: will change in another PR to shared logs

src/apps/bridge/bridgeLog.h
```diff
@@ -5,7 +5,9 @@ constexpr size_t SENSOR_LOG_BUF_SIZE = 512;
 
 typedef enum {
     AANDERAA_IND,
-    AANDERAA_AGG 
+    AANDERAA_AGG,
+    SOFT_IND,
+    SOFT_AGG
```
victorsowa12, line 10: will change to shared logs in another PR

src/apps/bridge/sensor_drivers/softSensor.h
```diff
@@ -0,0 +1,41 @@
...
+#include "sensorController.h"
+
+#include <stdint.h>
+#include <stdlib.h>
+
+#define SOFT_NUM_SAMPLE_MEMBERS 1
+
+typedef struct soft_aggregations_s {
+  double temp_mean_deg_c;
+  uint32_t reading_count;
```
victorsowa12, line 14: I autopilot-added reading_count to the SOFT sensors, but based on discussions this has been punted till later. 

I can leave this in or remove it. Either way it is not being sent out, but it is being logged in the AGG log (although the  IND/AGG logs will change in another PR too, so I can remove it in the PR if that makes sense to)

src/apps/bridge/sensor_drivers/softSensor.cpp
```diff
@@ -0,0 +1,138 @@
...
+    if (!logRtcGetTimeStr(time_str, TIME_STR_BUFSIZE, true)) {
+      printf("Failed to get RTC time string for Soft aggregation\n");
+      snprintf(time_str, TIME_STR_BUFSIZE, "0");
+    }
+    log_buflen = snprintf(log_buf, SENSOR_LOG_BUF_SIZE,
+                          "%s,"          // timstamp(ticks/UTC)
+                          "%" PRIx64 "," // Node Id
+                          "%" PRIu32 "," // reading_count
+                          "%.3f\n",      // temp_mean_deg_c
+                          time_str, node_id, soft_aggs.reading_count, soft_aggs.temp_mean_deg_c);
```
victorsowa12: log the reading count for now. Will decide on removing reading count when the shared log stuff is done.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -55,6 +57,8 @@ static aanderaa_aggregations_t NAN_AGG = {.abs_speed_mean_cm_s = NAN,
                                           .std_tilt_mean_rad = NAN,
                                           .reading_count = 0};
 
+static soft_aggregations_t SOFT_NAN_AGG = {.temp_mean_deg_c = NAN, .reading_count = 0};
+
```
victorsowa12, line 61: the reading count is not sent out, but is part of the struct so it can be logged in the softSensor.cpp. (like many other related comments mention, I think this can be decided on when doing the shared logs PR)

src/apps/bridge/reportBuilder.h
```diff
@@ -4,17 +4,14 @@
...
 
 typedef enum {
   REPORT_BUILDER_INCREMENT_SAMPLE_COUNT,
   REPORT_BUILDER_SAMPLE_MESSAGE,
   REPORT_BUILDER_CHECK_CRC,
 } report_builder_message_e;
 
-typedef enum {
-  AANDERAA_SENSOR_TYPE = 0,
-} report_builder_sensor_type_e;
```
victorsowa12, line 16: no longer needed and we will use the enum defined in abstractSensor.h

src/apps/bridge/reportBuilder.cpp
```diff
@@ -88,6 +92,7 @@ typedef struct ReportBuilderContext_s {
   uint32_t _transmitAggregations;
   ReportBuilderLinkedList _reportBuilderLinkedList;
   uint64_t _report_period_node_list[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
+  abstractSensorType_e _report_period_sensor_type_list[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
```
victorsowa12, line 95: use a list to map a node_id to an abstract sensor type enum value. This is to allow us to add a NAN report for a sensor that may be part of our topology + crc, but for some reason was not subscribed to in the sensorController.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -484,19 +540,40 @@ static void report_builder_task(void *parameters) {
...
+                            _ctx._report_period_sensor_type_list[i], NULL,
+                            sizeof(soft_aggregations_t), _ctx._samplesPerReport,
+                            (_ctx._sample_counter - 1));
+                        break;
+                      }
+                      default: {
+                        BRIDGE_LOG_PRINT("Invalid sensor type in report_builder_task\n");
+                        break;
+                      }
+                      }
```
victorsowa12: this is where we would add a sensor that is in our topology, but failed to be subscribed to by sensorController. The sensor type is required so we know what size of data we will be adding to the report.

src/lib/common/topology_sampler.cpp
```diff
@@ -53,6 +52,7 @@ typedef struct network_configuration_info {
 
 typedef struct node_list {
   uint64_t nodes[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
+  abstractSensorType_e sensor_type[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
```
victorsowa12, line 55: I created another list that will match the nodes (node_id) list such that element 0 in both lists will reference the same node.

src/lib/common/topology_sampler.cpp
```diff
@@ -716,3 +756,36 @@ void bm_topology_last_network_info_cb(void){
...
+ * aka zero's.
+ *
+ * Note: This function can only be called in a location that has taken the
+ * _node_list.node_list_mutex.
+ *
+ * @param node_id The ID of the node to update.
+ * @param app_name The name of the application associated with the node.
+ * @param app_name_len The length of the application name.
+ */
+void _update_sensor_type_list(uint64_t node_id, char *app_name, uint32_t app_name_len) {
```
russelldeguzman:
```suggestion
static void _update_sensor_type_list(uint64_t node_id, char *app_name, uint32_t app_name_len) {
```

src/apps/bridge/sensorController.cpp
```diff
@@ -4,12 +4,14 @@
...
 #include "device_info.h"
 #include "reportBuilder.h"
+#include "softSensor.h"
 #include "sys_info_service.h"
 #include "sys_info_svc_reply_msg.h"
 #include "task_priorities.h"
 #include "util.h"
 
 #define DEFAULT_CURRENT_READING_PERIOD_MS 60 * 1000 // default is 1 minute: 60,000 ms
+#define DEFAULT_SOFT_READING_PERIOD_MS 500 // default is 500 ms (2 HZ)
```
russelldeguzman, line 15: TODO: Once we have `bcmp_config` request reply, we should read this value from the modules.

src/apps/bridge/sensor_drivers/aanderaaSensor.h
```diff
@@ -8,6 +8,11 @@
 
 #define AANDERAA_NUM_SAMPLE_MEMBERS 8
 
+// Clang doesn't have M_TWOPI defined in math.h so for CI tests we define it here.
```
towynlin, line 11: 🤦‍♂️thx

towynlin (approved):

PR #83, opened 2024-02-09
Sc 198623/rbr app

russelldeguzman:
Implements the RBR driver on the mote for CodaT,  CodaD, and Coda DT
* Adds FTL + reporting via `SensorWatchdog`
* Adds Configuration behavior via RBR type
* Adds parsing behavior dep on RBR type
* Adds some string parsing protections 
https://www.notion.so/sofarocean/SM-product-spec-RBR-Coda-Adapter-Module-bcfcdfa5f3fa4ca2811415089d2f8a29


src/apps/bristleback_apps/bm_rbr/user_code/rbr_sensor.cpp
```diff
@@ -0,0 +1,234 @@
...
+  // Turn on the UART.
+  PLUART::enable();
+}
+
+/*!
+* @brief Probe the sensor type.
+* @param timeout_ms The timeout in milliseconds.
+* @return True if the sensor type was successfully probed, false otherwise.
+*/
+bool RbrSensor::probeType(uint32_t timeout_ms) {
```
russelldeguzman, line 50: Polling of the RBR sensor to get its type. If there is no config in flash, or a new RBR sensor is swapped in - save it to config. 
If we can't read the sensor type 3 times in a row - set the type to unknown.

src/lib/common/payload_uart.cpp
```diff
@@ -127,13 +131,13 @@ uint16_t readLine(char *buffer, size_t len) {
   int16_t rval = 0;
   size_t copy_len;
 
+  configASSERT(xSemaphoreTake(_user_line.mutex, portMAX_DELAY) == pdTRUE);
```
russelldeguzman, line 134: Moved this further up so `_user_line.len` stays in sync

victorsowa12 (approved):
🤘

towynlin (approved):
👏💙

PR #85, opened 2024-02-21
Sc 198619/get rbr type from cbor map

victorsowa12:
Here I have updated reportBuilder such that it keeps a local copy of the "max" CBOR map it receives during a sample interval. Then someone can get a copy of that "max" CBOR map and decode it to discover any relevant information they want. In this case upon aggregation, the rbrCodaSensor driver will copy the "max" CBOR map and parse through it looking for the desired node's `rbrCodaType` system config. The rbrCodaSensor driver will then stamp its aggregated sample with that type such that the Report Builder pass that RBR type on to Spotter who will then be able to bitpack the data appropriately. 

The reason for decoding the CBOR map is so that the `rbrCodaType` that the rbrCodaSensor driver, Report Builder, and Topology/CRC reference is the same. This will guarantee that when a BMSensorData1 message is received on the backend there will be no mismatch between the BMSensorData1 messages CRC and its bitpacked data.

### DIFF
--- a/src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
@@ -215,6 +215,8 @@ BmRbrDataMsg::SensorType_t RbrCodaSensor::rbrCodaGetSensorType(uint64_t node_id)
   if (network_config != NULL) {
     CborParser parser;
     CborValue all_nodes_array, individual_node_array, node_array_member;
+    char *app_name_str = NULL;
+    char *key_str = NULL;
     do {
       if (cbor_parser_init(network_config, cbor_config_size, 0, &parser, &all_nodes_array) !=
           CborNoError) {
@@ -305,15 +307,15 @@ BmRbrDataMsg::SensorType_t RbrCodaSensor::rbrCodaGetSensorType(uint64_t node_id)
             printf("Failed to get the length of the node app name\n");
             break;
           }
-          char *strbuf = static_cast<char *>(pvPortMalloc(len + 1));
-          configASSERT(strbuf);
-          if (cbor_value_copy_text_string(&node_array_member, strbuf, &len, NULL) !=
+          app_name_str = static_cast<char *>(pvPortMalloc(len + 1));
+          configASSERT(app_name_str);
+          if (cbor_value_copy_text_string(&node_array_member, app_name_str, &len, NULL) !=
               CborNoError) {
             break;
           }
-          strbuf[len] = '\0';
-          printf("RBR Node app name: %s\n", strbuf);
-          vPortFree(strbuf);
+          app_name_str[len] = '\0';
+          printf("RBR Node app name: %s\n", app_name_str);
+          vPortFree(app_name_str);
           if (cbor_value_advance(&node_array_member) != CborNoError) {
             printf("Failed to advance the node array member\n");
             break;
@@ -366,7 +368,7 @@ BmRbrDataMsg::SensorType_t RbrCodaSensor::rbrCodaGetSensorType(uint64_t node_id)
               printf("Failed to get the length of the sys_cfg_map key\n");
               break;
             }
-            char *key_str = static_cast<char *>(pvPortMalloc(len + 1));
+            key_str = static_cast<char *>(pvPortMalloc(len + 1));
             configASSERT(key_str);
             if (cbor_value_copy_text_string(&sys_cfg_map, key_str, &len, NULL) != CborNoError) {
               break;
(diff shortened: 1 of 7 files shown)


src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
```diff
@@ -193,4 +190,229 @@ RbrCoda_t *createRbrCodaSub(uint64_t node_id, uint32_t rbr_coda_agg_period_ms,
...
+          size_t len;
+          if (cbor_value_get_string_length(&node_array_member, &len) != CborNoError) {
+            printf("Failed to get the length of the node app name\n");
+            break;
+          }
+          char *strbuf = static_cast<char *>(pvPortMalloc(len + 1));
+          configASSERT(strbuf);
+          if (cbor_value_copy_text_string(&node_array_member, strbuf, &len, NULL) !=
+              CborNoError) {
+            break;
```
russelldeguzman, line 314: Breaking here will leak `strbuf`.
It's easier to manage the memory if you keep all the alloc pointers outside the `do {} while(0)` and free them at the end.

src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
```diff
@@ -193,4 +190,229 @@ RbrCoda_t *createRbrCodaSub(uint64_t node_id, uint32_t rbr_coda_agg_period_ms,
...
+            }
+            size_t len;
+            if (cbor_value_get_string_length(&sys_cfg_map, &len) != CborNoError) {
+              printf("Failed to get the length of the sys_cfg_map key\n");
+              break;
+            }
+            char *key_str = static_cast<char *>(pvPortMalloc(len + 1));
+            configASSERT(key_str);
+            if (cbor_value_copy_text_string(&sys_cfg_map, key_str, &len, NULL) != CborNoError) {
+              break;
```
russelldeguzman, line 373: Same comment as above

src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
victorsowa12: Good catch! This alloc and associated print are more for debugging purposes, so they could be removed entirely or do you think it is worth keeping it?

src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
victorsowa12: this one is also a bit more for debugging, but this info is a bit more relevant to be printed out. In the end we can deduce from the AGG log what the reportBuilder thought the RBR type was so it may just be better to remove this one as well?

src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
victorsowa12: I removed the potential leaks!

src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
victorsowa12: removed the potential leak!

src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
russelldeguzman: Keeping for debug is fine as long as we free it! (which you did)

src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
```diff
@@ -193,4 +190,237 @@ RbrCoda_t *createRbrCodaSub(uint64_t node_id, uint32_t rbr_coda_agg_period_ms,
...
+                printf("RBR type RBR.D\n");
+              } else if (current_sensor_type ==
+                         BmRbrDataMsg::SensorType::PRESSURE_AND_TEMPERATURE) {
+                printf("RBR type RBR.DT\n");
+              } else {
+                printf("Failed to get the sensor type, setting to UNKOWN\n");
+              }
+              break;
+            }
+            vPortFree(key_str);
```
russelldeguzman: Remove this free, otherwise we'll double free and crash.

src/apps/bridge/sensor_drivers/rbrCodaSensor.cpp
```diff
@@ -193,4 +190,237 @@ RbrCoda_t *createRbrCodaSub(uint64_t node_id, uint32_t rbr_coda_agg_period_ms,
...
+          }
+          app_name_str = static_cast<char *>(pvPortMalloc(len + 1));
+          configASSERT(app_name_str);
+          if (cbor_value_copy_text_string(&node_array_member, app_name_str, &len, NULL) !=
+              CborNoError) {
+            break;
+          }
+          app_name_str[len] = '\0';
+          printf("RBR Node app name: %s\n", app_name_str);
+          vPortFree(app_name_str);
```
russelldeguzman: Remove this free (same as below).

russelldeguzman (approved):
This looks good! Nice!

PR #89, opened 2024-02-27
Feature/bcmp linked list

victorsowa12:
This PR has a linked list for outgoing "sequenced" messages. These element on the linked list will be removed when we get a BCMP message with the same sequence number, otherwise after 500ms we will time out and remove the element from the linked list. 

This PR just has the linked list implemented and basic req/rep for configs and ping (note: PING needs to be removed before the final merge, but it is an easy one to test with in a 2 node system.)

I have created 2 helper functions to categorize the BCMP message types into sequenced requests and sequenced responses. These helper functions are used when sending a BCMP message out and when receiving a BCMP message to determine if we a) care about the sequence number, b) need to stamp the request message with the local sequence number, c) forward the received sequence number in a reply.

Example of a mote requesting sys config status from a bridge:
```
bm cfg status 4d58ce2f22dfb374 s
BCMP - Sending message with seq_num 4
Successful status request send
BCMP message with seq_num 4 found
BCMP - Received reply to message with seq_num 4
Response msg -- Node Id:4d58ce2f22dfb374,Partition:1, Commit Status:0
Num Keys: 10
bridgePowerControllerEnabled
samplesPerReport
transmitAggregations
sampleIntervalMs
sampleDurationMs
subsampleIntervalMs
subsampleEnabled
currentReadingPeriodMs
alignmentInterval5Min
subsampleDurationMs
```

Testing TODO:
3+ node system

### DIFF
--- a/src/lib/bcmp/bcmp.cpp
+++ b/src/lib/bcmp/bcmp.cpp
@@ -2,6 +2,7 @@
 #include "FreeRTOS.h"
 #include "task.h"
 #include "queue.h"
+#include "semphr.h"
 #include "timers.h"
 
 #include "bcmp.h"
@@ -21,6 +22,7 @@
 #include "bcmp_time.h"
 #include "bcmp_topology.h"
 #include "bcmp_resource_discovery.h"
+#include "timer_callback_handler.h"
 
 #include "bm_dfu.h"
 
@@ -37,12 +39,27 @@
 // and their sequence numbers and the callback to call
 // upon receipt of the reply.
 // The callback should also handle the timeout case.
+static constexpr uint32_t DEFAULT_MESSAGE_TIMEOUT_MS = 100;
+static constexpr uint32_t MESSAGE_TIMER_EXPIRY_PERIOD_MS = 500;
+
+typedef struct bcmp_element {
+  uint32_t seq_num;
+  uint16_t type;
+  uint32_t send_timestamp_ms;
+  uint32_t timeout_ms;
+  bcmp_element *next;
+} bcmp_element_t;
 
 typedef struct {
   struct netif* netif;
   struct raw_pcb *pcb;
   QueueHandle_t rx_queue;
   TimerHandle_t heartbeat_timer;
+  // Linked list of sent BCMP messages
+  bcmp_element_t *messages_list;
+  SemaphoreHandle_t messages_list_mutex;
+  TimerHandle_t messages_expiration_timer;
+  uint32_t message_count;
 } bcmpContext_t;
 
 typedef enum {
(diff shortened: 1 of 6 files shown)


src/lib/bcmp/bcmp.cpp
```diff
@@ -32,11 +34,27 @@
 // 1500 MTU minus ipv6 header
 #define MAX_PAYLOAD_LEN (1500 - sizeof(struct ip6_hdr))
 
+static constexpr uint32_t DEFAULT_MESSAGE_TIMEOUT_MS = 100;
```
russelldeguzman: I'd make this value > MESSAGE_TIMER_EXPIRY_PERIOD_MS

src/lib/bcmp/bcmp.cpp
```diff
@@ -32,11 +34,27 @@
...
+static constexpr uint32_t MESSAGE_TIMER_EXPIRY_PERIOD_MS = 500;
+
+typedef struct bcmp_element {
+  uint32_t seq_num;
+  uint16_t type;
+  uint32_t send_timestamp_ms;
+  uint32_t timeout_ms;
+  bcmp_element *next;
+} bcmp_element_t;
```
russelldeguzman: Nit: `bcmp_request_element_t` may be a bit easier for future users to grok

src/lib/bcmp/bcmp.cpp
```diff
@@ -131,6 +158,23 @@ int32_t bmcp_process_packet(struct pbuf *pbuf, ip_addr_t *src, ip_addr_t *dst) {
...
+      bcmp_element_t *sent_message = _message_list_find_message(header->seq_num);
+      if (sent_message) {
+        printf("BCMP - Received reply to message with seq_num %d\n", header->seq_num);
+        if (xSemaphoreTake(_ctx.messages_list_mutex, pdMS_TO_TICKS(DEFAULT_MESSAGE_TIMEOUT_MS)) == pdPASS) {
+          _message_list_remove_message(sent_message);
+          xSemaphoreGive(_ctx.messages_list_mutex);
+        }
+      } else {
+        // Not one of our messages, so we should ignore it.
+        printf("BCMP - Received message with seq_num %d that we didn't send\n", header->seq_num);
```
russelldeguzman: Will this drop messages that we should forward?

src/lib/bcmp/bcmp.cpp
```diff
@@ -488,3 +551,134 @@ void bcmp_init(struct netif* netif, NvmPartition * dfu_partition, Configuration*
...
+      type == BCMP_DFU_PAYLOAD_REQ ||
+      type == BCMP_DFU_PAYLOAD ||
+      type == BCMP_DFU_END ||
+      type == BCMP_DFU_ACK ||
+      type == BCMP_DFU_ABORT ||
+      type == BCMP_DFU_HEARTBEAT ||
+      type == BCMP_DFU_REBOOT_REQ ||
+      type == BCMP_DFU_REBOOT ||
+      type == BCMP_DFU_BOOT_COMPLETE ||
+      type == BCMP_ECHO_REPLY) {
```
victorsowa12: I'm not sure if all of these are replies, especially w.r.g. to the DFU messages.

victorsowa12: Woops I see I already left in the DFU_REBOOT_REQ - I'll remove it!

russelldeguzman: You can ignore all the DFU stuff for now since it's out of scope for this effort! (our just leave a TODO to audit the messages later

victorsowa12: right! I'll leave them out so they keep working as normal!

src/lib/bcmp/bcmp.cpp
victorsowa12: I haven't yet tested with 3+ nodes, but it looks like it will block the message form being forwarded! I think we just need to remove the break. I will test this out now!

victorsowa12: forwarding still works after removing the break.

PR #92, opened 2024-02-27
support logging of network configs

russelldeguzman:
```
2024-02-27T01:44:03.929Z [BRIDGE_CFG] [INFO] Node ID: 0 Partition: system Value: 60000
2024-02-27T01:44:03.933Z [BRIDGE_CFG] [INFO] Node ID: 0 Partition: system Value: 1
2024-02-27T01:44:03.941Z [BRIDGE_CFG] [INFO] Node ID: 0 Partition: system Value: 0
2024-02-27T01:44:03.949Z [BRIDGE_CFG] [INFO] Node ID: 0 Partition: hardware - Failed to decode value
Network configuration change detected! crc: 0xfd360c31 at UTC:1708998246
Bridge network config: 
[[8046209027809048845, "bridge-dbg", 3471088096, 2703802277, {"bridgePowerControllerEnabled": 1, "sampleIntervalMs": 300000, "sampleDurationMs": 180000, "subsampleEnabled": 0, "samplesPerReport": 1, "currentReadingPeriodMs": 60000, "alignmentInterval5Min": 1, "foo": 0}], [14453092352742153229, "bm_rbr-dbg", 4106273564, 81022053, {}]]

```
[0012_BRIDGE_CFG.log](https://github.com/bristlemouth/bm_protocol/files/14427010/0012_BRIDGE_CFG.log)

Linked to https://github.com/wavespotter/fleet-spotter-fw/pull/844

### DIFF
--- a/src/apps/bridge/sm_config_crc_list.cpp
+++ b/src/apps/bridge/sm_config_crc_list.cpp
@@ -69,6 +69,20 @@ void SMConfigCRCList::clear() {
   encode();
 }
 
+/*!
+  \brief Get the list of CRCs.
+
+  This function reads the list from the configuration.
+
+  \param[out] num_crcs[out] The number of CRCs in the list.
+  \return A pointer to the list of CRCs.
+*/
+uint32_t *SMConfigCRCList::get(uint32_t &num_crcs) {
+  decode();
+  num_crcs = _num_crcs;
+  return _crc_list;
+}
+
 // ---------------------------- PRIVATE ----------------------------
 
 /*!
@@ -175,18 +189,4 @@ void SMConfigCRCList::encode() {
     size_t buffer_size = cbor_encoder_get_buffer_size(&encoder, buffer);
     _cfg->setConfigCbor(KEY, KEY_LEN, buffer, buffer_size);
   } while (0);
-}
-
-/*!
-  \brief Get the list of CRCs.
-
-  This function reads the list from the configuration.
-
-  \param[out] num_crcs[out] The number of CRCs in the list.
-  \return A pointer to the list of CRCs.
-*/
-uint32_t *SMConfigCRCList::get(uint32_t &num_crcs) {
-  decode();
-  num_crcs = _num_crcs;
-  return _crc_list;
 }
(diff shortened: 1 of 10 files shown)


src/lib/common/topology_sampler.cpp
```diff
@@ -164,6 +169,7 @@ static void topology_sample_cb(networkTopology_t *networkTopology) {
         configASSERT(_node_list.last_network_configuration_info.cbor_config_map);
         memcpy(_node_list.last_network_configuration_info.cbor_config_map, cbor_buffer,
                cbor_bufsize);
+        log_cbor_network_configurations(cbor_buffer, cbor_bufsize);
```
russelldeguzman, line 176: We will log every time there is a network config change detected by the topology sampler

victorsowa12 (approved):
🪵

src/apps/bridge/sm_config_crc_list.cpp
```diff
@@ -69,6 +69,20 @@ void SMConfigCRCList::clear() {
...
+
+  This function reads the list from the configuration.
+
+  \param[out] num_crcs[out] The number of CRCs in the list.
+  \return A pointer to the list of CRCs.
+*/
+uint32_t *SMConfigCRCList::get(uint32_t &num_crcs) {
+  decode();
+  num_crcs = _num_crcs;
+  return _crc_list;
```
towynlin: So this makes me nervous because it provides a public API to get a pointer to a private member. Before this change there was no danger of an external user of this object modifying or corrupting the object itself, but this function opens the door for such mistakes in the future. This PR uses the returned pointer safely, but future folks may not be so careful.

I love the idea of logging the full list; that's very helpful for debugging.

Alternatives I can see easily are (1) allocating a copy of the list and requiring the caller to free, or (2) building the printed string internally and returning that, requiring the caller to free. Thoughts?

towynlin (changes requested):
Let's discuss the config crc list API change.

towynlin (changes requested):
Also, we should add unit tests for new public methods on the config crc list.

src/apps/bridge/sm_config_crc_list.cpp
russelldeguzman: Good point! I'll copy!

towynlin (approved):
Woot! ❤️ Perfect. 👍

PR #93, opened 2024-02-28
Feature/bm serial get bcmp cfgs

victorsowa12:
Using the node_id and adding callbacks to handle config req/rep! This allows Spotter to set/get/del configs from its CLI. I also added that we call the callback upon bcmp request message timeout. So I made sure the callbacks can handle a NULL payload ~however, for the handling of a failed request right now I am just doing a bm_printf~. I changed it to use a `BRIDGE_CFG_LOG_PRINT` of the failure.

Example from Spotter's CLI:
```
bm topo
Bristlemouth toplogy: 4d58ce2f22dfb374 | 910030066a2926a7 | 193e6faff528b841
```

```
bm cfg status 910030066a2926a7 s
Succesfull status request send
2024-02-28T22:09:00.625Z [BRIDGE_CFG] [INFO] Response msg -- Node Id:910030066a2926a7,Partition:system, Commit Status:0
2024-02-28T22:09:00.625Z [BRIDGE_CFG] [INFO] Num Keys: 1
2024-02-28T22:09:00.625Z [BRIDGE_CFG] [INFO] Key 0: sensorsPollIntervalMs
2024-02-28T22:09:00.640Z [BRIDGE_CFG] [INFO] Node ID: 910030066a2926a7 Partition: system Value: 0

bm cfg get 910030066a2926a7 s sensorsPollIntervalMs
Succesfully sent config get msg
2024-02-28T22:09:11.515Z [BRIDGE_CFG] [INFO] Node ID: 910030066a2926a7 Partition: system Value: 0

bm cfg status 193e6faff528b841 s
Succesfull status request send
2024-02-28T22:09:17.617Z [BRIDGE_CFG] [INFO] Response msg -- Node Id:193e6faff528b841,Partition:system, Commit Status:0
2024-02-28T22:09:17.617Z [BRIDGE_CFG] [INFO] Num Keys: 1
2024-02-28T22:09:17.617Z [BRIDGE_CFG] [INFO] Key 0: sensorsPollIntervalMs
2024-02-28T22:09:17.632Z [BRIDGE_CFG] [INFO] Node ID: 193e6faff528b841 Partition: system Value: 0

bm cfg get 193e6faff528b841 s sensorsPollIntervalMs
Succesfully sent config get msg
2024-02-28T22:09:40.687Z [BRIDGE_CFG] [INFO] Node ID: 193e6faff528b841 Partition: system Value: 0

bm cfg set 910030066a2926a7 s u sensorsPollIntervalMs 5000
Succesfully sent config set msg
2024-02-28T22:14:33.937Z [BRIDGE_CFG] [INFO] Node ID: 910030066a2926a7 Partition: system Value: 5000

bm cfg commit 910030066a2926a7 s
Succesfull config commit send

bm cfg get 910030066a2926a7 s sensorsPollIntervalMs
Succesfully sent config get msg
2024-02-28T22:15:15.882Z [BRIDGE_CFG] [INFO] Node ID: 910030066a2926a7 Partition: system Value: 5000

bm cfg del 910030066a2926a7 s sensorsPollIntervalMs
Successfully sent del key request
2024-02-28T22:17:04.757Z [BRIDGE_CFG] [INFO] Response msg -- Node Id:910030066a2926a7,Partition:system, Key: sensorsPollIntervalMs, Success: 1

bm cfg commit 910030066a2926a7 s
Succesfull config commit send

bm cfg status 910030066a2926a7 s
Succesfull status request send
2024-02-28T22:17:48.453Z [BRIDGE_CFG] [INFO] Response msg -- Node Id:910030066a2926a7,Partition:system, Commit Status:0
2024-02-28T22:17:48.453Z [BRIDGE_CFG] [INFO] Num Keys: 0
```

### DIFF
--- a/src/lib/bcmp/bcmp.cpp
+++ b/src/lib/bcmp/bcmp.cpp
@@ -622,6 +622,9 @@ static void _message_list_timer_expiry_cb(void *arg) {
     while (current) {
       if (pdTICKS_TO_MS(xTaskGetTickCount()) - current->send_timestamp_ms > current->timeout_ms) {
         printf("BCMP message with seq_num %d timed out\n", current->seq_num);
+        if (current->callback) {
+          current->callback(NULL);
+        }
         _message_list_remove_message(current);
         current = _ctx.messages_list;
         continue;
--- a/src/lib/bm_ncp/ncp_config.cpp
+++ b/src/lib/bm_ncp/ncp_config.cpp
@@ -3,7 +3,7 @@
 #include "device_info.h"
 #include "FreeRTOS.h"
 #include "bm_serial.h"
-#include "bm_printf.h"
+#include "bridgeLog.h"
 
 using namespace cfg;
 
@@ -210,7 +210,7 @@ bool _cfg_get_bcmp_cb(uint8_t *payload) {
             rval = false;
         }
     } else {
-        bm_printf(0, "Failed to get cfg\n");
+        BRIDGE_CFG_LOG_PRINT("Failed to get cfg\n");
     }
     return rval;
 }
@@ -224,7 +224,7 @@ bool _cfg_set_bcmp_cb (uint8_t *payload) {
             rval = false;
         }
     } else {
-        bm_printf(0, "Failed to set cfg\n");
+        BRIDGE_CFG_LOG_PRINT("Failed to set cfg\n");
     }
     return rval;
 }
@@ -239,7 +239,7 @@ bool _cfg_status_request_bcmp_cb(uint8_t *payload) {
             rval = false;
         }
     } else {
-        bm_printf(0, "Failed to get cfg status\n");
+        BRIDGE_CFG_LOG_PRINT("Failed to get cfg status\n");
     }
     return rval;
 }
@@ -253,7 +253,7 @@ bool _cfg_key_del_bcmp_cb(uint8_t *payload) {
             rval = false;
         }
     } else {
-        bm_printf(0, "Failed to delete cfg\n");
+        BRIDGE_CFG_LOG_PRINT("Failed to delete cfg\n");
     }
     return rval;
 }
--- a/src/lib/bm_ncp/ncp_config.cpp
+++ b/src/lib/bm_ncp/ncp_config.cpp
@@ -24,10 +24,11 @@ bool _cfg_key_del_bcmp_cb(uint8_t *payload);
 
 bool ncp_cfg_get_cb(uint64_t node_id, bm_common_config_partition_e partition, size_t key_len, const char* key) {
     bool rval = false;
-    if (node_id == getNodeId() || node_id == 0) {
-        uint8_t * valueBuf = reinterpret_cast<uint8_t*>(pvPortMalloc(MAX_CONFIG_BUFFER_SIZE_BYTES));
-        configASSERT(valueBuf);
-        do {
+    uint8_t *valueBuf = NULL;
+    do {
+        if (node_id == getNodeId() || node_id == 0) {
+            valueBuf = reinterpret_cast<uint8_t*>(pvPortMalloc(MAX_CONFIG_BUFFER_SIZE_BYTES));
+            configASSERT(valueBuf);
             Configuration* p = get_partition(partition);
             if(!p) {
                 printf("Invalid partition\n.");
@@ -42,13 +43,19 @@ bool ncp_cfg_get_cb(uint64_t node_id, bm_common_config_partition_e partition, si
                 printf("Failed to send cfg\n");
                 break;
             }
-            rval = true;
-        } while(0);
-        vPortFree(valueBuf);
-    } else {
-        err_t err = ERR_OK;
-        bcmp_config_get(node_id, partition, key_len, key, err, _cfg_get_bcmp_cb);
+        } else {
+            err_t err = ERR_OK;
+            if (!bcmp_config_get(node_id, partition, key_len, key, err, _cfg_get_bcmp_cb)) {
+                static char buf[64];
+                snprintf(buf, sizeof(buf), "Failed to issue bcmp cfg get, err: %" PRIi8 "\n", err);
+                BRIDGE_CFG_LOG_PRINT(buf);
+                break;
+            }
+        }
         rval = true;
+    } while(0);
+    if (valueBuf != NULL) {
+        vPortFree(valueBuf);
     }
     return rval;
 }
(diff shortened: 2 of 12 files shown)


src/lib/bm_ncp/ncp_config.cpp
```diff
@@ -165,3 +200,60 @@ static uint8_t* alloc_ncp_key_buffer(uint8_t num_keys, const ConfigKey_t* keys,
...
+bool _cfg_get_bcmp_cb(uint8_t *payload) {
+    bool rval = true;
+    if (payload != NULL) {
+        bm_common_config_value_t *msg = reinterpret_cast<bm_common_config_value_t *>(payload);
+        if (bm_serial_cfg_value(msg->header.source_node_id, msg->partition, msg->data_length, msg->data) != BM_SERIAL_OK) {
+            printf("Failed to send get cfg response\n");
+            rval = false;
+        }
+    } else {
+        bm_printf(0, "Failed to get cfg\n");
```
victorsowa12: right now I am just printing out that we faile to get/set/status requests, but this ends up looking like this:
```
bm cfg get 910030066a2926a7 s sensorsPollIntervalMs
Succesfully sent config get msg
1709160098.746 4d58ce2f22dfb374, Failed to get cfg
```
which is not to pretty and doesn't make sense. So I am thinking that we should create a known "invalid" Node-ID that we send here with empty/garbage data. Then Spotter will know that if it recieves the "invalid" Node-ID, to print that the cfg get/set/status request timedout and failed. Thoughts? How about 0xffffffffffffffff as the "invalid" NODE-ID?

victorsowa12: Another option would be to store more data in the bcmp message requests linked list and assemble a response message, however, not all of the requests have a built in success/fail boolean.

That is why I thought of the "invalid" node-id since that is a common shared variable in each of the messages.

victorsowa12: Using `bm_printf` is fine for now, but in the future we should have a shared way of handling successful config requests and failed config requests.

victorsowa12: I'm going to switch it to use `BRIDGE_CFG_LOG_PRINT`

src/lib/bcmp/bcmp.cpp
```diff
@@ -622,6 +622,9 @@ static void _message_list_timer_expiry_cb(void *arg) {
     while (current) {
       if (pdTICKS_TO_MS(xTaskGetTickCount()) - current->send_timestamp_ms > current->timeout_ms) {
         printf("BCMP message with seq_num %d timed out\n", current->seq_num);
+        if (current->callback) {
+          current->callback(NULL);
+        }
```
victorsowa12, line 627: new addition, we call the callback when the timer expires and pass it a NULL payload

src/lib/bm_ncp/ncp_config.cpp
```diff
@@ -3,46 +3,60 @@
...
 static Configuration* get_partition(bm_common_config_partition_e partition);
 static uint8_t* alloc_ncp_key_buffer(uint8_t num_keys, const ConfigKey_t* keys, size_t& len);
 
+// These are the callback functions that BCMP req/rep will call when we get a response
+// from the target node.
+bool _cfg_get_bcmp_cb(uint8_t *payload);
+bool _cfg_set_bcmp_cb (uint8_t *payload);
+bool _cfg_status_request_bcmp_cb(uint8_t *payload);
+bool _cfg_key_del_bcmp_cb(uint8_t *payload);
+
```
victorsowa12, line 24: these are the new callbacks for when we do a cfg request for a node id that is not the bridges. This file is the bulk of the additions to enable `bm cfg` for any node via Spotter

victorsowa12:
I realized since I had to merge develop into the `feature/bm_serial_get_bcmp_cfgs` branch for the BRIDGE_CFG_LOG_PRINT function there are a bunch of changes on develop that are now also being merged into `feature/bcmp_req_rep`. So I highlighted the two files to look at with comments, but those are `bcmp.cpp` and `ncp_config.cpp`.

russelldeguzman (approved):
Awesome! This looks good!

src/lib/bm_ncp/ncp_config.cpp
```diff
@@ -3,46 +3,60 @@
...
+            if(bm_serial_cfg_value(node_id, partition, value_len, valueBuf) != BM_SERIAL_OK) {
+                printf("Failed to send cfg\n");
+                break;
+            }
+            rval = true;
+        } while(0);
+        vPortFree(valueBuf);
+    } else {
+        err_t err = ERR_OK;
+        bcmp_config_get(node_id, partition, key_len, key, err, _cfg_get_bcmp_cb);
```
towynlin: Should we check rval and err here?

towynlin (approved):
Looks good! Thank you. Had one question about error checking.

I'd appreciate a quick comment explaining the flow here, especially in ncp_config.cpp. I see multiple sets of callbacks (functions with a `_cb` suffix) some previously existing and some new. The new ones start with an underscore.

For example, here's my quick guess without diving too deep... 

When we perform a `bm cfg get` via the Spotter USB, does `bm_serial` call `ncp_cfg_get_cb` in the bridge? And perhaps that used to be the only place that would be called, so it was unconditional. Now you've wrapped that preexisting code in "if 0 or my node id", otherwise... call `bcmp_config_get` so that the bridge passes the request to the rest of the bristlemouth network?

Ah, and maybe the new underscore callback `_cfg_get_bcmp_cb` (because of the new bcmp req-rep capability on this branch) then gets called either with null in the case of a timeout or with a response payload.

If there's a valid response payload we call `bm_serial_cfg_value` which perhaps passes the response over `bm_serial` back to spotter?

Does that sound right? Please correct any misconceptions.

towynlin (approved):
Looks good! Thank you. Had one question about error checking.

I'd appreciate a quick comment explaining the flow here, especially in ncp_config.cpp. I see multiple sets of callbacks (functions with a `_cb` suffix) some previously existing and some new. The new ones start with an underscore.

For example, here's my quick guess without diving too deep... 

When we perform a `bm cfg get` via the Spotter USB, does `bm_serial` call `ncp_cfg_get_cb` in the bridge? And perhaps that used to be the only place that would be called, so it was unconditional. Now you've wrapped that preexisting code in "if 0 or my node id", otherwise... call `bcmp_config_get` so that the bridge passes the request to the rest of the bristlemouth network?

Ah, and maybe the new underscore callback `_cfg_get_bcmp_cb` (because of the new bcmp req-rep capability on this branch) then gets called either with null in the case of a timeout or with a response payload.

If there's a valid response payload we call `bm_serial_cfg_value` which perhaps passes the response over `bm_serial` back to spotter?

Does that sound right? Please correct any misconceptions.

towynlin:
Sorry for the duplicate comment... something to do with using github's in-browser code editor to do the review. 🤷

src/lib/bm_ncp/ncp_config.cpp
victorsowa12: Yes, I will add that in!

victorsowa12:
Yes that flow is correct! 

> does `bm_serial` call `ncp_cfg_get_cb` in the bridge? And perhaps that used to be the only place that would be called, so it was unconditional. 

This is still true, `ncp_cfg_get_cb` is still only called from `bm_serial`. The `if node_id == 0 || node_id == getNodeID()` just decides whether we respond directly with the bridges configs, or send out a request to another node using `bcmp_config_get`. The "`_cb`" is used to tell `bcmp_config_get` that when it receives a reply, forward the reply via `bm_serial` back to Spotter.

towynlin:
I see, yeah, perfect, thank you. So really — the core feature we were aiming for is in that else clause. It used to ignore the node id and only work for the bridge directly. Now we actually check the node id, and if it's for someone else, we make the bcmp request. ✨

PR #94, opened 2024-02-29
Feature/bcmp req rep

victorsowa12:
Merge the `feature/bcmp_req_rep` branch into develop.

This will be a breaking change, where devices flashed before and after this change will NOT be able to talk BCMP to one another. All of the nodes (bridge, motes etc) will need to be updated.

### DIFF
--- a/src/lib/bcmp/bcmp.cpp
+++ b/src/lib/bcmp/bcmp.cpp
@@ -2,6 +2,7 @@
 #include "FreeRTOS.h"
 #include "task.h"
 #include "queue.h"
+#include "semphr.h"
 #include "timers.h"
 
 #include "bcmp.h"
@@ -21,6 +22,7 @@
 #include "bcmp_time.h"
 #include "bcmp_topology.h"
 #include "bcmp_resource_discovery.h"
+#include "timer_callback_handler.h"
 
 #include "bm_dfu.h"
 
@@ -37,12 +39,27 @@
 // and their sequence numbers and the callback to call
 // upon receipt of the reply.
 // The callback should also handle the timeout case.
+static constexpr uint32_t DEFAULT_MESSAGE_TIMEOUT_MS = 100;
+static constexpr uint32_t MESSAGE_TIMER_EXPIRY_PERIOD_MS = 500;
+
+typedef struct bcmp_element {
+  uint32_t seq_num;
+  uint16_t type;
+  uint32_t send_timestamp_ms;
+  uint32_t timeout_ms;
+  bcmp_element *next;
+} bcmp_element_t;
 
 typedef struct {
   struct netif* netif;
   struct raw_pcb *pcb;
   QueueHandle_t rx_queue;
   TimerHandle_t heartbeat_timer;
+  // Linked list of sent BCMP messages
+  bcmp_element_t *messages_list;
+  SemaphoreHandle_t messages_list_mutex;
+  TimerHandle_t messages_expiration_timer;
+  uint32_t message_count;
 } bcmpContext_t;
 
 typedef enum {
@@ -63,6 +80,14 @@ typedef struct {
 } bcmp_queue_item_t;
 
 static bcmpContext_t _ctx;
+
+static bool _message_list_add_message(bcmp_element_t *message);
+static bool _message_list_remove_message(bcmp_element_t *message);
+static bcmp_element_t *_message_list_create_element(uint16_t seq_num, uint16_t type, uint32_t timeout_ms);
+static void _message_list_timer_callback(TimerHandle_t tmr);
+static bcmp_element_t *_message_list_find_message(uint16_t seq_num, uint16_t type);
+static void _message_list_timer_expiry_cb(void *arg);
+static bool _message_uses_sequence_numbers(uint16_t type);
 
 /*!
   BCMP link change event callback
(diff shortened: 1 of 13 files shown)


src/lib/bm_ncp/ncp_config.cpp
```diff
@@ -3,77 +3,116 @@
...
+                BRIDGE_CFG_LOG_PRINT(buf);
+                break;
+            }
         }
         rval = true;
     } while(0);
-    vPortFree(valueBuf);
+    if (valueBuf != NULL) {
+        vPortFree(valueBuf);
+    }
```
victorsowa12, line 59: I edited these such that both bridge cfg requests and other node cfg requests fall into the do-while(0) statement and if they error they break out of the same do-while(0). I also added the checking of `err` and the return value for the `bcmp_cfg_...` requests.

src/lib/bcmp/bcmp.cpp
```diff
@@ -32,11 +34,27 @@
 // 1500 MTU minus ipv6 header
 #define MAX_PAYLOAD_LEN (1500 - sizeof(struct ip6_hdr))
 
+static constexpr uint32_t MESSAGE_TIMER_EXPIRY_PERIOD_MS = 500;
```
towynlin: How did we arrive at this number? It seems to me that it ought to be much smaller. If I imagine for example DFU or lots of video, something with hundreds of requests queued up and waiting half a second for each one once a failure happens, I imagine we might use up all heap and crash before this timeout, or else suddenly everything goes suuuuper sloooow... I don't know for certain, just wondering what our constraints are on this timeout.

src/lib/bcmp/bcmp.cpp
```diff
@@ -97,7 +124,7 @@ static void dfu_copy_and_process_message(struct pbuf *pbuf) {
   \param *dst packet destination
   \return 0 if processed ok, nonzero otherwise
 */
-int32_t bmcp_process_packet(struct pbuf *pbuf, ip_addr_t *src, ip_addr_t *dst) {
+int32_t bcmp_process_packet(struct pbuf *pbuf, ip_addr_t *src, ip_addr_t *dst) {
```
towynlin, line 127: 😂 thank you

src/lib/bcmp/bcmp.cpp
```diff
@@ -484,3 +556,124 @@ void bcmp_init(struct netif* netif, NvmPartition * dfu_partition, Configuration*
...
+    }
+    rval = true;
+    xSemaphoreGive(_ctx.messages_list_mutex);
+  }
+  return rval;
+}
+
+// NOTE: This function is not thread safe and can only be called within a section
+// protected by the messages_list_mutex semaphore.
+static bool _message_list_remove_message(bcmp_request_element *message) {
```
towynlin, line 582: The other message list modification functions manage the semaphore internally themselves. Why do we require the caller to manage the semaphore for this one?

towynlin (approved):
Had a few questions. Looks good!

src/lib/bcmp/bcmp.cpp
victorsowa12: `_message_list_remove_message` is called from a loop within `_message_list_timer_expiry_cb` where we have already taken the mutex.

src/lib/bcmp/bcmp.cpp
victorsowa12: I just set `MESSAGE_TIMER_EXPIRY_PERIOD_MS` = 500 and `DEFAULT_MESSAGE_TIMEOUT_MS` = 1000 for testing, but yes you are right they could/should be much smaller. I see for bm service req/reply:
```
static constexpr uint32_t DEFAULT_SERVICE_REQUEST_TIMEOUT_MS = 100;
static constexpr uint32_t EXPIRY_TIMER_PERIOD_MS = 500;
```
I could go with these or even smaller?

russelldeguzman: One constraint is we don't know the RTT of a message N hops away. This is a bit of a goldilocks problem and make sure we pick something big enough for N=5 maybe?

src/lib/bcmp/bcmp.cpp
towynlin: Ah, got it. I see that. Thanks, and yeah, we want to remove all expired messages in one pass rather than potentially having to wait for the semaphore in the middle of the loop. 👍

src/lib/bcmp/bcmp.cpp
russelldeguzman: If we have log print time stamps we can use that as an upper limit and just multiply that out by the number of N hops

towynlin: Victor, could you please test that and gather some data? Make the longest Bristlemouth chain you easily can with materials at hand. Send a bunch of bcmp cfg get messages as fast as you can from one end of the chain to the other. Gather some stats like min, mean, max, std.

russelldeguzman (approved):

src/lib/bcmp/bcmp.cpp
victorsowa12: on it!

russelldeguzman: Totally not for this PR, but we should probably open a separate characterization ticket, especially when we do DFU - which has a 1024 flash write for each message.

victorsowa12: For a system of `bridge - soft - soft - rbr - normal mote`, getting the normal motes (4th mote) configs resulted in:
```
Average difference: 3.5132743362831858 ticks
Standard deviation of differences: 0.5009332490069597 ticks
Minimum difference: 3 ticks
Maximum difference: 4 ticks
Number of differences: 226
```
or about 1 ms per node.

victorsowa12: will go with 12 ms timer, and 24 ms timeout. This is because we can support up to 16 motes so 16 ms + some padding.

towynlin: After IRL discussion: let's go with 12 ms timer and 24 ms message timeout.

russelldeguzman (changes requested):
Locking until we have a dfu fix

russelldeguzman (approved):
Just tested updating a bm soft as well as a fake aanderaa with the `dfu_confirm` flag  set to 0 and looks like things are working well!

PR #95, opened 2024-03-01
feat: sofar sensor topics

towynlin:
This PR changes the pub-sub topic of each of the 3 managed sensors sold by Sofar Ocean. It prepends `sensor/` and inserts `/sofar` in the  middle.

For example, before this change, a Sofar current meter module might publish readings to the topic `142da2ea9eccbe3c/aanderaa`. After this change, it would publish to `sensor/142da2ea9eccbe3c/sofar/aanderaa`.

We're envisioning a future with many companies making similarly named sensors and other modules. We want to create a reasonable naming system that folks can follow as we all build the ecosystem together, where we don't need to fight for top-level namespace.

Plus, even though it doesn't work yet, we're migrating toward a scheme where folks could create their own general-purpose aggregators that could subscribe to all sensor readings using a wildcard like `sensor/*`.

This will be a breaking change of sorts. Both the publishing sensor module and the subscribing bridge module have to be updated with this change. If only one is updated, then no sensor data will be sent via Spotter's telemetry.


victorsowa12 (approved):
👍

russelldeguzman (approved):

PR #96, opened 2024-03-01
Fix/toplogy node position log

victorsowa12:
This will only update the node-position in IND logs when we have a large gap between individual readings (indicating a potential unplugging event and replugging in a different order) or when we have the first reading. This change is to reduce the amount we have to take the node_list mutex and potentially wait for it, while maintaining a consistent node_position in the IND logs.

### DIFF
--- a/src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
@@ -53,7 +53,22 @@ void AanderaaSensor::aanderaSubCallback(uint64_t node_id, const char *topic, uin
         uint64_t sensor_reading_time_sec = d.header.sensor_reading_time_ms / 1000U;
         uint32_t sensor_reading_time_millis = d.header.sensor_reading_time_ms % 1000U;
 
-        int8_t node_position = topology_sampler_get_node_position(node_id, pdTICKS_TO_MS(5000));
+        // Keep track of when we last got a reading from this node
+        // If it is the first reading, or if it has been more than the current aggregation period + 30 seconds
+        // then we will update the node position.
+        static int8_t node_position = 0;
+        static uint32_t last_timestamp = 0;
+        static uint32_t current_timestamp = 0;
+
+        current_timestamp = pdTICKS_TO_MS(xTaskGetTickCount());
+        if ((current_timestamp - last_timestamp > aanderaa->current_agg_period_ms + 30000u) ||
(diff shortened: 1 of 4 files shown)


src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
```diff
@@ -53,7 +53,7 @@ void AanderaaSensor::aanderaSubCallback(uint64_t node_id, const char *topic, uin
         uint64_t sensor_reading_time_sec = d.header.sensor_reading_time_ms / 1000U;
         uint32_t sensor_reading_time_millis = d.header.sensor_reading_time_ms % 1000U;
 
-        int8_t node_position = topology_sampler_get_node_position(node_id, 1000);
+        int8_t node_position = topology_sampler_get_node_position(node_id, pdTICKS_TO_MS(5000));
```
victorsowa12: I changed this to 5 seconds since now we will try to get the node position and take the mutex infrequently and thus we want to make sure we actually get it and don't log an invalid value (unless we know for sure it is not in our topology).

victorsowa12:
I will soak overnight and see if any 255 node positions show up and then mark it ready for review.

victorsowa12:
There were no instances of a node position of 255 in the overnight soak!

russelldeguzman (approved):

src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
```diff
@@ -53,7 +53,22 @@ void AanderaaSensor::aanderaSubCallback(uint64_t node_id, const char *topic, uin
...
+        // If it is the first reading, or if it has been more than the current aggregation period + 30 seconds
+        // then we will update the node position.
+        static int8_t node_position = 0;
+        static uint32_t last_timestamp = 0;
+        static uint32_t current_timestamp = 0;
+
+        current_timestamp = pdTICKS_TO_MS(xTaskGetTickCount());
+        if ((current_timestamp - last_timestamp > aanderaa->current_agg_period_ms + 30000u) ||
+            aanderaa->reading_count == 1U) {
+          printf("Updating rbr_coda %" PRIx64 " node position, current_time = %" PRIu32
```
towynlin: This file is named `aanderaaSensor.cpp` and this log mentions `rbr_coda`. Copy pasta?

towynlin (commented):
One tiny typo to fix. Otherwise 👍

PR #97, opened 2024-03-01
Feature/dfu confirm workaround

russelldeguzman:
To future-proof BCMP/comms layer protocol update incompatibility, we are adding a NVM Config force flag to enable users to set `dfu_confirm` to `0`, which will allow the DFU client to autoconfirm the update without the need for an ACK back to it's DFU host.  

Tested on BM SOFT:
1. Downgrade bridge + mote to v0.9.1
2. Set the "dfu_confirm" sys config on the mote to 0
3. update mote to new firmware 
See that the nodes can't quite communicate properly (on soft the pubsub still works for some reason??)
But the heartbeat and topo is breaking 
4. update bridge to new firmware 
5. profit

Tested that in the case `dfu_confirm` is absent, the image reverts.

Remember to reset the 
`dfu_confirm` config  between attempts

A side effect of disabling confirm that when the update completes, the host will get timeout error on it's end, even though the update was successfull


towynlin (approved):
🚀

PR #98, opened 2024-03-04
Sc 198025/log level to bridge

russelldeguzman:
Bridge printing refactor

Does a couple things:
* Now, instead of pub/subbing a raw string, we add a header via `bm_common_messages` with metadata on how Spotter should process the log (including level, header, etc)
* We now use a variadic style for bridge printing which greatly simplifies calls to print, and also removes memory buffers and some mallocs.  

Should be coupled with https://github.com/wavespotter/fleet-spotter-fw/pull/847

```
110750t [BRIDGE_CFG] [INFO] Bridge network config:
[[8046209027809048845, "bridge-dbg", 692888193, 2703802277, {"bridgePowerControllerEnabled": 1, "sampleIntervalMs": 300000, "sampleDurationMs": 180000, "subsampleEnabled": 0, "samplesPerReport": 1, "currentReadingPeriodMs": 60000, "alignmentInterval5Min": 1, "foo": 0}]]
111726t [BRIDGE_CFG] [INFO] Bridge network info CRC: 5f8a3e26
111726t [BRIDGE_CFG] [INFO] Bridge FW Version: Major: 0 Minor: 9 Revision: 1
111727t [BRIDGE_CFG] [INFO] Bridge sys config partition CRC: a128bfa5
111727t [BRIDGE_CFG] [INFO] Bridge toplogy: 6fa9e0
```


src/apps/bridge/app_main.cpp
```diff
@@ -89,322 +89,323 @@ static void defaultTask(void *parameters);
...
 static void neighborDiscoveredCb(bool discovered, bm_neighbor_t *neighbor) {
-    configASSERT(neighbor);
-    static constexpr size_t bufsize = 50;
-    const char * action = (discovered) ? "added" : "lost";
-    char buffer[bufsize];
-    size_t len = snprintf(buffer, bufsize, "Neighbor %" PRIx64 " %s\n", neighbor->node_id, action);
-    BRIDGE_LOG_PRINTN(buffer, len);
+  configASSERT(neighbor);
+  const char *action = (discovered) ? "added" : "lost";
+  bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER, "Neighbor %" PRIx64 " %s\n",
```
russelldeguzman, line 297: Only new thing in this file, the rest is `clang-format`

victorsowa12 (approved):
🖨️

src/apps/bridge/bridgeLog.cpp
```diff
@@ -1,29 +1,59 @@
...
+    }
+    print_size++; // Add 1 for null terminator
+    size_t msg_len = sizeof(bm_common_log_t) + print_size;
+    log_msg = static_cast<bm_common_log_t *>(pvPortMalloc(msg_len));
+    configASSERT(log_msg);
+    memset(log_msg, 0, msg_len);
+    log_msg->level = level;
+    log_msg->message_length = print_size;
+    log_msg->print_header = print_header;
+    vsnprintf(log_msg->message, print_size, format, va_args);
```
towynlin: I see that the `timestamp_utc_s` field isn't set. Is that intentional, or should we set it?

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -91,23 +84,19 @@ void BridgePowerController::powerControlEnable(bool enable) {
   } else {
     printf("Bridge power controller disabled\n");
   }
-  xTaskNotify(
-      _task_handle, enable,
-      eNoAction); // Notify the power controller task that the state has changed.
+  xTaskNotify(_task_handle, enable,
+              eNoAction); // Notify the power controller task that the state has changed.
```
towynlin: nit: move comment to it's own line

src/apps/bridge/bridgeLog.cpp
```diff
@@ -1,29 +1,59 @@
...
+  va_end(va_args);
+}
+
+void vBridgeLogPrint(bridgeLogType_e type, bm_common_log_level_e level, bool print_header,
+                     const char *format, va_list va_args) {
+  bm_common_log_t *log_msg = NULL;
+  do {
+    int print_size = vsnprintf(nullptr, 0, format, va_args);
+    if (print_size < 0) {
+      printf("ERROR: vsnprintf failed in bridgeLogPrint\n");
```
towynlin: I think the error should be printed to the log indicated by `type`. Otherwise, we'll probably never know this happens. I know it'll be rare, so if you disagree, let me know. At least I see one case in this PR where we're dropping a previous failure — in topology sampler's `cborStreamFunct`. I was trying to figure out how we could catch that, if the cbor ends up being too big to print.

towynlin (changes requested):
Outstanding work! Thank you! This is a high-ROI refactor for sure. Great job with the variadic print function!

I have one change request at least for discussion, then one question, and one nit.

src/apps/bridge/bridgeLog.cpp
russelldeguzman: Good catch we should set it! (we currently don't use it at the moment, but we may want to later)

towynlin (approved):

PR #100, opened 2024-03-05
Bugfix/empty list bug

russelldeguzman:
Brand new Bridges will not have a list of CRCs

Make an empty crc list a non crashing thing.


victorsowa12 (approved):
Good find!

towynlin (approved):
Ah! Nice. Great find.

PR #101, opened 2024-03-06
fix: soft calibration offset

towynlin:


src/apps/bm_soft_module/user_code/user_code.cpp
```diff
@@ -41,49 +46,59 @@ static bool BmSoftStartAndValidate(void);
...
-    bm_fprintf(0, soft_log, "No calibration temperature (milliDegC) found\n");
-    bm_printf(0, "No calibration temperature (milliDegC) found");
+    printf("No calibration offset (milliDegC) found\n");
+    bm_fprintf(0, soft_log, "No calibration offset (milliDegC) found\n");
+    bm_printf(0, "No calibration offset (milliDegC) found");
   }
-  calibrationOffsetDegC = (calTempC + (calTempMilliC / 1000.0f));
-  if (userConfigurationPartition->getConfig("softReadingPeriodMs",
-                                            strlen("softReadingPeriodMs"), soft_delay_ms)) {
+  calibrationOffsetDegC = static_cast<float>(calOffsetMilliC) * 1e-3f;
```
russelldeguzman, line 87: Doesn't this need to incorporate the `calTempC` term?

russelldeguzman (changes requested):

russelldeguzman:
Regarding the polarity of the calibration term - here are some values of the offsets from the BoW test.
```
Timestamp,Timestamp(utc_s),SerialNumber,OffsetValueDegC
2023-12-25_11:11:11,1600000,398804,-0.03876
2023-12-25_11:11:11,1600000,398804,-0.0590343
2023-01-11_13:15:33/,1673471733.9574752,399504,0.15164988416988404
2023-01-11_13:15:33/,1673471733.9574752,401239,0.09835451737451692
```
If the TSYS reads higher than the RBR it's being calibrated against, it will read a negative value
sst_calibration_raw_plot
Yields the offsets:
```
{
    "423709": -0.06944950269506993,
    "423398": -0.1281154421201235,
    "423665": -0.11886492877309844,
    "423712": -0.04740188975872664,
    "423764": -0.07272016491273071
}
```
which we should add back to each reading to get the calibrated reading.

and the inverse for reading below the reference RBR

src/apps/bm_soft_module/user_code/user_code.cpp
towynlin: From my reading of the spec, no. It seems like `calTempC` is just informational. @evanShap could you please give us a reality check on what precisely gets programmed into the calibration offset config and how it should be used?

towynlin: The spec has this note:
> Note - calibration offset is stored such that: true_temp = reading_temp + calibration_offset.

towynlin: TSYS01.cpp line 76 is where the offset gets added

russelldeguzman: Got it - if this is true we should update Mike with this info.

russelldeguzman: Just saw the Notes section of the config! Approving!

russelldeguzman (approved):

russelldeguzman:
Nice find! 🎉

victorsowa12 (approved):

src/apps/bm_soft_module/user_code/user_code.cpp
```diff
@@ -41,49 +46,59 @@ static bool BmSoftStartAndValidate(void);
...
+    printf("No TSYS serial number found\n");
+    bm_fprintf(0, soft_log, "No TSYS serial number found\n");
+    bm_printf(0, "No TSYS serial number found");
+  }
+
   int32_t calTempC = 0;
-  int32_t calTempMilliC = 0;
-  if (sysConfigurationPartition->getConfig("calTempC", strlen("calTempC"), calTempC)) {
-    printf("calTempC: %" PRId32 "\n", calTempC);
+  if (sysConfigurationPartition->getConfig(soft_cfg_cal_temp_c, strlen(soft_cfg_cal_temp_c),
```
russelldeguzman, line 60: This runs contrary to the product spec, but if this is an informational term, I'd advocate this should be an `RO-app-write` and thus a `configSet` post computation of the milliDegC and reflect the system rather than a `R/W` by the calibrator.

evanShap: @russelldeguzman @towynlin - I'm having trouble following the thread here.
This just seems like a debug print of the calTemp value. Can you clarify the conflict with the product spec?

russelldeguzman: @evanShap The table shows R/W mutability on the spec at the moment - is the intention of this value to be set by the app? or be written at calibration time?

towynlin: I confirmed with Evan that this is set by the production tooling when calibration is done. It's just information about our calibration process.

evanShap: @towynlin @russelldeguzman Yep - I added more description to the Notes column for all of these.

Additional note – the R/W mutability column is a future-facing spec. I didn't expect we have a clean way of enforcing these currently, and that's not required. So today R/W describes intended behavior, but not an expectation we'll actually prevent writes via `bm cfg` commands to RO configs.

PR #104, opened 2024-03-07
Fix/node position

victorsowa12:
I recently updated the Node position code such that it only updates every once in a while. What I forgot is that the `softSubCallback` function is a static function and therefore called by all SoftSensor class objects. I had made the `node_position` variable a static variable within the `softSubCallback` thinking that each class object had its own `softSubCallback`. But what this really meant was that the `node_position` for one node was being used by every node. So in this PR I made the `node_position` and `last_timestamp` part of the `Soft_t`, `RbrCoda_t`, and `Aanderaa_t` classes so that each subscription tracks its own data.

Here is an example of lines in a log captured during a test where we have a node_position for one node override every other nodes positions:
```
48d57401250b4d7c,1,soft,678,0.000,0.000,22.351
246f92446ac231c3,2,soft,673,0.000,0.000,22.356
48d57401250b4d7c,2,soft,1198,0.000,0.000,22.351
246f92446ac231c3,2,soft,1193,0.000,0.000,22.356
48d57401250b4d7c,2,soft,1718,0.000,0.000,22.351
246f92446ac231c3,2,soft,1713,0.000,0.000,22.356
48d57401250b4d7c,2,soft,2238,0.000,0.000,22.351
246f92446ac231c3,2,soft,2233,0.000,0.000,22.356
48d57401250b4d7c,2,soft,2758,0.000,0.000,22.351
246f92446ac231c3,2,soft,2753,0.000,0.000,22.356
48d57401250b4d7c,2,soft,3278,0.000,0.000,22.351
246f92446ac231c3,2,soft,3273,0.000,0.000,22.356
48d57401250b4d7c,2,soft,3798,0.000,0.000,22.351
29bdebdc1e7268a2,4,soft,3798,0.000,0.000,22.337
497cd89f0feecac3,3,soft,3798,0.000,0.000,22.335
246f92446ac231c3,3,soft,3793,0.000,0.000,22.356
48d57401250b4d7c,3,soft,4318,0.000,0.000,22.351
29bdebdc1e7268a2,3,soft,4318,0.000,0.000,22.337
```

Here is an example of it working after these changes:
We get the node_position on the first sample for both nodes and we can see in the following lines that the node_positions stay 1 and 2 for the respective nodes.
Screenshot 2024-03-07 at 2 22 59 PM

Here we can see after I unplugged the nodes from the bridge that we detect a time difference of > 500ms and we update the node positions.
Screenshot 2024-03-07 at 2 22 13 PM

https://app.shortcut.com/sofar/story/199921/node-position-sticks-between-sensors

### DIFF
--- a/src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
@@ -53,22 +53,15 @@ void AanderaaSensor::aanderaSubCallback(uint64_t node_id, const char *topic, uin
         uint64_t sensor_reading_time_sec = d.header.sensor_reading_time_ms / 1000U;
         uint32_t sensor_reading_time_millis = d.header.sensor_reading_time_ms % 1000U;
 
-        // Keep track of when we last got a reading from this node
-        // If it is the first reading, or if it has been more than the current aggregation period + 30 seconds
-        // then we will update the node position.
-        static int8_t node_position = 0;
-        static uint32_t last_timestamp = 0;
-        static uint32_t current_timestamp = 0;
-
-        current_timestamp = pdTICKS_TO_MS(xTaskGetTickCount());
-        if ((current_timestamp - last_timestamp > aanderaa->current_agg_period_ms + 30000u) ||
+        uint32_t current_timestamp = pdTICKS_TO_MS(xTaskGetTickCount());
+        if ((current_timestamp - aanderaa->last_timestamp > aanderaa->current_agg_period_ms + 30000u) ||
             aanderaa->reading_count == 1U) {
           printf("Updating aanderaa %" PRIx64 " node position, current_time = %" PRIu32
                  ", last_time = %" PRIu32 ", reading count: %" PRIu32 "\n",
-                 node_id, current_timestamp, last_timestamp, aanderaa->reading_count);
-          node_position = topology_sampler_get_node_position(node_id, pdTICKS_TO_MS(5000));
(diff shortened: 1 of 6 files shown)


src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
```diff
@@ -53,22 +56,15 @@ void AanderaaSensor::aanderaSubCallback(uint64_t node_id, const char *topic, uin
...
-        // If it is the first reading, or if it has been more than the current aggregation period + 30 seconds
-        // then we will update the node position.
-        static int8_t node_position = 0;
-        static uint32_t last_timestamp = 0;
-        static uint32_t current_timestamp = 0;
-
-        current_timestamp = pdTICKS_TO_MS(xTaskGetTickCount());
-        if ((current_timestamp - last_timestamp > aanderaa->current_agg_period_ms + 30000u) ||
+        uint32_t current_timestamp = pdTICKS_TO_MS(xTaskGetTickCount());
+        if ((current_timestamp - aanderaa->last_timestamp > DEFAULT_AANDERAA_READING_PERIOD_MS + 30000u) ||
```
victorsowa12, line 60: Another fix is now we are using the correct period for comparison. Previously we were using `aanderaa->current_agg_period_ms` but this is equal to the `sampleDuration` not the individual reading period. Since this code runs on the bridge I created a #define for the comparison, but sometime in the future we should get this value from the mote itself.

towynlin (approved):

PR #108, opened 2024-03-11
Refactor/remove sensor controller timer

victorsowa12:
I removed the `sensorController`'s 30s timer and moved the task notification to `topology_sampler`. Now we will sample for sensors 5 seconds after BUS ON and this will decrease the latency between start-up time and the time SOFT/RBR data makes it into SENS_IND.log on Spotter.

here is an example where on boot we can see that we sampled for sensors right after the first topology sample (where we assembled/printed the network cbor map every time on boot!):
Screenshot 2024-03-11 at 2 35 30 PM

### DIFF
--- a/src/apps/bridge/bridgePowerController.cpp
+++ b/src/apps/bridge/bridgePowerController.cpp
@@ -226,7 +226,7 @@ void BridgePowerController::_update(void) {
         if (nextSampleEpochS > currentCycleS) {
           powerBusAndSetSignal(false);
         }
-        xTaskNotify(sensor_controller_task_handle, AANDERAA_AGGREGATION_TIMER_BITS, eSetBits);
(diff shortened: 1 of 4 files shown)


src/apps/bridge/bridgePowerController.cpp
```diff
@@ -226,7 +226,7 @@ void BridgePowerController::_update(void) {
         if (nextSampleEpochS > currentCycleS) {
           powerBusAndSetSignal(false);
         }
-        xTaskNotify(sensor_controller_task_handle, AANDERAA_AGGREGATION_TIMER_BITS, eSetBits);
+        xTaskNotify(sensor_controller_task_handle, AGGREGATION_TIMER_BITS, eSetBits);
```
victorsowa12, line 229: this is no longer aanderaa specific, so I removed the `AANDERAA`

russelldeguzman (approved):
👍

PR #111, opened 2024-03-14
Sc 200071/print mods and timer

russelldeguzman:
This PR cleans up a few prints:

- Print to log when we fail to probe the Type of the RBR 
- Clean up "recovered" -> "online"
- Only print ^ when we've exceeded the debounce count. 

and also adds comment clarification about timer choices

### DIFF
--- a/src/apps/bristleback_apps/bm_rbr/user_code/rbr_sensor.cpp
+++ b/src/apps/bristleback_apps/bm_rbr/user_code/rbr_sensor.cpp
@@ -95,7 +95,15 @@ bool RbrSensor::probeType(uint32_t timeout_ms) {
           _type = BmRbrDataMsg::SensorType::TEMPERATURE;
           success = true;
           break;
+        } else {
+          bm_fprintf(0, RBR_RAW_LOG, "Invalid outputformat: %s\n", _payload_buffer);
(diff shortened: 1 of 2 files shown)


src/apps/bristleback_apps/bm_rbr/user_code/rbr_sensor.cpp
```diff
@@ -108,10 +116,10 @@ bool RbrSensor::probeType(uint32_t timeout_ms) {
...
     }
   } else {
-    if (_sensorDropDebounceCount) {
-      bm_fprintf(0, RBR_RAW_LOG, "RBR sensor was recovered\n");
-      bm_printf(0, "RBR sensor was recovered\n");
-      printf("RBR sensor was recovered\n");
+    if (_sensorDropDebounceCount >= SENSOR_DROP_DEBOUNCE_MAX_COUNT) {
+      bm_fprintf(0, RBR_RAW_LOG, "RBR sensor online\n");
+      bm_printf(0, "RBR sensor online\n");
+      printf("RBR sensor online\n");
```
russelldeguzman, line 122: "recovered" kind of implied the mote took action to put the sensor back online. 

We have a different FTL print for this and the online detection should be kept distinct.

towynlin (approved):

PR #112, opened 2024-03-15
Sc 200214/bridge write default configs

russelldeguzman:
```
[[8046209027809048845, "bridge-dbg", 2140659395, 1707199808, {"sampleIntervalMs": 1200000, "sampleDurationMs": 300000, "subsampleIntervalMs": 60000, "subsampleDurationMs": 30000, "subsampleEnabled": 0, "bridgePowerControllerEnabled": 0, "alignmentInterval5Min": 1, "samplesPerReport": 2, "transmitAggregations": 1, "currentReadingPeriodMs": 60000, "softReadingPeriodMs": 500, "rbrCodaReadingPeriodMs": 500}]]
```

### DIFF
--- a/src/apps/bridge/app_config.cpp
+++ b/src/apps/bridge/app_config.cpp
@@ -1,37 +1,100 @@
 #include "app_config.h"
+#include "bridgeLog.h"
 #include "bridgePowerController.h"
 
 power_config_s getPowerConfigs(cfg::Configuration &syscfg) {
   power_config_s pwrcfg;
 
+  bool save_config = false;
   pwrcfg.sampleIntervalMs = BridgePowerController::DEFAULT_SAMPLE_INTERVAL_S * 1000;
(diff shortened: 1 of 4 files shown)


victorsowa12 (approved):

towynlin:
I think it's probably fine, but can you imagine any issue with not rebooting after writing these configs?

russelldeguzman:
@towynlin This is pretty safe since we're doing all of this in the init functions before we actually use the configs. I could only imagine a problem if we try to save a new config on the fly while the config is trying to be used. 

If we want to reboot - we may want to consolidate where we read the configs out of NVM. Otherwise, as it's written, we'd reboot three times for a blank config

towynlin:
Thanks. The triple reboot is not a reasonable tradeoff in the manufacturing process for handling an edge case that we think may never happen. Love it. :shipit:

towynlin (approved):

PR #116, opened 2024-03-19
print to spotter if we can't talk to power sensors

russelldeguzman:
Adding a bit of observability up to spotter (printf) when a module can't talk to the power sensor(s).


towynlin (commented):
Good call — and while we're at it, let's add a `bm_fprintf` to power.log as well for later analysis.

towynlin (approved):

PR #121, opened 2024-03-26
Sc 200592/debug loggin

russelldeguzman:
Gate raw long printing behind `sensorBmLogEnable`

I'd love to consolidate this to a single file especially for things we do across all modules, but I think that should be done in a separate ticket effort how we clean up/design configurations going forward.


towynlin (approved):
Agreed and agreed

PR #122, opened 2024-03-26
Always clear the reportBuilder list

victorsowa12:
This is a fix so that we always clear the reportBuilder linked list, even if we have `transmitAggregations = 0`.

https://app.shortcut.com/sofar/story/200593/soft-bow-bugs


russelldeguzman (approved):
🎉

towynlin (approved):
🙌

PR #127, opened 2024-04-04
Bridge reset memory leak fix

victorsowa12:
In the rare case that we receive an RX message after the adin thread has been paused, we need to make sure that we do not free the `event.data` and instead re-submit the RX buffer, just like if we were processing a normal RX message when the adin is on.


src/lib/drivers/adin2111/src/eth_adin2111.cpp
```diff
@@ -295,6 +295,18 @@ static void adin2111_thread(void *parameters) {
...
+                case EVT_ETH_RX:{
+                    // Sometimes we get a RX event when we're paused. In order to make
+                    // sure we don't leak memory, lets resubmit the buffer just like
+                    // we would do if the adin was not paused.
+                    rxMsgEvt_t *rxMsg = static_cast<rxMsgEvt_t *>(event.data);
+                    // Re-submit buffer into ADIN's RX queue
+                    adi_eth_Result_e result = adin2111_SubmitRxBuffer(rxMsg->dev, &rxMsg->bufDesc);
+                    if (result != ADI_ETH_SUCCESS) {
+                        printf("Unable to re-submit RX Buffer\n");
+                    }
```
towynlin: Is there anything else we should do here?

I see 3 existing uses of `adin2111_SubmitRxBuffer`:
- one just prints like we're doing here
- ~one retries up to 8 times~ (sorry I misread "entries" as "retries")
- one asserts false, causing a reboot.

I'm trying to think through what would happen if we failed to submit but kept running... maybe that buffer is now just lost, and we're operating with one fewer than before? That's not catastrophic but also not ideal. Retrying and/or asserting false seem like better ideas to me. Thoughts?

towynlin (approved):

src/lib/drivers/adin2111/src/eth_adin2111.cpp
russelldeguzman: +1 to Asserting.
Our app code is under the operating assumption that there are `RX_QUEUE_NUM_ENTRIES (8)` slots available. If we're failing to submit at this point we're probably doing something critically wrong and an assert would be a clean place to gather a coredump before memory gets potentially more corrupted down the road.

russelldeguzman (approved):

src/lib/drivers/adin2111/src/eth_adin2111.cpp
victorsowa12: Good points I'll add the assert! one thing also to consider is that we are re-submitting the entire buffer array in the `adin2111_power_cb` so I'm not sure what would happen if we re-submit the same buffer twice. I'll look into that.

victorsowa12: I can also add the assert to the other locations where we can potentially fail to submit the buffer

russelldeguzman: (and also adapt the others error handles to this function to assert as well)

russelldeguzman: >re-submitting the entire buffer array 

the power togging calls `adin2111_Init` which eventually calls `Mac_Init`
which calls `queueInit` which sets all of the `pBufDesc` to null and resets the queue state

PR #128, opened 2024-04-09
Sc 201297/bridge power controller ticks

russelldeguzman:
Main changes:
* Adding `kTicksSamplingEnabled` configuration for ticks sampling
* Use ticks as the timebase instead of RTC if config is enabled
* Rename a lot of variables
* Unit tests

This approach rolls over every ~50 days for sampling (which causes either an extra long period or extra short period). We could write/re-write the ticks implementation to better handle rolls - but this is a larger effort and probably should be separate from this PR. 

Soak logs

[0000_BRIDGE_CFG.log](https://github.com/bristlemouth/bm_protocol/files/14923297/0000_BRIDGE_CFG.log)
[0000_BRIDGE_SYS.log](https://github.com/bristlemouth/bm_protocol/files/14923298/0000_BRIDGE_SYS.log)

### DIFF
--- a/src/apps/bridge/app_config.h
+++ b/src/apps/bridge/app_config.h
@@ -19,7 +19,7 @@ constexpr const char *SAMPLES_PER_REPORT = "samplesPerReport";
 constexpr const char *CURRENT_READING_PERIOD_MS = "currentReadingPeriodMs";
 constexpr const char *SOFT_READING_PERIOD_MS = "softReadingPeriodMs";
 constexpr const char *RBR_CODA_READING_PERIOD_MS = "rbrCodaReadingPeriodMs";
-constexpr const char *kTicksSamplingEnabled = "ticksSamplingEnabled";
+constexpr const char *TICKS_SAMPLING_ENABLED = "ticksSamplingEnabled";
 
 } // namespace AppConfig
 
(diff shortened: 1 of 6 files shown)


src/apps/bridge/app_config.h
```diff
@@ -19,12 +19,14 @@ constexpr const char *SAMPLES_PER_REPORT = "samplesPerReport";
 constexpr const char *CURRENT_READING_PERIOD_MS = "currentReadingPeriodMs";
 constexpr const char *SOFT_READING_PERIOD_MS = "softReadingPeriodMs";
 constexpr const char *RBR_CODA_READING_PERIOD_MS = "rbrCodaReadingPeriodMs";
+constexpr const char *kTicksSamplingEnabled = "ticksSamplingEnabled";
```
towynlin: Haha, ok, so consistency is the most important thing here. Per our recent conversation we could either do constants `LIKE_THIS` or `kLikeThis`. You and I both searched and found no super compelling reason to do one over the other.

I have a slight preference to avoid SHOUTING, so I would support this change to the `k` prefix. However, consistency is queen — if we change the style of this one, we need to change at least all the constants in this file to the new style.

russelldeguzman: I'll change to use to use CAPS for the moment - if we decide to do a cleanliness thing later we can change it then.

towynlin (approved):

victorsowa12 (approved):

PR #130, opened 2024-04-16
fix: subsampling

towynlin:
[sc-201530]

This fixes a bug in the bridge power controller's subsampling behavior, often used with RBR sensors. Prior to this change, if subsampling was enabled, and if the end of the sample duration fell between the end of the last subsample duration and the start of the next aligned sample interval, then the start of the next sample interval would actually again leave the bus off to align to the _next_ sample interval.

I added some unit tests involving a couple different cases of subsampling.

The fix is essentially to avoid sleeping across that end-of-sample-duration boundary, allowing the power controller to wake up and align the next sample before it starts.

### DIFF
--- a/src/apps/bridge/bridgePowerController.cpp
+++ b/src/apps/bridge/bridgePowerController.cpp
@@ -197,13 +197,15 @@ void BridgePowerController::_update(void) {
             time_to_sleep_ms = MAX(subsampleTimeRemainingS * 1000, MIN_TASK_SLEEP_MS);
             break;
           } else {
-            uint32_t nextSubsampleEpochS = _subsampleIntervalStartS + _subsampleIntervalS;
+            const uint32_t nextSubsampleEpochS = _subsampleIntervalStartS + _subsampleIntervalS;
             _subsampleIntervalStartS = nextSubsampleEpochS;
-            stateLogPrintTarget("Subsampling Off", nextSubsampleEpochS);
-            time_to_sleep_ms =
-                (currentCycleS < nextSubsampleEpochS)
-                    ? MAX((nextSubsampleEpochS - currentCycleS) * 1000, MIN_TASK_SLEEP_MS)
-                    : MIN_TASK_SLEEP_MS;
+            const uint32_t secondsUntilNextSubsample =
+                nextSubsampleEpochS > currentCycleS ? nextSubsampleEpochS - currentCycleS : 0;
+            // Check whether this is the last subsample within a sample duration
+            // If so, only sleep until sampling off, to align the next sample
+            const uint32_t timeToSleepS = MIN(sampleTimeRemainingS, secondsUntilNextSubsample);
+            stateLogPrintTarget("Subsampling Off", currentCycleS + timeToSleepS);
+            time_to_sleep_ms = MAX(timeToSleepS * 1000, MIN_TASK_SLEEP_MS);
             // Prevent bus thrash
             if (nextSubsampleEpochS > currentCycleS) {
               powerBusAndSetSignal(false);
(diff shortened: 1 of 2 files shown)


src/apps/bridge/bridgePowerController.cpp
```diff
@@ -197,13 +197,15 @@ void BridgePowerController::_update(void) {
...
-            stateLogPrintTarget("Subsampling Off", nextSubsampleEpochS);
-            time_to_sleep_ms =
-                (currentCycleS < nextSubsampleEpochS)
-                    ? MAX((nextSubsampleEpochS - currentCycleS) * 1000, MIN_TASK_SLEEP_MS)
-                    : MIN_TASK_SLEEP_MS;
+            const uint32_t secondsUntilNextSubsample =
+                nextSubsampleEpochS > currentCycleS ? nextSubsampleEpochS - currentCycleS : 0;
+            // Check whether this is the last subsample within a sample duration
+            // If so, only sleep until sampling off, to align the next sample
+            const uint32_t timeToSleepS = MIN(sampleTimeRemainingS, secondsUntilNextSubsample);
```
towynlin, line 206: The crux of the fix

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -351,8 +353,8 @@ uint32_t BridgePowerController::_alignNextInterval(uint32_t nowEpochS,
       alignedEpoch += adjustment;
       bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
                      "Aligning next sample interval to %s by delaying an additional %" PRIu32
-                     " seconds to %" PRIu32 "\n", (_ticksSamplingEnabled) ? "uptime" : "UTC",
-                     adjustment, alignedEpoch);
+                     " seconds to %" PRIu32 "\n",
+                     (_ticksSamplingEnabled) ? "uptime" : "UTC", adjustment, alignedEpoch);
```
towynlin, line 357: This is just clang-format adjusting whitespace.

test/src/common/bridgePowerController_ut.cpp
```diff
@@ -147,6 +147,287 @@ TEST_F(BridgePowerControllerTest, alignment) {
...
 
+// Sample duration ends at 19 minutes, which is after subsample duration ends at 16 minutes,
+// and before the next aligned sample interval starts at 20 minutes.
+TEST_F(BridgePowerControllerTest, subsampling1) {
+  const uint32_t kTwentyMinutes = 1200000;
+  const uint32_t kNineteenMinutes = 1140000;
+  const uint32_t kFiveMinutes = 300000;
+  const uint32_t kOneMinute = 60000;
+  BridgePowerController powerController(FAKE_VBUS_EN, kTwentyMinutes, kNineteenMinutes,
+                                        kFiveMinutes, kOneMinute, true, true);
```
towynlin, line 158: This test case is exactly the configuration that exposed the bug in our RBR testing.

test/src/common/bridgePowerController_ut.cpp
```diff
@@ -147,6 +147,287 @@ TEST_F(BridgePowerControllerTest, alignment) {
...
+
+// Subsample duration end lines up with sample duration end at 7 minutes.
+// Subsample interval is 3 minutes, which is not 5-minute aligned.
+TEST_F(BridgePowerControllerTest, subsampling2) {
+  const uint32_t kTenMinutes = 600000;
+  const uint32_t kSevenMinutes = 420000;
+  const uint32_t kThreeMinutes = 180000;
+  const uint32_t kOneMinute = 60000;
+  BridgePowerController powerController(FAKE_VBUS_EN, kTenMinutes, kSevenMinutes, kThreeMinutes,
+                                        kOneMinute, true, true);
```
towynlin, line 313: I added this test case to ensure I didn't break the existing behavior, and to prove to myself that subsampling is unaffected by UTC alignment.

russelldeguzman (approved):

victorsowa12 (approved):

src/apps/bridge/bridgePowerController.cpp
evanShap: @towynlin one thing I don't understand - why did we only see the glitch every 10 hours or so?

towynlin: Agreed that it's odd. The only explanation I can think of is that most of the time the power controller task must wake up slightly early with respect to the RTC goal. In that case the extra 20-minute gap bug doesn't manifest. If there were no clock drift and sleep durations were perfectly precise then there would be a 20-minute gap alternating with a 20-minute sampling period.

PR #134, opened 2024-04-24
Sc 201662/bridge turbidity support

victorsowa12:
Add the turbidity sensor to the bridges known sensors. Now the bridge can add/send turbidity data to the spotter in its reports.


src/apps/bridge/sensorController.cpp
```diff
@@ -242,19 +263,36 @@ static bool node_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
       } else if (strncmp(reply.app_name, "bm_rbr",
                          MIN(reply.app_name_strlen, strlen("bm_rbr"))) == 0) {
         if (!sensorControllerFindSensorById(reply.node_id)) {
-          uint32_t rbr_coda_agg_period_ms =
+          uint32_t rbr_coda_reading_period_ms =
```
victorsowa12, line 266: small re-name to more properly reflect what the variable means

src/apps/bridge/sensorController.cpp
```diff
@@ -29,7 +31,8 @@ typedef struct sensorControllerCtx {
   cfg::Configuration *_sys_cfg;
   uint32_t current_reading_period_ms;
   uint32_t soft_reading_period_ms;
-  uint32_t rbr_coda_agg_period_ms;
+  uint32_t rbr_coda_reading_period_ms;
```
victorsowa12, line 34: renamed to more appropriately reflect what the variable is

src/apps/bridge/sensor_drivers/seapointTurbiditySensor.cpp
```diff
@@ -0,0 +1,175 @@
...
+  configASSERT(log_buf);
+  if (xSemaphoreTake(_mutex, portMAX_DELAY)) {
+    size_t log_buflen = 0;
+    seapoint_turbidity_aggregations_t turbidity_aggs = {
+        .turbidity_s_mean_ftu = NAN, .turbidity_r_mean_ftu = NAN, .reading_count = 0};
+    if (turbidity_s_ftu.getNumSamples() > MIN_READINGS_FOR_AGGREGATION) {
+      turbidity_aggs.turbidity_s_mean_ftu = turbidity_s_ftu.getMean();
+      turbidity_aggs.turbidity_r_mean_ftu = turbidity_r_ftu.getMean();
+      turbidity_aggs.reading_count = reading_count;
+    }
```
victorsowa12, line 116: Normally we do some sort of sanity check on here for min/max values, but I talked to Evan about this and we agreed to just let Spotters bitpacking do the filtering based on the encodable min/max values.

src/apps/bridge/sensor_drivers/seapointTurbiditySensor.cpp
```diff
@@ -0,0 +1,175 @@
...
+                     "%016" PRIx64 ","     // Node Id
+                     "%" PRIi8 ","         // node_position
+                     "seapoint_turbidity," // node_app_name
+                     "%" PRIu64 ","        // reading_uptime_millis
+                     "%" PRIu64 "."        // reading_time_utc_ms seconds part
+                     "%03" PRIu32 ","      // reading_time_utc_ms millis part
+                     "%" PRIu64 "."        // sensor_reading_time_ms seconds part
+                     "%03" PRIu32 ","      // sensor_reading_time_ms millis part
+                     "%.4f,"               // s_signal
+                     "%.3f\n",             // r_signal
```
towynlin, line 85: Just checking/learning — is this the precision that actually comes from the sensor?

towynlin (approved):
👏 Had one small question.

src/apps/bridge/sensor_drivers/seapointTurbiditySensor.cpp
victorsowa12: For the s_signal I believe it has 4 decimal places. The r_signal may only have 1 (based on my testing).
Here are some example raw lines from the sensor (I got different values by moving my hand over it):
```
line: 0.0030,990.3
line: 4.6124,1084.5
line: 50.102,1290.8
line: 103.88,1413.2
line: 94.568,1398.3
line: 77.588,1359.9
line: 15.611,1163.3
line: 0.0054,992.8
```
The s_signal definitely goes down to 4 decimals at really close to 0 FTU. The R signal only ever seems to have 1. I cannot find in the data sheet an explicit number of decimals for either signal. 

There is this tid-bit in the data sheet:
> • Noise Floor: < 0.005 FTU RMS

It apparently applies to both s and r signals. I'm not sure if that means the the output will ever have 3 decimal places or not. I put three for the r_signal just in case for now and four for the s_signal since it is outputting that many.

towynlin: Excellent. Thanks for the quick research!

russelldeguzman (approved):

PR #135, opened 2024-04-24
fix: make bridge power controller more resilient to clock drift

towynlin:


test/src/common/bridgePowerController_ut.cpp
```diff
@@ -466,12 +633,19 @@ TEST_F(BridgePowerControllerTest, goldenPath) {
...
+  EXPECT_EQ(xTaskGetTickCount(), curtime);
+
   // The bus stays on until the next Sample Off time
-  uint32_t curtime =
-      (BridgePowerController::DEFAULT_SAMPLE_INTERVAL_S + SAMPLE_DURATION_S) * 1000;
+  rtcGetMicroSeconds_fake.return_val = curtime * 1000;
+  BridgePowerController._update();
+  EXPECT_EQ(fake_io_read_func_fake.call_count, 5);
+  EXPECT_EQ(fake_io_write_func_fake.call_count, 1);
+  curtime += SAMPLE_DURATION_S * 1000;
```
towynlin, line 648: The golden path test had to be slightly modified because it was written to rely on the behavior of sleeping past a sample start, which is the bug we're fixing. Other expectations remain the same, but there's an extra wake-up and so the test contains an extra call to `_update()`. I split the `curtime` expectations into 2 components, first up to the sample interval, then to the end of the sample duration.

test/src/common/bridgePowerController_ut.cpp
```diff
@@ -428,6 +428,173 @@ TEST_F(BridgePowerControllerTest, subsampling2) {
...
+
+  // Replicate the bug: wake early
+  M = 18;
+  S = 59;
+  rtcGetMicroSeconds_fake.return_val = 1713233939000000;
+  powerController._update();
+  EXPECT_EQ(fake_io_read_func_fake.call_count, 12);
+  EXPECT_EQ(fake_io_write_func_fake.call_count, 10);
+  curtime += 1000;
+  EXPECT_EQ(xTaskGetTickCount(), curtime);
```
towynlin, line 565: Wake a second early. Expect to go back to sleep for a second without touching the bus power. This was where the test failed before I modified the bridge power controller.

russelldeguzman (approved):

victorsowa12 (approved):

PR #137, opened 2024-05-14
diff signal lib

russelldeguzman:
A library for computing a difference signal of N samples.

### DIFF
--- a/src/lib/common/differenceSignal.cpp
+++ b/src/lib/common/differenceSignal.cpp
@@ -1,0 +1,43 @@
+#include "differenceSignal.h"
+#include "FreeRTOS.h"
+
+DifferenceSignal::DifferenceSignal(uint32_t numTotalSamples) : r_i(0) {
+  configASSERT(numTotalSamples > 0);
+  r_n = numTotalSamples;
+  r = static_cast<double *>(pvPortMalloc((r_n) * sizeof(double)));
+  configASSERT(r);
+}
+
+bool DifferenceSignal::addSample(double sample) {
+  if (r_i < r_n) {
+    r[r_i++] = sample;
+    return true;
+  }
+  return false;
+}
+
+bool DifferenceSignal::encodeDifferenceSignalToBuffer(double *d_n, size_t &numSamples) {
+  configASSERT(d_n);
+  configASSERT(numSamples > 0);
+  bool rval = false;
+  do {
+    if (numSamples < (r_i)) {
+      break;
+    }
+    if (!r_i) {
+      break;
+    }
+    d_n[0] = 0.0;
+    for (uint32_t i = 1; i < r_i; i++) {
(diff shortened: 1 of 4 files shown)


src/lib/common/differenceSignal.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "FreeRTOS.h"
+
+DifferenceSignal::DifferenceSignal(uint32_t numTotalSamples) : r_i(0) {
+  configASSERT(numTotalSamples > 0);
+  r_n = numTotalSamples;
+  r = static_cast<double *>(pvPortMalloc((r_n) * sizeof(double)));
+  configASSERT(r);
+}
+
+bool DifferenceSignal::addSample(double sample) {
```
russelldeguzman, line 11: We could make this a circular buffer later too, chose to punt on it because we probably want a `circularBuffer` abstract class, which should be it's own PR

src/lib/common/differenceSignal.cpp
```diff
@@ -0,0 +1,45 @@
...
+    numSamples = r_i;
+    rval = true;
+  } while (0);
+  if (!rval) {
+    numSamples = 0;
+  }
+  return rval;
+}
+
+void DifferenceSignal::clear() { r_i = 0; }
```
victorsowa12, line 43: should we clear the buffer incase the next set of samples is smaller than the previous set?

victorsowa12 (approved):
Overall looks good to me!

src/lib/common/differenceSignal.cpp
russelldeguzman: We shouldn't need to since we only process up to `r_i` samples.

victorsowa12: Ahh yes! thanks! 👍

towynlin (commented):
Looks good! Thanks. Sorry I'm just getting to this.

evanShap:
@russelldeguzman This looks right to me, but we'll need to make a few additions (or can handle one of them in higher scope):
1. When encoding to buffer, we should preserve or return the "residual" to the caller - which is simply `r[0]` in your code. This is required to reconstruct the full decoded info. We may or may not preserve the residuals of our encoding depending on what Isabel comes up with.

2. Technically the difference signal should simply be shorter than the input series by 1 - rather than setting the first element to 0. Implementing it that way may minimize confusion - especially if we end up going more than 1 order up in encoding.

3. We'll need to enforce the 500ms reading gap check when adding to the buffer, but can do that outside the scope of this class?

russelldeguzman:
> We'll need to enforce the 500ms reading gap check when adding to the buffer, but can do that outside the scope of this class? 

@evanShap We can do this in the controller outside of this class. 
I'll make a follow-up PR to address your first two bullets!

PR #138, opened 2024-05-22
Bugfix/fixing difference signal

russelldeguzman:
Fixing No 1. and 2. Here:
https://github.com/bristlemouth/bm_protocol/pull/137#issuecomment-2123577873

### DIFF
--- a/src/lib/common/differenceSignal.cpp
+++ b/src/lib/common/differenceSignal.cpp
@@ -43,3 +43,14 @@ bool DifferenceSignal::encodeDifferenceSignalToBuffer(double *d_n, size_t &numSa
 void DifferenceSignal::clear() { r_i = 0; }
 
 bool DifferenceSignal::isFull() { return r_i == r_n; }
+
+double DifferenceSignal::mean() {
+    if(!r_i) {
+        return 0.0;
+    }
+    double sum = 0.0;
+    for (uint32_t i = 0; i < r_i; i++) {
+        sum += r[i];
+    }
+    return sum / r_i;
+}
--- a/src/lib/common/differenceSignal.cpp
+++ b/src/lib/common/differenceSignal.cpp
@@ -44,7 +44,7 @@ void DifferenceSignal::clear() { r_i = 0; }
 
 bool DifferenceSignal::isFull() { return r_i == r_n; }
 
-double DifferenceSignal::mean() {
+double DifferenceSignal::signalMean() {
     if(!r_i) {
         return 0.0;
     }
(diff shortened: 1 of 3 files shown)


towynlin (commented):
Should probably add a unit test for `getReferenceSignal`.

I was expecting to see the reference signal encoded into the buffer, but I now think you're planning to encode it in a bitpacked message separately, outside this lib. Is that correct?

russelldeguzman:
@towynlin See 
`test/src/common/differenceSignal_ut.cpp:52,69`!

>  I now think you're planning to encode it in a bitpacked message separately, outside this lib. Is that correct

Yes - That's correct!

src/lib/common/differenceSignal.cpp
```diff
@@ -16,6 +25,12 @@ bool DifferenceSignal::addSample(double sample) {
   return false;
 }
 
+/*!
+ * @brief Encodes the difference signal to a buffer
+ * @param d_n[out] The buffer to store the difference signal
+ * @param numSamples[in, out] In: The buffer sample size, Out: The number of samples encoded into the buffer
+ * @return True if the difference signal was encoded, false otherwise
+ */
 bool DifferenceSignal::encodeDifferenceSignalToBuffer(double *d_n, size_t &numSamples) {
```
evanShap, line 35: @russelldeguzman it looks like this DifferenceSignal class is responsible for allocating the buffer for storing the raw data, but the caller is responsible for allocating the buffer to store the output difference signal and passing it in here, yea?

src/lib/common/differenceSignal.cpp
```diff
@@ -40,6 +54,41 @@ bool DifferenceSignal::encodeDifferenceSignalToBuffer(double *d_n, size_t &numSa
...
+    }
+    return sum / r_i;
+}
+
+/*!
+ * @brief Gets the reference signal
+ * @param r0[out] The reference signal
+ * @return True if the reference signal was retrieved, false otherwise
+ */
+bool DifferenceSignal::getReferenceSignal(double &r0) {
```
evanShap, line 87: @russelldeguzman this is the residual required for decoding, right?

evanShap:
> Should probably add a unit test for `getReferenceSignal`.
> 
> I was expecting to see the reference signal encoded into the buffer, but I now think you're planning to encode it in a bitpacked message separately, outside this lib. Is that correct?

Isabel is working on the data analysis, but right now the thinking is we're probably going to ditch the residuals - we don't really need them.

src/lib/common/differenceSignal.cpp
russelldeguzman: Yes thats right!

src/lib/common/differenceSignal.cpp
russelldeguzman: Yes that's correct!

evanShap: It also looks like it's OK to pass in any sized buffer we want - so if we only want 300 samples for our output we can pass in 300 even if there are 1500 samples in the raw buffer. We can pass in 2000 samples, and it will make sure to not run off the end of the raw buffer.
One thing that the caller is responsible for is making sure they have actually allocated enough memory in `d` per how many samples we request.

towynlin (approved):

PR #140, opened 2024-06-04
function to get 2nd order difference

russelldeguzman:
Adding a static function hook to get a difference signal from a double buffer.

### DIFF
--- a/src/lib/common/differenceSignal.cpp
+++ b/src/lib/common/differenceSignal.cpp
@@ -92,6 +92,13 @@ bool DifferenceSignal::getReferenceSignal(double &r0) {
   return true;
 }
 
+/*!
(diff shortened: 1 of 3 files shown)


towynlin (approved):
Looks good. Would you want to have `encodeDifferenceSignalToBuffer` use `differenceSignalFromBuffer`? It's only a very small code duplication, so either way seems fine to me. Just curious whether you'd considered that.

russelldeguzman:
@towynlin that's a good suggestion - It's definitely possible, the caveat is that `encodeDifferenceSignalToBuffer` isn't designed to be in place, so a call to differenceSignalFromBuffer, would modify the reference signal in an unintended way. My inclination is to leave it for now, and optimize later.

PR #141, opened 2024-06-11
Serial IRQ fix + adding debug features

victorsowa12:
Here is the serial IRQ fix.

I have also left in some debugging features, but commented them out.

There are also a few very rough scripts that I wrote for processing outputs and debugging.

All of the testing I did for validating the fix is here:
https://app.shortcut.com/sofar/story/201546/rbr-003-freezer-resets-add-fw-diagnostics#activity-203216


src/lib/common/serial.c
```diff
@@ -261,7 +266,6 @@ void serialGenericUartIRQHandler(SerialHandle_t *handle) {
 
   // Let the RTOS know if a task needs to be woken up
   portYIELD_FROM_ISR(higherPriorityTaskWoken);
-  // TODO - call this from actual irqhandler?
```
victorsowa12, line 264: We do not need to do this since FreeRTOS will not context switch until the ISR is fully completed.

tools/scripts/misc/bm_concat_logs.py
```diff
@@ -0,0 +1,25 @@
+import sys
+import datetime
```
victorsowa12, line 2: These next few files are random scripts I created while testing rtc-0, rbr resets, soft resets, capturing usb output. They are not polished and I can delete them from this PR. But I thought others/myself might be able to quickly modify, and use these again in the future and having them committed prevented "dirty" builds 😅.

src/bsp/common/stm32/startup_stm32u575zitxq.c
```diff
@@ -184,6 +184,139 @@ void MDF1_FLT5_IRQHandler         (void) __attribute__ ((weak, alias("Default_Ha
 void CORDIC_IRQHandler            (void) __attribute__ ((weak, alias("Default_Handler")));
 void FMAC_IRQHandler              (void) __attribute__ ((weak, alias("Default_Handler")));
 
+
```
victorsowa12, line 187: I used this heavily for the RBR resets debugging. It doesn't do anything unless you enable tracing and uncomment one of the lines/follow the steps outlined here.

src/apps/bristleback_apps/bristleback_apps_common/task_priorities.h
```diff
@@ -28,8 +28,8 @@
 
 #define SERIAL_TX_TASK_PRIORITY 2
 #define CONSOLE_RX_TASK_PRIORITY 2
+#define USER_TASK_PRIORITY 2
```
victorsowa12, line 31: Since this is used for the LPUART(user's) RX task priority, I thought it would make more sense that both serial tasks had the same priority.

src/apps/bristleback_apps/bristleback_apps_common/memfault_platform_config.h
```diff
@@ -18,4 +18,4 @@
 // Send memfault metric updates every six hours
 #define MEMFAULT_METRICS_HEARTBEAT_INTERVAL_SECS (6 * 60 * 60)
 
-#define MEMFAULT_WATCHDOG_SW_TIMEOUT_SECS 30
+#define MEMFAULT_WATCHDOG_SW_TIMEOUT_SECS 4
```
victorsowa12, line 21: Not used, but if the software watchdog is ever enabled again in the future, I set it to 4 since our hardware watchdog is I believe 5 seconds.

src/apps/bristleback_apps/bristleback_apps_common/app_main.cpp
```diff
@@ -322,6 +322,10 @@ static void defaultTask(void *parameters) {
   // Inhibit low power mode during boot process
   lpmPeripheralActive(LPM_BOOT);
 
+  // If we ever include the SWWDG, we need to have the memfault
+  // platform boot before starting the watchdog task
+  memfault_platform_boot();
```
victorsowa12, line 327: as the comment states, I used the software watchdog for debugging this build and if it is ever enabled again in the future, this needs to be called before the watchdog task starts.

src/apps/bristleback_apps/bm_rbr/CMakeLists.txt
```diff
@@ -58,6 +58,7 @@ set(LIB_FILES
     ${SRC_DIR}/lib/common/serial.c
     ${SRC_DIR}/lib/common/serial_console_u5.cpp
     ${SRC_DIR}/lib/common/stress.c
+    # ${SRC_DIR}/lib/common/trace.c
```
victorsowa12, line 62: to have the ability to enable ISR tracing this file needs to be included. I left it commented out, but maybe we can have a cmake define that protects/enables this?

src/apps/bristleback_apps/bm_rbr/CMakeLists.txt
```diff
@@ -107,6 +108,7 @@ set(LIB_FILES
     ${SRC_DIR}/lib/drivers/w25.cpp
     ${SRC_DIR}/lib/bm_serial/bm_serial.c
     ${SRC_DIR}/lib/memfault/memfault_platform_core_u5.c
+    # ${SRC_DIR}/lib/memfault/memfault_lptim_software_watchdog.c
```
victorsowa12, line 113: I used a software watchdog for debugging and we could enable this, however, if we want it, it will need to be added to all of the builds since it would share the same task as the hardware watchdog. That should also be its own PR since it involves turning on the LPTIM2 for every bsp.

I left it commented out for now as an example, but if it makes it cleaner to remove I can do that.

src/apps/bristleback_apps/bm_rbr/CMakeLists.txt
```diff
@@ -280,6 +282,14 @@ set(APP_DEFINES
...
     CFG_TUSB_MCU=OPT_MCU_STM32U5
 
+    # Uncomment/change these to enable tracing!
+    # TRACE_ENABLE
+    # TRACE_BUFF_LEN=1024
+    # TRACE_TASKS
+    # TRACE_USE_COREDEBUG
+    # TRACE_ISR
+    # TRACE_SERIAL
+
```
victorsowa12: These are used to enable tracing and must be uncommented in addition to the trace.c file being included.

src/bsp/bm_mote_bristleback_v1_0/CMakeLists.txt
```diff
@@ -70,6 +71,8 @@ set(BSP_DEFINES
         STM32U575xx
         MEMFAULT_COREDUMP_STORAGE_WRITE_SIZE=512
         BSP_MOTE_BRISTLEBACK_V1_0
+        # MEMFAULT_SOFWARE_WATCHDOG_SOURCE=LPTIM2_BASE
+        # MEMFAULT_EXC_HANDLER_WATCHDOG=LPTIM2_IRQHandler
```
victorsowa12, line 77: used by the memfault software watchdog code.

src/bsp/bm_mote_bristleback_v1_0/CMakeLists.txt
```diff
@@ -53,6 +53,7 @@ set(BSP_FILES
         ${CMAKE_CURRENT_SOURCE_DIR}/Drivers/STM32U5xx_HAL_Driver/Src/stm32u5xx_ll_usart.c
         ${CMAKE_CURRENT_SOURCE_DIR}/Drivers/STM32U5xx_HAL_Driver/Src/stm32u5xx_ll_usb.c
         ${CMAKE_CURRENT_SOURCE_DIR}/Drivers/STM32U5xx_HAL_Driver/Src/stm32u5xx_ll_utils.c
+        # ${CMAKE_CURRENT_SOURCE_DIR}/Drivers/STM32U5xx_HAL_Driver/Src/stm32u5xx_hal_lptim.c
```
victorsowa12, line 57: After adding the lptim2 to the BSP this would need to be uncommented in order for the lptim to work.

src/lib/common/serial.c
```diff
@@ -237,19 +239,22 @@ void serialGenericUartIRQHandler(SerialHandle_t *handle) {
...
     if (bytesAvailable > 0) {
       // Transmit current byte
       usart_TransmitData8((USART_TypeDef *)handle->device, txByte);
-    } else   if(LL_USART_IsActiveFlag_TC((USART_TypeDef *)handle->device)) {
+    } else {
+      usart_DisableIT_TXE((USART_TypeDef *)handle->device);
+    }
+  }
+
+  if (!bytesAvailable && LL_USART_IsActiveFlag_TC((USART_TypeDef *)handle->device) && !usart_IsEnabledIT_TXE((USART_TypeDef *)handle->device)) {
```
victorsowa12: This is the heart of the fix.

Instead of waiting for the TC(transmission complete) interrupt while leaving the TXE(transmission buffer empty) interrupt enabled, we disable the TXE interrupt when there are 0 bytes left to send. We will then only interrupt when we get a TC flag.

Previously, because we didn't disable the TXE interrupt, the TXE flag would cause us to continually execute the interrupt. Luckily, disabling the TXE interrupt does not "clear" the flag, therefore the TC interrupt, which relies on the TXE flag being set still triggers an interrupt.

src/lib/common/serial.c
```diff
@@ -237,19 +239,22 @@ void serialGenericUartIRQHandler(SerialHandle_t *handle) {
...
       // Disable this interrupt if there are no more bytes to transmit
       LL_USART_ClearFlag_TC((USART_TypeDef *)handle->device);
-      usart_DisableIT_TXE((USART_TypeDef *)handle->device);
```
victorsowa12, line 248: already disabled, so we don't need to do that again.

src/lib/common/watchdog.c
```diff
@@ -5,10 +5,12 @@
 #include "task.h"
 #include "iwdg.h"
 #include "task_priorities.h"
+// #include "memfault/ports/watchdog.h"
```
victorsowa12, line 12: I left these commented out for now in case we ever want to include them again. I can remove them if that is cleaner.

src/apps/bristleback_apps/bm_rbr/CMakeLists.txt
russelldeguzman: Can you put a comment like "Enable for ISR timing debugging" next to it?

src/apps/bristleback_apps/bm_rbr/CMakeLists.txt
russelldeguzman: Can you add a comment pointing at `# ${SRC_DIR}/lib/common/trace.c`?

src/bsp/bm_mote_bristleback_v1_0/bm_mote_bristleback_v1_0.ioc
```diff
@@ -312,8 +315,9 @@ RCC.I2C1Freq_Value=160000000
...
+RCC.LPTIM2CLockSelectionVirtual=RCC_LPTIM2CLKSOURCE_LSI
+RCC.LPTIM2Freq_Value=32000
```
russelldeguzman: @victorsowa12 Can you describe what's happening in the IOC change?
Hard to grok this.

russelldeguzman (approved):
Looks great 🎉  A few comment requests and a question about the IOC but otherwise 👍

src/bsp/bm_mote_bristleback_v1_0/bm_mote_bristleback_v1_0.ioc
victorsowa12: Hmmm. I enabled a few new things in CubeMX, but then disabled them again, such as the LPTIM2 for the software watchdog. I think that CubeMX doesn't reset those values in the IOC file when you disable after enabling them and therefore it keeps them in the IOC. 

Since nothing in the HAL or LL coda has changed nothing in the code that will be run has changed.

These changes to the IOC should only impact what is displayed by CubeMX when you open it up again.

I agree that this is pretty unclear and I will give a quick test where I revert the IOC to the one on develop, since nothing should have changed.

src/lib/memfault/memfault_lptim_software_watchdog.c
```diff
@@ -4,7 +4,7 @@
 
 #include <stdbool.h>
 
-#include "stm32l4xx_hal.h"
+#include "stm32u5xx_hal.h"
```
victorsowa12, line 7: All of the changes in this file are to allow the use of the memfault software watchdog on the STM32U575 with LPTIM2.

src/bsp/bm_mote_bristleback_v1_0/bm_mote_bristleback_v1_0.ioc
victorsowa12: So I just copied the develop IOC into this branch and had it regenerate the code and nothing changed. So I will commit the develop IOC and that should get rid of all of these changes 🤞.

But regardless, i think it has something to do with the way CubeMX handles enabling - then disabling peripherals and saving the "last used" peripheral setting even though the peripheral is no longer enabled.

src/apps/bristleback_apps/bm_rbr/CMakeLists.txt
victorsowa12: Done

src/apps/bristleback_apps/bm_rbr/CMakeLists.txt
victorsowa12: Done

PR #145, opened 2024-06-17
Sc 200744/usb detection

victorsowa12:
Change the VUSB_DETECT pin from GPIO_Input to EXTI9 so we can detect Rising/Falling edge events and use the interrupt to disconnect/reconnect USB comms.

Note: The Bridge would require a routing change such that `VUSB_DETECT` and `BM_INT` aren't competing for `EXTI0`. This PR only modifies the BSP's for mote based bristlemouth devices.


src/lib/common/pcap.c
```diff
@@ -105,6 +105,5 @@ void pcapEnable() {
 void pcapDisable() {
   if(pcapSerialHandle) {
     pcapSerialHandle->enabled = false;
-    printf("PCAP Stream disabled!\n");
```
victorsowa12, line 108: this causes a crash because `pcapDisable` is executed from the `vusbDetectIRQHandler`.

src/lib/common/serial_console_u5.cpp
```diff
@@ -167,6 +167,7 @@ static void processConsoleRxByte(void *serialHandle, uint8_t byte) {
 void serialConsoleEnable() {
   serialConsoleHandle->enabled = true;
   ulConsoleBuffIdx = 0;
+  serialConsoleHandle->flags &= ~SERIAL_TX_IN_PROGRESS;
```
victorsowa12, line 170: This flag needs to be cleared in order to tell the `serialGenericTx` task that the USB is not currently transmitting. If we happened to have unplugged the USB during a transmission, this flag will not have been cleared, so lets make sure when we plug back in that we clear it. If we don't clear it, the `serialGenericTx` task will never send any bytes out to the USB.

src/bsp/bm_mote_bristleback_v1_0/Core/Src/stm32u5xx_it.c
```diff
@@ -112,6 +112,33 @@ void EXTI8_IRQHandler(void)
...
+  {
+    LL_EXTI_ClearRisingFlag_0_31(LL_EXTI_LINE_9);
+    /* USER CODE BEGIN LL_EXTI_LINE_9_RISING */
+    rval |= STM32IOHandleInterrupt((const STM32Pin_t *)VUSB_DETECT.pin);
+    /* USER CODE END LL_EXTI_LINE_9_RISING */
+  }
+  /* USER CODE BEGIN EXTI9_IRQn 1 */
+  portYIELD_FROM_ISR(rval);
+  /* USER CODE END EXTI9_IRQn 1 */
+}
```
victorsowa12, line 140: this interrupt will call `vusbDetectIRQHandler` from `usb.c:30` for every rising/falling event detected on VUSB_DETECT.

russelldeguzman (approved):
🎉

PR #146, opened 2024-06-27
use configured value for reading period instead of default

russelldeguzman:
use configured value for RBR reading period instead of default


victorsowa12 (approved):
🤘

PR #147, opened 2024-06-27
modify scripts for windows

victorsowa12:
Modiffying the debugging scripts to check if we are on windows and then using the windows executable names instead of the mac/linux ones. 

Mainly using `gdb-multiarch` instead of `gdb` and `python` instead of `python3`.

Check the shortcut ticket for an example of make flash and make debug!

[SC-203450]
https://app.shortcut.com/sofar/story/203450/run-through-all-the-dev-kit-guides-on-windows

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -432,9 +432,9 @@ add_custom_command(
 #		
 if(CMAKE_HOST_WIN32)
 	# This is the windows version of the horrible command
-	set(HORRIBLE_COMMAND, "bash -c ${CMAKE_OBJDUMP} --prefix=${SRC_DIR} --prefix-strip=$$(echo $${PWD//\\//} | tr -cd '/' | wc -c) -S -d -l -w ${EXECUTABLE} | tr -s ' ' | cut -d' ' -f2- | sed 's=${SRC_DIR}/./==' > ${EXECUTABLE}.lst")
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -427,10 +436,46 @@ add_custom_command(TARGET ${EXECUTABLE}
 # to make the .lst file consistent across different build paths.
 #
 # If you come up with a better solution, please remove this :D
+#		
+	COMMAND bash -c "${CMAKE_OBJDUMP} --prefix=${SRC_DIR} --prefix-strip=$$(echo $${PWD//\\//} | tr -cd '/' | wc -c) -S -d -l -w ${EXECUTABLE} | tr -s ' ' | cut -d' ' -f2- | sed 's=${SRC_DIR}/./==' > ${EXECUTABLE}.lst"
+	BYPRODUCTS ${EXECUTABLE}.lst
+)
+else()
+# Generate lst file
+add_custom_command(TARGET ${EXECUTABLE}
```
russelldeguzman, line 440: Is `else()` and below the same as above? Could we put the ` CMAKE_HOST_WIN32` to only encompass the delta in the _horrible looking command_

russelldeguzman (approved):
Small suggestion if it makes sense, otherwise looks good.

src/CMakeLists.txt
victorsowa12: you can't nest the if/else into the `add_custom_command` ( i tried it and it didn't work 😅), but I suppose I could do a 

```
if(win)
 horrible_command = windows horrible command
else()
 horrible_command = normal horrible command
endif()

add_custom_command(
.
.
.
horrible_command
)
```

victorsowa12: Good suggestion! I changed it and it works on both windows and mac!

PR #151, opened 2024-07-19
fix: print the bridge network config in a single pass

towynlin:
Previously the internal cbor stream handler would build a whole publish packet for every character or two!

This resulted in spotter console logs having the bridge network config being slowly printed a little at a time while other lines interrupt it.

Example spotter console log line with a couple nodes attached:

2024-07-19T07:42:38.207Z [BRIDGE_CFG] [INFO] Bridge network config: [[3126851968281441696, "bridge-dbg", 1499363141, 3152952663, {"sampleIntervalMs": 300000, "sampleDurationMs": 240000, "subsampleIntervalMs": 60000, "subsampleDurationMs": 30000, "subsampleEnabled": 0, "bridgePowerControllerEnabled": 0, "alignmentInterval5Min": 1, "ticksSamplingEnabled": 0, "samplesPerReport": 2, "transmitAggregations": 1, "currentReadingPeriodMs": 60000, "softReadingPeriodMs": 500, "rbrCodaReadingPeriodMs": 500, "turbidityReadingPeriodMs": 1000}], [10047966341406729948, "serial_bridge-dbg", 1499363141, 1985325339, {"sensorsPollIntervalMs": 10000, "sensorsCheckIntervalS": 60, "currentAggPeriodMin": 0.f, "nSkipReadings": 0, "readingIntervalMs": 60000, "payloadWdToS": 70, "plUartBaudRate": 9600, "dfu_confirm": 0, "rbrCodaType": 2}], [816502619147696319, "rbr_coda_example-dbg", 553272471, 3352688053, {"sensorsPollIntervalMs": 60000, "rbrCodaType": 2}]]

Notice that currentAggPeriodMin is `0.f` which is annoying to parse in production tooling but we've determined that fixing it isn't worth the effort. It's the way tinycbor does "pretty" output. There are json output functions too, but we'd have to convert most numbers to strings, especially node IDs to prevent them from being rounded, because all json numbers are floating point under the hood.

### DIFF
--- a/src/lib/common/network_config_logger.cpp
+++ b/src/lib/common/network_config_logger.cpp
@@ -1,0 +1,44 @@
+#include "network_config_logger.h"
+#include "FreeRTOS.h"
+#include "bridgeLog.h"
+#include "cbor.h"
+
+static char *log_buffer = NULL;
+static size_t log_buffer_size = 0;
+static constexpr size_t log_buffer_max_size = 2048;
+
+static CborError print_stream_handler(void *out, const char *fmt, ...) {
+  configASSERT(log_buffer != NULL);
+  (void)out;
+
+  va_list args;
+  va_start(args, fmt);
+  int n = vsnprintf(nullptr, log_buffer_max_size - log_buffer_size, fmt, args);
+  if (n > 0) {
+    vsnprintf(&log_buffer[log_buffer_size], log_buffer_max_size - log_buffer_size, fmt, args);
+    log_buffer_size += n;
+  }
+  va_end(args);
+
+  return n < 0 ? CborErrorIO : CborNoError;
+}
+
+void log_cbor_network_configurations(const uint8_t *cbor_buffer, size_t cbor_buffer_size) {
+  configASSERT(cbor_buffer != NULL);
+  configASSERT(log_buffer == NULL);
+  log_buffer = static_cast<char *>(pvPortMalloc(log_buffer_max_size));
+  configASSERT(log_buffer);
+  log_buffer_size = 0;
+  CborParser parser;
+  CborValue it;
+  CborError err = cbor_parser_init(cbor_buffer, cbor_buffer_size, 0, &parser, &it);
+  if (err == CborNoError) {
+    cbor_value_to_pretty_stream(print_stream_handler, NULL, &it, CborPrettyDefaultFlags);
+    bridgeLogPrint(BRIDGE_CFG, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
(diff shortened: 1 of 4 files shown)


src/lib/common/network_config_logger.cpp
```diff
@@ -0,0 +1,44 @@
...
+  configASSERT(log_buffer == NULL);
+  log_buffer = static_cast<char *>(pvPortMalloc(log_buffer_max_size));
+  configASSERT(log_buffer);
+  log_buffer_size = 0;
+  CborParser parser;
+  CborValue it;
+  CborError err = cbor_parser_init(cbor_buffer, cbor_buffer_size, 0, &parser, &it);
+  if (err == CborNoError) {
+    cbor_value_to_pretty_stream(print_stream_handler, NULL, &it, CborPrettyDefaultFlags);
+    bridgeLogPrint(BRIDGE_CFG, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
```
towynlin, line 37: Here's the single call to `bridgeLogPrint`.

src/lib/common/network_config_logger.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+static CborError print_stream_handler(void *out, const char *fmt, ...) {
+  configASSERT(log_buffer != NULL);
+  (void)out;
+
+  va_list args;
+  va_start(args, fmt);
+  int n = vsnprintf(nullptr, log_buffer_max_size - log_buffer_size, fmt, args);
+  if (n > 0) {
+    vsnprintf(&log_buffer[log_buffer_size], log_buffer_max_size - log_buffer_size, fmt, args);
```
towynlin, line 18: We no longer call `vBridgeLogPrint` for every call to this function, which occurs almost for every character. Instead we build the log line in a heap-allocated buffer.

src/lib/common/network_config_logger.cpp
```diff
@@ -0,0 +1,44 @@
...
+  va_end(args);
+
+  return n < 0 ? CborErrorIO : CborNoError;
+}
+
+void log_cbor_network_configurations(const uint8_t *cbor_buffer, size_t cbor_buffer_size) {
+  configASSERT(cbor_buffer != NULL);
+  configASSERT(log_buffer == NULL);
+  log_buffer = static_cast<char *>(pvPortMalloc(log_buffer_max_size));
+  configASSERT(log_buffer);
```
towynlin, line 30: The static pointer variable `log_buffer` must be null before this call, then we allocate it, free it, and set it back to null before returning.

victorsowa12 (approved):
🤘

PR #153, opened 2024-07-30
Fix config check for Raw Pressure Processor Init code

victorsowa12:
Move up the raw pressure processor init code so that it get/sets configs before NCP code can begin to process incoming packets.

See comment below for actual fix + explanation

### DIFF
--- a/src/apps/bridge/app_config.cpp
+++ b/src/apps/bridge/app_config.cpp
@@ -161,7 +161,7 @@ raw_pressure_config_s getRawPressureConfigs(cfg::Configuration &syscfg) {
     save_config = true;
   }
   cfg.rbrCodaReadingPeriodMs = RbrCodaSensor::DEFAULT_RBR_CODA_READING_PERIOD_MS;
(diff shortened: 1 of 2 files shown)


towynlin (approved):

victorsowa12:
So one drawback of this quick fix, is if we don't have the configs we will try to do a bridge CFG log print saying we are using the hardcoded default value. Since we have moved this print to before the NCP init, we don't yet have our TX callback set and therefore will not send the print to the Spotter.

Side note: this minor bug of not printing that we are using the defaults is already present in the power config init code since this also occurs before ncpInit!

towynlin (approved):

victorsowa12:
Alright so the latest change still suffers from the chance of a crash if the configs are not set yet(this should only happen after a flash/MCUboot update AND if the configs were not set previously). BUT now we will not be resetting the configs each boot. After the first successful boot the configs will be set to the defaults. Once they are set we will no longer have to call the set/save for `rbrCodeReadingPeriodMs` and thus we will not delay the default task just enough to prevent NCP UART packet processing.

PR #155, opened 2024-07-31
Adds Linked List To Sensor Sampler

matt001k:
Expands sensor list array to utilize linked list
Adds static allocation based linked list module in lib/common Updates sensorSamplerAdd calls to utilize update API

This needs to be tested on actual hardware (I don't have any yet 😄, idk if a dev kit can be sent before the 26th).
Wanted to get my hands a bit dirty by digging in the code a bit.
Thought taking on a TODO would be a good first step.
Utilized a LL library I wrote a while back that does not utilize any heap.

I was able to run the unit tests and might add unit tests for both of these lib items.


test/src/common/bcmp_dfu_ut.cpp
```diff
@@ -118,7 +118,7 @@ class BcmpDfuTest : public ::testing::Test {
         .hwVersion = 1,
         .flags = 0,
         .versionStrLen = 5,
-        .versionStr = {"test"},
```
matt001k, line 121: This is apparently a C99 bug that was causing this test to fail. The suggested fix was found here:
https://pcbartists.com/firmware/designator-outside-aggregate-initializer-solved/

towynlin:
Thanks @matt001k! 👏 I approved CI workflows to run, and it caught a failure, so check that out.

test/src/common/bcmp_dfu_ut.cpp
matt001k: @towynlin I would double check this runs fine on Mac. I have seen silly things like this break across Linux->Mac and visa-versa.I see our CI uses Ubuntu. Let me know!

towynlin: It does in fact fail to build on my mac arm machine:
```
bm_protocol/test/src/common/sensorSampler_ut.cpp:99:17: error: braces around scalar initializer [-Werror,-Wbraced-scalar-init]
      node[i] = SENSOR_NODE_INIT(testInit, testPoll, testCheck);
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
bm_protocol/src/lib/sensor_sampler/sensorSampler.h:21:20: note: expanded from macro 'SENSOR_NODE_INIT'
          .timer = {0}, \
                   ^~~
```
I'm looking into it.

towynlin: I'm noticing now that I see a different failure than you called out here. Sorry! I'll clarify once I figure that out.

src/lib/sensor_sampler/sensorSampler.h
```diff
@@ -1,38 +1,30 @@
...
-  RTCTimeAndDate_t rtcTime;
-  float temperature;
-  float humidity;
-} __attribute__((packed)) humTempSample_t;
+/// Used to initialize sensor node
+#define SENSOR_NODE_INIT(init, sample, check) \
+    { .list = { \
+          .sensor = { .initFn = init, .sampleFn = sample, .checkFn = check }, \
+          .name = NULL, \
+          .timer = {0}, \
```
towynlin:
```suggestion
          .timer = NULL, \
```

towynlin: This change fixed the test build for me. Took me a while to wrap my head around where this came from. All the comparisons for timer in the code are whether `timer == NULL`, and the original initialization prior to this PR was in `sensorSamplerAdd` where each item gets malloc'd and memset to 0.

test/src/common/bcmp_dfu_ut.cpp
towynlin: So yeah, thanks for calling out the need to test. It passes on my machine. That said, I see we don't have any tests that exercise whether `versionStr` is set as expected. Could you add that expectation to at least one of the tests?

towynlin:
@matt001k Do you have all the rights to contribute the linked list library to Bristlemouth under the Apache 2 license?

src/lib/common/ll.h
```diff
@@ -0,0 +1,119 @@
...
+ *          to determine if the iterator is at the end of the list
+ *          (ll->p == NULL).
+ *
+ * @param ll list to pop node from
+ * @param data data of popped node
+ *
+ * @return 0 on success
+ * @return stderr on failure
+ *****************************************************************************/
+int LLPop(LL_t *ll, void **data);
```
towynlin: Since "pop" as a name is typically used in stacks and removes the element last added to the end of the stack, I think this should have a different name. Since this function and `LLReset` are mainly provided to assist in traversal, my suggestion would be:

- Rename `LL_t` member `p` to `cursor`
- Rename `LLReset` to `LLResetCursor` ("reset" makes me think it will clear the list)
- Rename `LLPop` to `LLAdvanceCursor` or maybe `LLIterate` or `LLIterateCursor`
- Make both of those ☝️ functions private (remove from .h, make static in .c)

And thus make `LLTraverse` the only obvious way to iterate over the list, removing some potential confusion by future developers. Thoughts?

matt001k: I like that! Agreed on the naming scheme, this helps prevent confusion.
Making these functions static will also limit redundant API that developers most likely will not utilize.

test/src/common/bcmp_dfu_ut.cpp
matt001k: Will do!

src/lib/sensor_sampler/sensorSampler.h
towynlin: Ah! And just sharing something I just realized. I was wondering, since we were using `pvPortMalloc` for each sensor before, and since you removed that, where did the same info go and how is it allocated. Every place that `SENSOR_NODE_INIT` is called is in a static variable initializer in the individual sensor sampler cpp files, like htuSampler.cpp, loadCellSampler.cpp, powerSampler.cpp, etc. Those statically allocated items get passed to `sensorSamplerAdd` where we save a pointer to the node in `LLNodeAdd`.

matt001k:
@towynlin yep! This is something I developed on my own time a while back when challenging myself to use data structures with static allocation.

test/src/common/bcmp_dfu_ut.cpp
matt001k: I added an `EXPECT_STREQ` for `versionStr` to ensure it is properly set during `SetUp` of the tests.

src/lib/common/ll.h
matt001k: Updated!

src/apps/bm_devkit/bm_loadcell/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ set(LIB_FILES
     ${SRC_DIR}/lib/common/uptime.c
     ${SRC_DIR}/lib/common/util.c
     ${SRC_DIR}/lib/common/watchdog.c
+    ${SRC_DIR}/lib/common/ll.c
```
matt001k, line 63: @towynlin  Just curious, and maybe we can have this discussion at another time, but why do we not build the lib/bsp/third_party sources as libraries separately and then link them in? I feel like a lot of redundancy could be cut down here to ensure that new source files do not have to be added to to each app, but that's an assumption and I do not have the full background of the why!

towynlin: No particular reason — has occurred to me too — could totally do that in a separate PR.

towynlin:
@matt001k @victorsowa12 How does this apply to our work these days? Should we merge it or close it?

matt001k:
@towynlin I saw we are good to close it!

PR #157, opened 2024-08-21
fix: keep diff sig capacity const, reset total samples each pass

towynlin:
Since creating the difference signals shrinks the total number of samples (passed by reference) by one each time, we need to reset to the full capacity each time. This fixes a bug where the total samples in each message got progressively smaller and smaller over time during a deployment.

### DIFF
--- a/src/apps/bridge/rbrPressureProcessor.cpp
+++ b/src/apps/bridge/rbrPressureProcessor.cpp
@@ -92,10 +92,9 @@ bool rbrPressureProcessorIsStarted(void) { return _ctx.started; }
 static void runTask(void *param) {
   (void)param;
   BmRbrDataMsg::Data rbr_data;
-  uint32_t num_samples = (_ctx.rawSampleS * 1000) / _ctx.rbrCodaReadingPeriodMs;
-  size_t diffSignalCapacity = num_samples;
+  const uint32_t diffSignalCapacity = (_ctx.rawSampleS * 1000) / _ctx.rbrCodaReadingPeriodMs;
   DifferenceSignal diffSignal(diffSignalCapacity);
-  const size_t d_n_size = num_samples * sizeof(double);
+  const size_t d_n_size = diffSignalCapacity * sizeof(double);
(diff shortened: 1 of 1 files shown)


victorsowa12 (approved):
🙌

evanShap (approved):
- Edge case to ponder that I don't fully grok: we're not guaranteed to receive (rawSampleS * 100) / rbrCodaReadingPeriodMs samples from the pressure sensor.
-- If we receive more than this, I expect we just truncate those.
-- If we receive less than this, do we only send what we received, or do we end up with random data or 0s to fill out the buffer up to diffSignalCapacity?

Otherwise, LGTM!

towynlin:
Great questions @evanShap. Thanks. 

> * Edge case to ponder that I don't fully grok: we're not guaranteed to receive (rawSampleS * 100) / rbrCodaReadingPeriodMs samples from the pressure sensor.
>       -- If we receive more than this, I expect we just truncate those.

Yes, because of https://github.com/bristlemouth/bm_protocol/blob/04d2a0db7e5c1eb8c5836567f1f469f9dffb07c5/src/lib/common/differenceSignal.cpp#L25 where `r_i` is the number of readings added so far and `r_n` is the capacity of the buffer, we only add to the difference signal buffer up to the end and then just ignore subsequent readings.

> If we receive less than this, do we only send what we received, or do we end up with random data or 0s to fill out the buffer up to diffSignalCapacity?

We only send what was received, because here https://github.com/bristlemouth/bm_protocol/blob/04d2a0db7e5c1eb8c5836567f1f469f9dffb07c5/src/lib/common/differenceSignal.cpp#L50-L51 `num_samples` is passed by reference and updated to contain only as many as we've actually received (minus one, because it's the first-order difference signal).

Then when encoding to cbor, we only encode `num_samples` values from the second-order difference signal here https://github.com/bristlemouth/bm_common_messages/blob/2a44e0ca19ce4eae3174c7ec77a16e95d4c49966/bm_rbr_pressure_difference_signal_msg.cpp#L156-L157

PR #159, opened 2024-08-29
fix: drop rbr readings that arrive after timer fires

towynlin:
One task catches RBR pressure readings and enqueues them. Another task reads from the queue and adds the readings to the difference signal buffer. When the "end of sample" timer fires, we already don't enqueue any more readings. However, the question is how we should deal with the ones that might be sitting in the queue, not yet processed.

After discussion, we agreed that it's better to drop these on the floor rather than add them to the beginning of the next sample, since the difference signal buffer may already have been cleared.

The tradeoff here is that we may drop valid readings at the tail end of a sample before calculating the difference signal and sending messages to Spotter. That's acceptable.

### DIFF
--- a/src/apps/bridge/bridgePowerController.cpp
+++ b/src/apps/bridge/bridgePowerController.cpp
@@ -235,10 +235,11 @@ void BridgePowerController::_update(void) {
         } else { // Subsampling disabled
           stateLogPrintTarget("Sample", currentCycleS + sampleTimeRemainingS);
 #ifdef RAW_PRESSURE_ENABLE
-      if(!rbrPressureProcessorIsStarted()) {
-        rbrPressureProcessorStart();
-        printf("Started rbrPressureProcessor\n");
-      }
+          if (!rbrPressureProcessorIsStarted()) {
+            rbrPressureProcessorStart();
+            bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
+                           "Started rbrPressureProcessor\n");
+          }
 #endif // RAW_PRESSURE_ENABLE
           powerBusAndSetSignal(true);
           time_to_sleep_ms = MAX(sampleTimeRemainingS * 1000, MIN_TASK_SLEEP_MS);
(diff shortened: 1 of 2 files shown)


src/apps/bridge/rbrPressureProcessor.cpp
```diff
@@ -178,13 +178,21 @@ static void runTask(void *param) {
     }
     if (bits & kQRcv) {
       if (xQueueReceive(_ctx.q, &rbr_data, 0) == pdTRUE) {
-        if (diffSignal.addSample(rbr_data.pressure_deci_bar)) {
-          bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
-                         "%s Added sample %.2f to Diff signal \n", kRbrPressureProcessorTag,
-                         rbr_data.pressure_deci_bar);
+        if (_ctx.started) {
+          if (diffSignal.addSample(rbr_data.pressure_deci_bar)) {
```
towynlin, line 182: This is the most important change — checking `_ctx.started` to make sure the timer is currently running before adding a reading to the difference signal buffer.

src/apps/bridge/rbrPressureProcessor.cpp
```diff
@@ -178,13 +178,21 @@ static void runTask(void *param) {
...
+                           "%s Unable to add reading to Diff signal, buffer is full\n",
+                           kRbrPressureProcessorTag);
+          }
         } else {
-          bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_WARNING, USE_HEADER,
-                         "%s Unable to add sample to Diff signal \n", kRbrPressureProcessorTag);
+          bridgeLogPrint(
+              BRIDGE_SYS, BM_COMMON_LOG_LEVEL_WARNING, USE_HEADER,
+              "%s Not adding late reading %.4f to Diff signal because timer has stopped\n",
+              kRbrPressureProcessorTag, rbr_data.pressure_deci_bar);
```
towynlin, line 195: New warning log line in BRIDGE_SYS to track these late arrivals.

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -235,10 +235,11 @@ void BridgePowerController::_update(void) {
...
 #ifdef RAW_PRESSURE_ENABLE
-      if(!rbrPressureProcessorIsStarted()) {
-        rbrPressureProcessorStart();
-        printf("Started rbrPressureProcessor\n");
-      }
+          if (!rbrPressureProcessorIsStarted()) {
+            rbrPressureProcessorStart();
+            bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
+                           "Started rbrPressureProcessor\n");
+          }
```
towynlin, line 242: When debugging this, it would have been nice to have "Started rbrPressureProcessor" in the SD card logs, so I changed this from `printf` to `bridgeLogPrint`.

victorsowa12 (approved):
🤘

matt001k (approved):

PR #162, opened 2024-09-19
Feat/rs232 bsp

victorsowa12:
I have created a new apps folder: `rs232_expander_apps` where we follow the same folder structure as BMDK/bristleback. There is an `rs232_expander_apps_common` where the app_main.cpp lives. This is a modified copy of the bristleback one. I also copied/modified a `serial_payload_example` to run using rs232 expander board. I believe `serial_payload_example` is sufficient as a "minimal" test app since it provides access to toggling of the gpios. See [this comment](https://github.com/bristlemouth/bm_protocol/pull/162#discussion_r1767361758) below for my train of thought. 

To build you need a cmake command like:
```
cmake ../.. -DCMAKE_TOOLCHAIN_FILE=../../cmake/arm-none-eabi-gcc.cmake -DBSP=bm_mote_rs232 -DAPP=serial_payload_example -DCMAKE_APP_TYPE=rs232_expander -DCMAKE_BUILD_TYPE=Debug -DUSE_BOOTLOADER=1
```

In the BSP I disabled all of the unused gpios/peripherals:
Screenshot 2024-09-19 at 10 30 02 AM

One side effect of turning off all the unused peripherals is we no longer have a USART peripheral, only the LPUART(this is the PAYLOAD_UART). CubeMX creates different LL/HAL files for these peripherals and we need the USART ones for serial.c to funciton properly. Funcitonally these are exactly the same, just different naming for usart v.s. lpuart. It would be wise at some point to modify serial.c such that it doesn't depend on there being a USART, but in the mean time we just need to make sure that if this BSP is regenerated by CubeMX, we don't commit the deletion of the usart LL files.


src/bsp/bm_mote_rs232/Drivers/STM32U5xx_HAL_Driver/Inc/stm32u5xx_ll_usart.h
```diff
@@ -0,0 +1,4488 @@
+/**
```
victorsowa12, line 1: Until we make `serial.c` agnostic to USART v.s. LPUART, we need to make sure this stays in the bsp.

src/bsp/bm_mote_rs232/Drivers/STM32U5xx_HAL_Driver/Src/stm32u5xx_ll_usart.c
```diff
@@ -0,0 +1,446 @@
+/**
```
victorsowa12, line 1: Same as the .h, Until we make serial.c agnostic to USART v.s. LPUART, we need to make sure this stays in the bsp.

src/lib/common/serial.c
```diff
@@ -17,8 +17,6 @@
 
 #include "lpm.h"
 
-// Log_t *SerialLog;
```
victorsowa12, line 20: Old stuff from when it was copied form spotter repo

src/bsp/bm_mote_rs232/bsp.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  spiInit(&spi2);
+  spiInit(&spi3);
+  i2cInit(&i2c1);
+
+  // Turn on Adin2111
+  IOWrite(&ADIN_PWR, 1);
+
+  // Turn off the BUCK by default (it's enabled when low)
+  IOWrite(&PL_BUCK_EN, 1);
```
victorsowa12, line 69: Do I need to have a default value for the DISCHARGE_ON pin?

src/bsp/bm_mote_rs232/bsp.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
```
victorsowa12, line 1: Not auto-generated

src/bsp/bm_mote_rs232/bsp.cpp
```diff
@@ -0,0 +1,88 @@
+#include "bsp.h"
```
victorsowa12, line 1: not auto-generated

src/bsp/bm_mote_rs232/bsp.h
victorsowa12, line 1: not auto-generated, please double check

src/bsp/bm_mote_rs232/bsp_pins.c
```diff
@@ -0,0 +1,22 @@
+#include "bsp.h"
```
victorsowa12, line 1: not auto-generated, please double check

src/bsp/bm_mote_rs232/Core/Inc/main.h
```diff
@@ -0,0 +1,122 @@
...
+#include "stm32u5xx_ll_lpgpio.h"
+#include "stm32u5xx_ll_bus.h"
+#include "stm32u5xx_ll_cortex.h"
+#include "stm32u5xx_ll_utils.h"
+#include "stm32u5xx_ll_pwr.h"
+#include "stm32u5xx_ll_dma.h"
+
+/* Private includes ----------------------------------------------------------*/
+/* USER CODE BEGIN Includes */
+#include "stm32u5xx_ll_usart.h"
```
victorsowa12, line 49: force the inclusion of ll usart.h for serial.c to be able to function

src/apps/rs232_test/app_main.cpp
```diff
@@ -0,0 +1,315 @@
...
+  startIWDGTask();
+  startSerial();
+
+  startSerialConsole(&usbCLI);
+  // Serial device will be enabled automatically when console connects
+  // so no explicit serialEnable is required
+  pcapInit(&usbPcap);
+
+  startCLI();
+  startDebugUart();
```
victorsowa12: This app(is deleted in the latest commits) is a blend of bringup + bristlemouth/existing app features. It has all of the bristlemouth features, but does not use the "user_code" and "payload_uart" to collect data from a serial sensor. Instead it has debug_uart included and you need to use `serial enable lpuart` `serial baud lpuart 9600` to be able to talk to an RBR. 

On the contrary, the power monitor running on the I2C bus is running automatically with the power/sensor sampler. 

If this feels a bit redundant, then perhaps I can just create a "serial_payload_example" app and get rid of this app. 

Or if we need a true bringup app, then I can create that here.

victorsowa12: I talked with Sophie and she menitoned that as far as production goes the mote goes through bringup and then after the expander board is attached there really isn't another step of bringup requiring the ability to turn things like I2C/serial on/off.

So I think just creating a serial_payload_example app will suffice. These apps already have the ability to toggle the gpio's on/off so I think that will be enough for most testing.

victorsowa12: With that in mind I'm thinking to not merge this app as is, but convert it into a "bm_rs232_apps/serial_payload_example" type of app.

src/bsp/bm_mote_rs232/bsp.cpp
victorsowa12: After talking to Sohpie, the DISCHARGE_ON should default low so that only the apps that need it (RBR) toggle it high when necessary.

src/apps/rs232_test/app_main.cpp
victorsowa12: In the latest commits I have created a serial_payload_example to run on the rs232 and removed the rs232_test app.

src/CMakeLists.txt
```diff
@@ -222,9 +222,11 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
   set(APP_NAME ${APP_NAME}-dbg)
 endif()
 if(CMAKE_APP_TYPE STREQUAL "BMDK")
-    add_subdirectory(apps/bm_devkit/${APP})
+	add_subdirectory(apps/bm_devkit/${APP})
 elseif(CMAKE_APP_TYPE STREQUAL "bristleback")
-    add_subdirectory(apps/bristleback_apps/${APP})
+	add_subdirectory(apps/bristleback_apps/${APP})
+elseif(CMAKE_APP_TYPE STREQUAL "rs232_expander")
```
victorsowa12, line 228: add new app type... not a huge fan of the name `rs232_expander` but I think that "rs232" would be too generic since you can technically do rs232 on both the devkit and bristleback.

src/bsp/bm_mote_rs232/bsp.cpp
victorsowa12: done

src/apps/rs232_test/app_main.cpp
victorsowa12: I can't revert to the commit where both the serial_paylaod_example and rs232_test app exist if we want both.

victorsowa12:
To keep the PR's smaller I will make the rbr, turbidity, and aanderaa in a seperate PR each.

matt001k:
Code looks good, was able to flash and run fine. Nice work!

matt001k (approved):

src/apps/rs232_test/app_main.cpp
towynlin: No need to revert — that test app was just my WIP, making sure I understood everything. I like serial_payload_example as the final thing that stays in the code. 👍

src/CMakeLists.txt
towynlin: Good call.

src/bsp/bm_mote_rs232/bsp_pins.c
towynlin: Looks great — thanks so much for the cleanup. I see that you removed a pin named `BB_3V3_EN` — what is that on Bristleback? Is there a broken out 3.3V rail?

_... opens altium to answer own question..._

Ah, I see it. Yeah, and there's a load switch for that rail, which is what that pin is for. 👌 🙇‍♂️

towynlin (approved):
Great work!

victorsowa12:
I made a ticket for updating `serial.c`: https://app.shortcut.com/sofar/story/207207

PR #167, opened 2024-09-22
Feature/pluart hd transactions

evanShap:
This is on-path to the Rx-Live demo, which s a Half Duplex RS-485 device.
### 1
For Half Duplex devices, we need to manage the transition in and out of transmitting mode. To do this, I added a generic model of "transactions" to the PLUART interface:
- user can assign start and end functions to bookend a transaction.
- user is responsible for marking a transaction start with call to `startTransaction`, which will call the start function exactly once.
- user can then execute from 0 to any number of `write`s within the context of the transaction.
- user is responsible for marking a transaction end with a call to `endTransaction`, which will eventually call the end function exactly once. This will either happen immediately of all writes are complete, or it will be triggered by the PLUART TC interrupt when the final write eventually completes.
- I updated usage in pluart debug CLI to wrap write in a transaction.

**All of this is expected to have no impact on our FD RS232 usage, except maybe for some tweaks to the serial.c ISR handling to make TC -> CB function call more reliable?**

### 2
I implemented the RS-485 Half Duplex helper functions in the Bristlefin class to set HD mode, enable Tx, enable Rx.

### 3
All of this is being manually tested for RS485 HD in the in-progress Rx Live Demo app. Usage simply looks like:
```
  // Set the RS485 transceiver to Half-Duplex
  bristlefin.setRS485HalfDuplex();
  // Setup PLUART transactions for proper RS485 enable Tx / enable Rx
  PLUART::enableTransactions(Bristlefin::setRS485Tx, Bristlefin::setRS485Rx);
```
- Working through some edge case scenarios on paper, will add doc to the PR later.
- Should manually test on RBR Coda and Aanderaa apps.
- **one open question - what should the task priority of the postTxCallback executor task be? Note there will be silly stuff in that function (like 1ms delay for transceiver switching setups), so should probably be lower priority than the UART task?**


src/lib/common/payload_uart.cpp
```diff
@@ -28,6 +28,97 @@ static bool _useLineBuffer = true;
...
+    taskEXIT_CRITICAL();
+  }
+}
+
+// Post-transmission callback function, attached as uart_handle.postTxCb and called
+//  from ISR when the TC interrupt fires [see common/serial.c::serialGenericUartIRQHandler].
+static void pluartPostTxCallback(SerialHandle_t *handle) {
+  (void)handle; // Mark 'handle' as intentionally unused
+
+  BaseType_t xHigherPriorityTaskWoken = pdFALSE;
```
evanShap, line 104: @towynlin @victorsowa12 @matt001k  FYI - I don't fully grok the safety handling for ENTER / EXIT_CRITICAL_FROM_ISR, which includes detection and yielding if higher priority preemption, preserving saved interrupt context.

This was all suggested by ChatGPT. 😬

src/lib/common/payload_uart.cpp
```diff
@@ -208,6 +331,10 @@ BaseType_t init(uint8_t task_priority) {
                                 2048, &PLUART::uart_handle, task_priority, NULL);
   configASSERT(rval == pdTRUE);
 
+  // Create the task to execute the _postTxFunction callback
```
evanShap: @victorsowa12 @towynlin @matt001k - Not sure if the PostTxCBTask should have the same priority we pass to `init`. It's gonna be relatively low priority and often include delays, so should probably be user task priority?

src/lib/common/payload_uart.cpp
```diff
@@ -188,6 +287,26 @@ SerialHandle_t uart_handle = {
...
+        _postTxFunctionPending = false;
+        taskEXIT_CRITICAL();
+        _postTxFunction(); // Safely call the function
+      } else {
+        taskEXIT_CRITICAL();
+      }
+    }
+  }
+}
+
```
victorsowa12: What if instead of spinning up another task, `endTransaction()` has the wait for semaphore such that the `endTransaction()` is responsible for calling the post TX function (delays setting pins etc)?

src/lib/common/payload_uart.cpp
victorsowa12: It mainly means that in ciritical secitons the scheduler is not going to do any switches and interrupts are disabled. In this case it will ensure that nothing else will access `_postTxFunctionPending` while we access it here.

src/lib/common/serial.c
```diff
@@ -244,8 +245,10 @@ void serialGenericUartIRQHandler(SerialHandle_t *handle) {
     }
   }
 
-  if (!bytesAvailable && LL_USART_IsActiveFlag_TC((USART_TypeDef *)handle->device) && !usart_IsEnabledIT_TXE((USART_TypeDef *)handle->device)) {
+  // Check if transmission just completed, and clear flag if so
+  if (LL_USART_IsActiveFlag_TC((USART_TypeDef *)handle->device)) {
```
victorsowa12: this will require a good amount of testing to validate. I don't see anything inherently wrong with it, but the previous if statement(the deleted line) was created when fixing the RBR freezer reset bug. So we will likely need to set up a serial trace and/or freezer tests to make sure that doesn't come back.

src/lib/common/payload_uart.cpp
matt001k: This is an appropriate use case for entering/exiting sections critical sections within ISR context (prevents lower priority interrupts from pre-empting this service routine). The yielding will essentially allow the `postTxCBTask` task to pre-empt other tasks of lower priority which is necessary due to the semaphore giving here. The answer to this question provides a nice graphical look at what not running this would look like: https://stackoverflow.com/questions/58631246/freertos-why-to-call-taskyield-from-isr-method-within-the-isrhandler

src/lib/common/payload_uart.cpp
matt001k: To me it feels that if the callback might have random delays and potentially unknown behavior, it should be low priority to prevent any funky handling with other tasks.

src/lib/common/payload_uart.cpp
victorsowa12: I need to put some more thought into this, but I think there is a path where we don't have the `postTxCbTask` and instead make more use of the `preTxCb` and `postTxCb` that are called in serial.c. 

Perhaps we can use a counting semaphore, such that each preTxCb will take(increment) the semaphore and then each postTxCb will give(decrement) the semaphore. This can only work so long as the serial tx task is higher priority than the task that calls serialTx, thus the Tx event in the serial tx queue should be processed (and increment the semaphore)  before we reach the endTransaction. 

Another possibility is that the pluart::write function increments the semaphore, therefore the user task (or other task) will increment the semaphore, and then the postTxCb will give it back(decrement it).

Then we can wait in `endTransaction` until the counting semaphore is 0 and then do the post tx function stuff.

victorsowa12: I'm mainly trying to think of ways to avoid the need of another task and keep the task interactions isolated to between the user_task (or whoever is calling the pluart functions) and the serial task.

src/lib/common/serial.c
```diff
@@ -54,6 +54,7 @@ void startSerial() {
   configASSERT(rval == pdTRUE);
 }
 
+// ReSharper disable once CppDFAEndlessLoop
```
victorsowa12: Is this some editor/code formatting thing?

victorsowa12 (commented):
Put some thoughts down on a potential way to hopefully simplify the task interactions. lmk you thoughts or questions.

src/lib/common/payload_uart.cpp
evanShap: @victorsowa12 The downside would be some blocking delay in the user task until all Tx are all complete, right? That would probably be acceptable.

src/lib/common/serial.c
evanShap: @victorsowa12 I'd vote for reverting this change then, and keeping that as a separate item.
The intent of the change is to make the TC -> callback firing more reliable. However, we don't have anything that uses this feature currently, so we don't know if there are any reliability issues.

src/lib/common/serial.c
evanShap: Yea, I'll delete

src/lib/common/payload_uart.cpp
victorsowa12: Yes correct, you would need some sort of polling loop in `endTransaction` like:
```
end Transaction() {
  while (semaphoreGetCount != 0) {
    delay(1);
  }
  // do end of tx stuff
}
```

victorsowa12: It would prevent someone from doing this:
```
start();
write();
end();  ---> if this is handled in another task then the next start may overlap during the delay this end may have
start();
write();
end();
```

victorsowa12: And then yeah any _postTxFunction delays would have to occur in the user context as well.

src/lib/common/serial.c
victorsowa12: Did you experience any issues with it in your tests?

evanShap:
@victorsowa12 I liked your suggestions of:
- no new tasks, try to keep things simpler.
- endTransaction should block the user's calling task to keep things simpler. If they need concurrency, they can add their own additional tasks.
- FYI - I don't think the counting semaphore idea works because there's not a 1:1 correspondence between calls to `write()` and UART transmissions complete. Often, data from subsequent calls to `write` all end up in the streambuffer together, and will be effectively grouped into a single transmission.

changes:
- simplified state complexity and flow.
- `write` takes token (polling so do not block if not available).
- TC callback gives token (able to remove all if conditions from the ISR.
- endTransaction waits for token (there will always be 1 at the end of all the writes in a transaction.

- There are some open questions in TODOs in payload_uart.cpp.
- I added an md doc that runs through the flow of some main cases.

victorsowa12:
@evanShap did you push the commit with your changes?

evanShap:
> @evanShap did you push the commit with your changes?

Woops, pushed now @victorsowa12

src/lib/common/payload_uart.cpp
```diff
@@ -28,6 +28,99 @@ static bool _useLineBuffer = true;
...
+  if (!_transactionInProgress) {
+    taskEXIT_CRITICAL();
+    return false;
+  }
+
+  // If there were no writes during the transaction, don't wait for write complete semaphore token.
+  if (!_writeDuringTransaction) {
+    _transactionInProgress = false;
+    taskEXIT_CRITICAL();
+    _postTxFunction(); // Safely call the function
```
matt001k: Do we want to check that this is non-NULL before calling?

src/lib/common/payload_uart.cpp
```diff
@@ -28,6 +28,99 @@ static bool _useLineBuffer = true;
...
+  //    which will give a token to the semaphore.
+  // Each subsequent write inside the transaction will take from the semaphore,
+  //    so only the final write will leave a token available here.
+  bool ret_value = true;
+  if (xSemaphoreTake(_postTxSemaphore, pdMS_TO_TICKS(wait_ms)) == pdFALSE) {
+    // Failed to take sempaphore within transaction timeout
+    ret_value = false; // TODO - maybe better to assert(false) here?
+  }
+  // Call the post-transaction function after the final transmission is fully complete
+  if (_postTxFunction != nullptr) _postTxFunction();
```
matt001k, line 95: I like this a lot better, I can understand wanting to offload it into its own task, but this gives a nice way to give the user control of when this callback will fire. Creative and intuitive.

src/lib/common/payload_uart.cpp
evanShap: oooo - good catch, thanks @matt001k

evanShap: fix pushed

src/lib/common/payload_uart.cpp
evanShap: 👍

src/lib/common/payload_uart.cpp
```diff
@@ -28,6 +28,97 @@ static bool _useLineBuffer = true;
...
+  uart_handle.postTxCb = pluartPostTransactionCb;
+}
+
+// Start a Tx transaction
+// TODO - return false if transaction already started?
+void startTransaction() {
+  taskENTER_CRITICAL();
+  _transactionInProgress = true;
+  _writeDuringTransaction = false;
+  taskEXIT_CRITICAL();
```
victorsowa12: I think (would be good to double check) that all the `taskENTER/EXIT_CRITICAL()`'s surrounding `_transactionInProgress` and `_writeDuringTransaction` can be removed since these variables should now only be set/read from one task.

victorsowa12 (approved):
One more suggestion but looks good! Much easier to follow what the flow is!

matt001k (approved):

src/lib/common/payload_uart.cpp
evanShap: Yep, you're right - good catch. Removed them.

evanShap:
Manually tested on Rx-LIVE app and RBR BristleBack app - merging!

PR #168, opened 2024-09-23
LineParser and sub-classes support string value types

evanShap:
This adds support for strings in the LineParser base class and propagates to its derived subclasses OrderedSeparatorLineParser and OrderedKVPLineParser.

This is useful for more complex serial frame string formats, like that for the new VEMCO Rx-Live integration:
```
Example: 667057,004,2019-09-05 10:21:24.106,A69-9006,1025,123,S=63.0,N=39.0,C=0,#4B\r\n
```
NOTE - the changes in OrderedSeparatorLineParser are entirely NO-OP – code comments and a change for clarity that doesn't actually change our execution.


src/lib/common/LineParser.cpp
```diff
@@ -21,12 +21,31 @@ bool LineParser::init() {
...
     }
     return true;
   }
   return false;
 }
 
+LineParser::~LineParser() {
+  if (_values != nullptr) {
+    // Free any allocated string values
+    for (size_t i = 0; i < _numValues; ++i) {
```
matt001k: Is there a reason we pre-increment i on both this loop and the loop on line 22?

src/lib/common/LineParser.cpp
```diff
@@ -63,24 +82,29 @@ bool LineParser::parseLine(const char* line, uint16_t len) {
...
+const Value& LineParser::getValue(uint16_t index) {
   if (_values == nullptr) {
     printf("ERR Parser values uninitialized!\n");
+    static Value invalidValue = {TYPE_INVALID, {0}};
+    return invalidValue;
   }
   else if (index >= _numValues) {
     printf("ERR Parsed value at %u out of range %zu!\n", index, _numValues);
+    static Value invalidValue = {TYPE_INVALID, {0}};
+    return invalidValue;
```
matt001k: Nitpick but can this be a singular `static Value` before the initial `if` statement rather than 2 individual? Small optimization but it also reads a bit better.

src/lib/common/LineParser.cpp
evanShap: @matt001k just habit, will revert as I see post-increment is the prevalent style in this project.

src/lib/common/LineParser.cpp
evanShap: yep, good catch

src/lib/common/LineParser.cpp
matt001k: I was a bit sleepy this morning and I was thinking that pre-incrementing was actually starting our loop at 1, sorry for the extreme nit pick, oops 💤

matt001k (approved):

src/lib/common/LineParser.cpp
evanShap: no problem - style consistency is still worthwhile imo

victorsowa12 (approved):

towynlin (approved):

PR #170, opened 2024-09-24
add the rs232 apps to the release build

victorsowa12:
I forgot to add the new rs232 apps to the release build. Should I also remove the bristleback bsp version of these apps from the `default.yml`?


towynlin:
Ah, thank you!
- Yes, please remove the bristleback versions.
- I suspect we don't need to build the aanderaa debug version anymore for releases. Anybody have evidence to the contrary?

victorsowa12:
I don't see any reason to keep the aanderaa debug build. Removing it!

matt001k (approved):

PR #171, opened 2024-09-26
Update bcmp_ping

victorsowa12:
update bcmp_ping (echo) and now its only non-bcmp/core/common dependency is "device_info"
Screenshot 2024-09-26 at 2 29 26 PM

### DIFF
--- a/src/lib/bcmp/bcmp_ping.cpp
+++ b/src/lib/bcmp/bcmp_ping.cpp
@@ -31,12 +31,12 @@ BmErr bcmp_send_ping_request(uint64_t node_id, const void *addr, const uint8_t *
 
   uint16_t echo_len = sizeof(BcmpEchoRequest) + payload_len;
 
-  uint8_t *echo_req_buff = static_cast<uint8_t *>(bm_malloc(echo_len));
+  uint8_t *echo_req_buff = (uint8_t *)bm_malloc(echo_len);
   // configASSERT(echo_req_buff);
 
   memset(echo_req_buff, 0, echo_len);
 
-  BcmpEchoRequest *echo_req = reinterpret_cast<BcmpEchoRequest *>(echo_req_buff);
+  BcmpEchoRequest *echo_req = (BcmpEchoRequest *)echo_req_buff;
 
   echo_req->target_node_id = node_id;
   echo_req->id = (uint16_t)getNodeId(); // TODO - make this a randomly generated number
@@ -53,7 +53,7 @@ BmErr bcmp_send_ping_request(uint64_t node_id, const void *addr, const uint8_t *
   // Lets only copy the paylaod if it isn't NULL just in case
   if (payload != NULL && payload_len > 0) {
     memcpy(&echo_req->payload[0], payload, payload_len);
-    _expected_payload = static_cast<uint8_t *>(bm_malloc(payload_len));
+    _expected_payload = (uint8_t *)bm_malloc(payload_len);
     memcpy(_expected_payload, payload, payload_len);
     _expected_payload_len = payload_len;
   }
@@ -62,7 +62,7 @@ BmErr bcmp_send_ping_request(uint64_t node_id, const void *addr, const uint8_t *
          echo_req->payload_len);
 
   BmErr rval =
-      bcmp_tx(addr, BcmpEchoRequestMessage, reinterpret_cast<uint8_t *>(echo_req), echo_len, 0);
+      bcmp_tx(addr, BcmpEchoRequestMessage, (uint8_t *)echo_req, echo_len, 0);
 
   _ping_request_time = bm_ticks_to_ms(bm_get_tick_count());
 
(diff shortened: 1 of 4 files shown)


src/lib/bcmp/bcmp_ping.cpp
```diff
@@ -20,23 +20,23 @@ static uint16_t _expected_payload_len = 0;
...
   if (payload == NULL) {
     payload_len = 0;
   }
 
-  uint16_t echo_len = sizeof(bcmp_echo_request_t) + payload_len;
+  uint16_t echo_len = sizeof(BcmpEchoRequest) + payload_len;
 
-  uint8_t *echo_req_buff = static_cast<uint8_t *>(pvPortMalloc(echo_len));
-  configASSERT(echo_req_buff);
+  uint8_t *echo_req_buff = static_cast<uint8_t *>(bm_malloc(echo_len));
```
matt001k: I think we should try to C-ify these files as much as possible seeing as all of bm_core will be C based. Could you change the `static_cast`/other C++ casts to C casts? (`uint8_t *`).

src/lib/bcmp/bcmp_ping.cpp
```diff
@@ -45,28 +45,28 @@ err_t bcmp_send_ping_request(uint64_t node_id, const ip_addr_t *addr, const uint
...
 
   printf("PING (%016" PRIx64 "): %" PRIu16 " data bytes\n", echo_req->target_node_id,
          echo_req->payload_len);
 
-  err_t rval =
+  BmErr rval =
       bcmp_tx(addr, BcmpEchoRequestMessage, reinterpret_cast<uint8_t *>(echo_req), echo_len, 0);
 
-  _ping_request_time = uptimeGetMicroSeconds();
+  _ping_request_time = bm_ticks_to_ms(bm_get_tick_count());
```
matt001k, line 67: Will this be a problem seeing as it is milliseconds?

src/lib/bcmp/bcmp_ping.cpp
```diff
@@ -113,10 +112,9 @@ err_t bcmp_process_ping_request(bcmp_echo_request_t *echo_req, const ip_addr_t *
   \param *dst - destination ip of request (used for responding to the correct multicast address)
   \
   */
-err_t bcmp_process_ping_reply(bcmp_echo_reply_t *echo_reply) {
-  configASSERT(echo_reply);
-
-  err_t rval = ERR_VAL;
+BmErr bcmp_process_ping_reply(BcmpProcessData data) {
```
matt001k: This can be made static

src/lib/bcmp/bcmp_ping.cpp
```diff
@@ -91,18 +91,17 @@ err_t bcmp_send_ping_reply(bcmp_echo_reply_t *echo_reply, const ip_addr_t *addr,
...
   \param *src - source ip of requester
   \param *dst - destination ip of request (used for responding to the correct multicast address)
-  \ret ERR_OK if successful
+  \ret BmOK if successful
 */
-err_t bcmp_process_ping_request(bcmp_echo_request_t *echo_req, const ip_addr_t *src,
-                                const ip_addr_t *dst, uint16_t seq_num) {
-  (void)src;
-  configASSERT(echo_req);
+BmErr bcmp_process_ping_request(BcmpProcessData data) {
```
matt001k: This can be static as well

matt001k (commented):
Nice work! Couple of review comments. Let me know your thoughts.

src/lib/bcmp/bcmp_ping.h
```diff
@@ -1,9 +1,9 @@
...
-#include "lwip/ip_addr.h"
 
-err_t bcmp_send_ping_request(uint64_t node_id, const ip_addr_t *addr, const uint8_t* payload, uint16_t payload_len);
-err_t bcmp_send_ping_reply(bcmp_echo_reply_t *echo_reply, const ip_addr_t *addr, uint16_t seq_num);
-err_t bcmp_process_ping_request(bcmp_echo_request_t *echo_req, const ip_addr_t *src, const ip_addr_t *dst, uint16_t seq_num);
-err_t bcmp_process_ping_reply(bcmp_echo_reply_t *echo_reply);
+BmErr bcmp_send_ping_request(uint64_t node_id, const void *addr, const uint8_t* payload, uint16_t payload_len);
+// BmErr bcmp_send_ping_reply(BcmpEchoReply *echo_reply, const void *addr, uint16_t seq_num);
+// BmErr bcmp_process_ping_request(BcmpProcessData data, const void *src, const void *dst, uint16_t seq_num);
+// BmErr bcmp_process_ping_reply(BcmpProcessData data);
```
matt001k: These also look like they can be removed.

src/lib/bcmp/bcmp_ping.cpp
victorsowa12: The printout is in milliseconds, and realistically this could be user defined in the future as right now the only way to get this is through CLI. So I don't think it's a problem to convert the time from micro to millis.

matt001k (approved):

PR #172, opened 2024-09-26
Finish LWIP Abstractions From BCMP

matt001k:
This finishes the abstractions necessary for bcmp source file. It does not rely on any lwip items now.


src/lib/bcmp/bcmp.cpp
```diff
@@ -204,35 +195,34 @@ BmErr bcmp_tx(const void *dst, BcmpMessageType type, uint8_t *data, uint16_t siz
     See section 5.4.4.2 of the Bristlemouth spec for details.
 
     \param[in] pbuf Packet buffer to forward.
-    \param[in] ingress_port Port on which the packet was received.
+    \param[in] ingress_port Porr on which the packet was received.
```
matt001k: Hmm weird change but ok...

src/lib/bcmp/bcmp_heartbeat.cpp
```diff
@@ -21,7 +18,7 @@ extern neighbor_discovery_callback_t
   \return ERR_OK if successful
 */
 BmErr bcmp_send_heartbeat(uint32_t lease_duration_s) {
-  BcmpHeartbeat heartbeat = {.time_since_boot_us = uptimeGetMicroSeconds(),
+  BcmpHeartbeat heartbeat = {.time_since_boot_us = bm_ticks_to_ms(bm_get_tick_count()),
```
victorsowa12: this one I technically time since boot microseconds.... not sure how hard that requirement is, but it is mentioned in the bcmp v1 google sheet. 

I'm up for using millis (maybe *100) for the time being and leaving a TODO for including a micro-seconds platform/bsp abstraction function.

src/lib/bcmp/bcmp.cpp
```diff
@@ -208,31 +199,30 @@ BmErr bcmp_tx(const void *dst, BcmpMessageType type, uint8_t *data, uint16_t siz
...
-  err = raw_sendto_if_src(CTX.pcb, pbuf, &port_specific_dst, CTX.netif, src) == ERR_OK
-            ? BmOK
-            : BmEBADMSG;
+  bm_ip_tx_perform(forward, &port_specific_dst);
   if (err != BmOK) {
     printf("Error forwarding BMCP packet link-locally: %d\n", err);
   }
-  pbuf_free(pbuf);
+  bm_ip_tx_cleanup(forward);
   return err;
```
victorsowa12: Looks like `err` is not being set anywhere in the new changes

matt001k: Will adjust

src/lib/bcmp/bcmp_heartbeat.cpp
matt001k: Agreed, oops meant to add it in there (*1000) but I had forgotten good catch.

victorsowa12 (approved):

PR #176, opened 2024-09-30
Port Over Resource Discovery

matt001k:
Ports over resouce discovery functionality to be compatible with bm_core and other accessibility repositories. This also improves some error handling of other modules!

The following is screen captures of the dev kit subscribing to topics on the bridge:
image

Receiving Data from subscribed topics:
image


src/lib/bcmp/bcmp.cpp
```diff
@@ -73,9 +73,7 @@ void bcmp_link_change(uint8_t port, bool state) {
 //  \return none
 //*/
 //static void dfu_copy_and_process_message(BcmpProcessData data) {
-//  bcmp_header_t *header = static_cast<bcmp_header_t *>(pbuf->payload);
-//  uint8_t *buf = static_cast<uint8_t *>(pvPortMalloc((pbuf->len) - sizeof(bcmp_header_t)));
-//  configASSERT(buf);
+//  bcmp_header_t *header = static_cast<bcmp_header_t *>(pbuf->payload); uint8_t *buf = static_cast<uint8_t *>(pvPortMalloc((pbuf->len) - sizeof(bcmp_header_t))); configASSERT(buf);
```
towynlin: Looks like a change that should be removed and left as-is.

src/lib/bcmp/bcmp.h
```diff
@@ -3,6 +3,9 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+extern "C" {
+#include "device.h"
+}
```
towynlin, line 8: Could you explain this? Looks like some tribal knowledge we'll have to remember, that some set of files need extern C while others don't.

matt001k: For the time being, until all of the messages files are c-ified, C++ files need to know how to link the current C source files. Since the whole library for `bm_core` will be in C, I figured we do not want to link "private" api to `bm_core` with the possibility of there being C++ files ex:
## Some 'bm_core` header file:
```c
#ifdef __cplusplus
extern "C" {
#endif

...

#ifdef __cplusplus
}
#endif
```

Once these files are brought into `bm_core`, the `extern "C"` directive can be removed.

src/lib/bcmp/bcmp_config.cpp
```diff
@@ -557,18 +556,17 @@ BmErr bcmp_config_init(Configuration *user_cfg, Configuration *sys_cfg) {
...
-  //TODO please handle this better
-  err = packet_add(&config_get, BcmpConfigGetMessage);
-  err = packet_add(&config_set, BcmpConfigSetMessage);
-  err = packet_add(&config_commit, BcmpConfigCommitMessage);
-  err = packet_add(&config_status_request, BcmpConfigStatusRequestMessage);
-  err = packet_add(&config_status_response, BcmpConfigStatusResponseMessage);
-  err = packet_add(&config_delete_request, BcmpConfigDeleteRequestMessage);
-  err = packet_add(&config_delete_response, BcmpConfigDeleteResponseMessage);
-  err = packet_add(&config_value, BcmpConfigValueMessage);
+  bm_err_check(err, packet_add(&config_get, BcmpConfigGetMessage));
```
towynlin, line 563: Nice! 👏 

Please queue up a separate task (not this PR) for documenting `bm_err_check`. People using it (including us) need to know (!) that it will fall through and never call the second argument if the first one isn't BmOK. It must only be used with functions that return a BmErr. Think about any common typos someone might make that would still compile, guard against them if possible, but at least document.

src/lib/bcmp/bcmp_info.cpp
```diff
@@ -26,48 +25,38 @@ static BCMP_Linked_List_Generic INFO_REQUEST_LIST;
 static BmErr bcmp_send_info(void *dst) {
 
   BmErr err = BmENOMEM;
-  char *ver_str = (char *)bm_malloc(ver_str_max_len);
+  uint8_t ver_str_len = strlen(vers_str());
```
towynlin: Can we improve naming here? I don't know how widely used these things are, but I'd prefer to rename `vers_str` to either `version_str` or `version_string`. If it's widely used and not a good idea to change, then let's align the naming of the assigned variable and the function. Right now the difference between "ver" and "vers" looks like a mistake.

src/lib/bcmp/bcmp.h
towynlin: Ah, got it. Thank you. Very clear. 👍

src/lib/bcmp/bcmp_config.cpp
matt001k: Task created here:

https://app.shortcut.com/sofar/story/207754/document-bm-err-check-usage-in-bristlemouth-integration-docs

src/lib/bcmp/bm/bm_util.h
```diff
@@ -23,8 +23,10 @@ extern const ip6_addr_t multicast_ll_addr;
 static inline uint64_t ip_to_nodeid(void *ip) {
   uint32_t msb = ((uint32_t *)(ip))[2];
   uint32_t lsb = ((uint32_t *)(ip))[3];
-  swap_32bit(&msb);
-  swap_32bit(&lsb);
+  if (is_little_endian()) {
+    swap_32bit(&msb);
+    swap_32bit(&lsb);
+  }
```
towynlin: It looks like this logic is incorrect in the original, before this PR. Please add unit tests to confirm/deny. Also the format of the incoming `void *ip` might be endian-dependent? Not sure.

In any case, I think that if we assume `msb` is `0x01234567` which would be represented in little-endian memory as `{ 0x67, 0x45, 0x23, 0x01 }`, and if `lsb` is `0x89abcdef` with LE representation `{ 0xef, 0xcd, 0xab, 0x89 }`, then after the swaps, `msb` is `0x67452301` and `lsb` is `0xefcdab89`, then we would return `0x67452301efcdab89` when what I think we actually want to return is `0xefcdab8967452301`.

Also nitpick: in MSB/LSB the B means byte/bit depending on context, but here we're dealing in 32-bit words, not bytes. I might suggest `high_word` and `low_word`?

src/lib/bcmp/bcmp_info.cpp
```diff
@@ -26,48 +25,34 @@ static BCMP_Linked_List_Generic INFO_REQUEST_LIST;
...
+  uint8_t *buf = (uint8_t *)bm_malloc(info_len);
 
-  if (ver_str) {
-    memset(ver_str, 0, ver_str_max_len);
+  if (buf) {
+    memset(buf, 0, info_len);
 
-    uint8_t ver_str_len =
-        snprintf(ver_str, ver_str_max_len, "%s@%s", APP_NAME, getFWVersionStr());
+    BcmpDeviceInfoReply *dev_info = (BcmpDeviceInfoReply *)buf;
```
towynlin: Rather than casting once to `uint8_t *` then casting again to `BcmpDeviceInfoReply *`, can we just cast the original alloc to `BcmpDeviceInfoReply *` from the beginning?

src/lib/bcmp/bcmp_neighbors.cpp
```diff
@@ -67,14 +66,19 @@ void bcmp_neighbor_foreach(neighbor_callback_t cb) {
...
-    printf("🏚  Neighbor offline :'( %016" PRIx64 "\n", neighbor->node_id);
-    if (NEIGHBOR_DISCOVERY_CB) {
-      NEIGHBOR_DISCOVERY_CB(false, neighbor);
+  int32_t time_remaining = 0;
+  int32_t time_start = (int32_t)neighbor->last_heartbeat_ticks;
+  int32_t time_end = 0;
+  if (neighbor->online) {
+    time_end = (int32_t)bm_ms_to_ticks(2 * neighbor->heartbeat_period_s * 1000);
+    time_remaining = time_start + time_end - (int32_t)bm_get_tick_count();
+    if (time_remaining <= 0) {
```
towynlin: This seems very easy to subtly mess up. Time rollover has been the source of multiple serious bugs for us. The casting to `int32_t` raises red flags for me. Tell me more about the rationale for changing it?

towynlin (commented):
Had some comments. Let's also strive to keep PRs smaller in the future. I'm running out of focus and not feeling confident in my review.

src/lib/bcmp/bm/bm_util.h
matt001k: Agreed on the naming!
The logic does appear to be correct tho.
The following is a screen capture of with the current logic vs what is in the develop branch as reported by the bridge:
## New Changes
Screenshot 2024-09-30 at 3 09 50 PM

## Develop
Screenshot 2024-09-30 at 3 13 28 PM

I belive there has to be a standardized way to represent our IP addresses across bristlemouth to be agnostic to the IP stack. There is a card to address this [here](https://app.shortcut.com/sofar/story/207759/create-standardized-ip-address-representation-across-bristlemouth-agnostic-of-ip-stack)

src/lib/bcmp/bcmp_neighbors.cpp
matt001k: This was a function that was tied to external API outside of bcmp/bristlemouth. I decided to carry over the logic to here. Maybe the better choice is to place it in `util.h` in `bm_common`?

matt001k: I adressed this by placing the API it in `bm_common` and keeping the function call logic the same.

victorsowa12 (approved):
I'll wait for this to be merged before making the time PR so I can rebase my branch and then use some of the additional abstractions like node_id()!

towynlin (approved):

PR #177, opened 2024-09-30
Experiment/bcmp time

victorsowa12:
Tried to include usage of new functions like node_id and bm_err_check. Lmk if I there are any other functions like those that I didn't use yet.

Here are some examples of setting and getting time from a bridge on "old" fw and a hello world mote on the "new" fw.

Bridge telling hello world mote to set its time:
Screenshot 2024-09-30 at 3 18 36 PM
Mote after having its time set from the bridge:
Screenshot 2024-09-30 at 3 18 46 PM
Mote getting the bridges time:
Screenshot 2024-09-30 at 3 19 09 PM
Bridge getting the motes time:
Screenshot 2024-09-30 at 4 28 36 PM
Mote setting the bridges time:
Screenshot 2024-09-30 at 4 29 59 PM
Bridge after mote set its time:
Screenshot 2024-09-30 at 4 29 51 PM

### DIFF
--- a/src/lib/bcmp/bcmp_time.cpp
+++ b/src/lib/bcmp/bcmp_time.cpp
@@ -9,8 +9,17 @@ extern "C" {
 #include "packet.h"
 }
 
-bool bcmp_time_set_time(uint64_t target_node_id, uint64_t utc_us) {
-  bool ret = true;
+/*!
+ @brief Sets the system time on a target node.
+
+ This function sends a system time set message to the specified target node.
+
+ @param[in] target_node_id The ID of the target node.
+ @param[in] utc_us The UTC time in microseconds to set on the target node.
+ @return BmOK if the message was sent successfully, BmEINVAL otherwise.
+ */
+BmErr bcmp_time_set_time(uint64_t target_node_id, uint64_t utc_us) {
+  BmErr err = BmOK;
   uint64_t source_node_id = getNodeId();
   BcmpSystemTimeSet set_msg;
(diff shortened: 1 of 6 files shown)


src/lib/bcmp/bcmp_time.cpp
```diff
@@ -118,5 +168,31 @@ bool bcmp_time_process_time_message(bcmp_message_type_t bcmp_msg_type, uint8_t *
     }
   } while (0);
 
-  return should_forward;
+  if (should_forward) {
+    bcmp_ll_forward(data.header, data.payload, data.size, data.ingress_port);
```
matt001k: Could we assign `err = bcmp_ll_forward(...)` here?

src/lib/bcmp/bcmp_time.cpp
```diff
@@ -2,54 +2,89 @@
 #include "app_util.h"
 #include "bcmp.h"
 #include "bm_util.h"
-#include "device_info.h"
 #include "stm32_rtc.h"
```
matt001k, line 5: Is there still a dependency on this?

victorsowa12: Yeah I put some to-do's to abstract the RTC code away. That should probably be in bm_common right? something like a `bm_rtc.h` or `bm_time.h` file probably?

matt001k: That sounds good! `bm_time` fits more I feel like.

matt001k (approved):

PR #182, opened 2024-10-05
more fully featured bristleback serial_payload_example app

evanShap:
Modifications to support PME dev, can run MicrDOT with the default example apps:
- copied relevant code from BMDK serial_payload_example to BB version for more full PLUART features and configs.
- enable 5V output on BMDK version (this has tripped up at least one forum user already too).
- fix a logging bug in BMDK version.

### DIFF
--- a/src/apps/bristleback_apps/serial_payload_example/user_code/user_code.cpp
+++ b/src/apps/bristleback_apps/serial_payload_example/user_code/user_code.cpp
@@ -12,22 +12,48 @@
 #include "usart.h"
 #include "util.h"
 
-#define LED_ON_TIME_MS 20
-#define LED_PERIOD_MS 1000
+#define LED_ON_TIME_MS 75
+#define LED_PERIOD_MS 5000
+#define DEFAULT_BAUD_RATE 9600
+#define DEFAULT_LINE_TERM 10 // newline, '\n', 0x0A
+#define BYTES_CLUSTER_MS 50  // used for console printing convenience
+#define DEFAULT_UART_MODE MODE_RS232
 
-bool led_state = false;
+// app_main passes a handle to the config partitions in NVM.
+extern cfg::Configuration *userConfigurationPartition;
+extern cfg::Configuration *systemConfigurationPartition;
 
-// A buffer to put data from out payload sensor into.
+// variables to store configurations retrieved from NVM
+static u_int32_t baud_rate_config = DEFAULT_BAUD_RATE;
+static u_int32_t line_term_config = DEFAULT_LINE_TERM;
+static u_int32_t bm_log_enable = false;
+
+// A buffer to put Rx data from our payload sensor into.
 char payload_buffer[2048];
+
+// A timer variable we can set to trigger a pulse on LED2 when we get payload serial data
+static int32_t ledLinePulse = -1;
 
 void setup(void) {
   /* USER ONE-TIME SETUP CODE GOES HERE */
+  // Retrieve user-set config values out of NVM.
+  userConfigurationPartition->getConfig("plUartBaudRate", strlen("plUartBaudRate"), baud_rate_config);
+  userConfigurationPartition->getConfig("plUartLineTerm", strlen("plUartLineTerm"), line_term_config);
+  systemConfigurationPartition->getConfig("sensorBmLogEnable", strlen("sensorBmLogEnable"), bm_log_enable);
+
   // Setup the UART – the on-board serial driver that talks to the RS232 transceiver.
   PLUART::init(USER_TASK_PRIORITY);
   // Baud set per expected baud rate of the sensor.
-  PLUART::setBaud(9600);
+  PLUART::setBaud(baud_rate_config);
+  // Enable passing raw bytes to user app.
+  PLUART::setUseByteStreamBuffer(true);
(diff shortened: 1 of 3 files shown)


src/apps/bristleback_apps/serial_payload_example/user_code/user_code.cpp
```diff
@@ -36,44 +62,94 @@ void setup(void) {
...
+  // Note - PLUART::setUseByteStreamBuffer must be set true in setup to enable bytes.
+  if (readingBytesTimer == -1 && PLUART::byteAvailable()) {
+    // Get the RTC if available
+    RTCTimeAndDate_t time_and_date = {};
+    rtcGet(&time_and_date);
+    char rtcTimeBuffer[32];
+    rtcPrint(rtcTimeBuffer, &time_and_date);
+    printf("[payload-bytes] | tick: %llu, rtc: %s, bytes:", uptimeGetMs(), rtcTimeBuffer);
+    // not very readable, but it's a compact trick to overload our timer variable with a -1 flag
+    readingBytesTimer = (int64_t)((u_int32_t)uptimeGetMs());
```
victorsowa12, line 86: why not make the `readingBytesTimer` a uint64_t, initialize it to 0 and make all the compares against 0? That would remove the need for the casting magic.

Is there an edge case with 0 I'm not seeing?

victorsowa12 (approved):
Got a question about using an int for the timer, but otherwise looks good!

src/apps/bristleback_apps/serial_payload_example/user_code/user_code.cpp
```diff
@@ -36,44 +62,94 @@ void setup(void) {
...
+  if (!led1State && ((u_int32_t)uptimeGetMs() - ledPulseTimer >= LED_PERIOD_MS)) {
+    IOWrite(&LED_GREEN, 1);
+    ledOnTimer = uptimeGetMs();
+    ledPulseTimer += LED_PERIOD_MS;
+    led1State = true;
+  }
+  // If LED1 has been on for LED_ON_TIME_MS milliseconds, turn it off.
+  else if (led1State && ((u_int32_t)uptimeGetMs() - ledOnTimer >= LED_ON_TIME_MS)) {
+    IOWrite(&LED_GREEN, 0);
+    led1State = false;
```
matt001k, line 153: Could these and maybe some other chunks of code be broken down into functions (even inline functions), just to add more clarity, and improve readability?
If anything the auxiliary functions like blinking LEDs might help reduce some of the eye clutter.
i.e.
```
payload_print_to_file(...);

receive_rx_led_blink_blue();

heartbeat_led_green()
```

src/apps/bristleback_apps/serial_payload_example/user_code/user_code.cpp
matt001k: Seconded, it took me a second to realize what was going on. I think it would make it more readable if it was kept at a `uint64_t`. That way we also would risk truncating `uptimeGetMs` (although very unlikely to ever happen).

towynlin: Thirded 👍 A related improvement would be changing the name of `readingBytesTimer` (since it's a number not a timer) to something like `firstByteUptimeMs` or (new bm_core style) `first_byte_uptime_ms`.

Also (nit) I'd love to get rid of the non-standard types with the extra underscore. We should use `uint64_t` instead of `u_int64_t`.

evanShap: @towynlin @victorsowa12 @matt001k - so there's actually an evil quirk with `uptimeGetMs` (our function, our fault) that bit us during some very early stm32 work - I know Russ was in the weeds on that one, but I don't know if you remember it Victor? If RTC is not set, the return value will actually rollover at uint32_max. If RTC is set, it will rollover at uint64_max.

So the very slim edge case is a rollover at ~32 days of runtime falsely triggering a new byte cluster while we're already in the middle of one. 🤷

evanShap:
@towynlin @matt001k @victorsowa12 - I have no objections to any of the proposed improvements / changes, but this is all just straight copied over from the BMDK version of this app. Would prefer to save improvements for an eventual cleanup of the overall serial payload examples.

Going to merge this one in now as-is for PME support.

towynlin:
Makes perfect sense. I queued up a backlog story for the improvements: 208149.

src/apps/bristleback_apps/serial_payload_example/user_code/user_code.cpp
victorsowa12: Sounds familiar. Adding this for context for the SC-208149 ticket. Took a quick look through and I see that `uptimeGetMs` refers to FreeRTOS ticks when there is no RTC. FreeRTOS ticks are stored as a `uint32_t`, so it will rollover.

I think in the context of the system you would need three things to get this rollover:
1. no gps signal for 32 days (or an RTC-0 blocking node ahead of you in the chain)
2. power controller enabled continuously
3. power controller tick timing enabled

I don't know the use case for PME so I don't know if this is a likely scenario. 🤷‍♂️

PR #184, opened 2024-10-10
Ports L2 Module To bm_core

matt001k:
Refactors code to utilize bm_core API (bm_os, etc.)
Abstracts lwip elements from l2 file and provides a generic API that can be used across multiple network stacks

### DIFF
--- a/src/lib/bcmp/bm/bm_l2.cpp
+++ b/src/lib/bcmp/bm/bm_l2.cpp
@@ -1,524 +1,510 @@
-#include "FreeRTOS.h"
-#include "bm_config.h"
 #include "bm_l2.h"
+#include "bm_config.h"
 #include "eth_adin2111.h"
 #include "lwip/ethip6.h"
 #include "lwip/prot/ethernet.h"
 #include "lwip/snmp.h"
 #include "task_priorities.h"
+extern "C" {
+#include "bm_ip.h"
+#include "bm_os.h"
+#include "util.h"
+}
 
-#define IFNAME0                     'b'
-#define IFNAME1                     'm'
-#define NETIF_LINK_SPEED_BPS        10000000
-#define ETHERNET_MTU                1500
+//TODO: this will have to be moved as we go along
+extern struct netif netif;
 
-#define ADD_EGRESS_PORT(addr, port) (addr[sizeof(struct eth_hdr) + offsetof(struct ip6_hdr, src) + EGRESS_PORT_IDX] = port)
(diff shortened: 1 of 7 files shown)


matt001k:
I have not moved the L2 file yet to bm_core. I plan on opening another PR after this just to migrate it over. That way y'all can see the diff of what I changed in this PR.

src/lib/bcmp/bm/bm_l2.cpp
```diff
@@ -1,524 +1,510 @@
...
 #include "eth_adin2111.h"
 #include "lwip/ethip6.h"
 #include "lwip/prot/ethernet.h"
 #include "lwip/snmp.h"
 #include "task_priorities.h"
+extern "C" {
+#include "bm_ip.h"
+#include "bm_os.h"
+#include "util.h"
+}
```
matt001k, line 9: These will go away when migrated to bm_core repo.

src/lib/bcmp/bm/bristlemouth.h
```diff
@@ -1,21 +1,8 @@
 #pragma once
 
+#include "serial.h"
```
victorsowa12: why is serial.h needed here?

src/lib/bcmp/bm/bristlemouth.cpp
```diff
@@ -126,6 +82,8 @@ void bcl_init(void) {
   \return pointer to string with ip address
 */
 const char *bcl_get_ip_str(uint8_t idx) {
-  configASSERT(idx < LWIP_IPV6_NUM_ADDRESSES);
-  return (ip6addr_ntoa(netif_ip6_addr(&netif, idx)));
+  (void)idx;
+  //configASSERT(idx < LWIP_IPV6_NUM_ADDRESSES);
+  //return (ip6addr_ntoa(netif_ip6_addr(&netif, idx)));
+  return "none\n";
```
victorsowa12: Is it possible to also move this over to the IP abstraction? I think it is only used by the CLI command `info` so I don't think it really needs to be a part of bcl.

src/lib/bcmp/bm/bristlemouth.h
matt001k: Oops, I really have no idea how this got added :(

src/lib/bcmp/bm/bristlemouth.cpp
matt001k: Ahhh yes I meant to do that, thanks for the catch.

victorsowa12 (approved):
🤘

PR #185, opened 2024-10-10
Experiment/use bm os in dfu

victorsowa12:
Convert bm_dfu files from freertos to bm_os and use the newly brought over crc + timer_callback_handler files form bm_common.

Need to wait for https://github.com/bristlemouth/bm_core/pull/14 and https://github.com/bristlemouth/bm_common/pull/4 so I can make sure they are pointing to the right sha's etc.


src/lib/bcmp/dfu/bm_dfu_client.cpp
```diff
@@ -307,7 +303,8 @@ void s_client_receiving_entry(void) {
     bm_dfu_req_next_chunk(client_ctx.host_node_id, client_ctx.current_chunk);
 
     /* Kickoff Chunk timeout */
-    configASSERT(xTimerStart(client_ctx.chunk_timer, 10));
+    // configASSERT(xTimerStart(client_ctx.chunk_timer, 10));
+    bm_timer_start(client_ctx.chunk_timer, 10);
```
victorsowa12, line 310: I left most of the config asserts in for now because these take place inside of states within the statemachine and the error cannot be returned up to the statemachine (as it exists right now).

But I was thinking perhaps in the future we could utilize a similar thing to the `bm_err_check` macro for asserts where we assert if the user has set some flag for using asserts, or just print an error message. Not super urgent, but I thought leaving in the asserts might be a good reminder to do something.

src/lib/bcmp/bcmp_time.cpp
```diff
@@ -104,12 +103,20 @@ static void bcmp_time_process_time_request_msg(const BcmpSystemTimeRequest *msg)
...
+                         .day = datetime.day,
+                         .hour = datetime.hour,
+                         .minute = datetime.min,
+                         .second = datetime.sec,
+=======
   RtcTimeAndDate time = {// TODO: Consolidate the time functions into util.h
                          .year = datetime.year,       .month = datetime.month,
                          .day = datetime.day,         .hour = datetime.hour,
                          .minute = datetime.min,      .second = datetime.sec,
+>>>>>>> experiment/bm_core
```
matt001k: Looks like some conflicts sneaking through here

src/lib/bcmp/dfu/bm_dfu_client.cpp
matt001k: Is it worthwhile to place a TODO or something at the top of the header file or this file?

src/lib/common/topology_sampler.cpp
```diff
@@ -3,7 +3,9 @@
 #include "semphr.h"
 #include "task.h"
 #include "task_priorities.h"
+extern "C" {
 #include "timer_callback_handler.h"
+}
```
matt001k: This feels off-putting to put in the top of so many files, I get why we were doing it for BCMP stuff because all of bristlemouth stack has not been c-ified, but it feels awkward here, should we place this directive in the `timer_callback_handler.h` file? @victorsowa12  @towynlin ??

matt001k (dismissed):

src/lib/bcmp/dfu/bm_dfu_client.cpp
victorsowa12: Sure I will add that!

src/lib/bcmp/bcmp_time.cpp
victorsowa12: woops forgot to commit that one!

src/lib/common/topology_sampler.cpp
victorsowa12: Either one is fine with me. I'll move it over to the header file.

src/apps/bm_devkit/hello_world/CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ set(BRISTLEMOUTH_FILES
     ${SRC_DIR}/lib/bm_core/src/platform/bm_lwip.c
     ${CMAKE_BINARY_DIR}/_deps/bm_common-src/src/platform/bm_freertos.c
     ${CMAKE_BINARY_DIR}/_deps/bm_common-src/src/lib_state_machine.c
+    ${CMAKE_BINARY_DIR}/_deps/bm_common-src/src/timer_callback_handler.c
+    ${CMAKE_BINARY_DIR}/_deps/bm_common-src/third_party/crc/crc16.c
+    ${CMAKE_BINARY_DIR}/_deps/bm_common-src/third_party/crc/crc32.c
```
matt001k: Do you think you can add these to the bm,_common directory so it can be built independently?

src/lib/bcmp/bcmp.cpp
matt001k, line 1: This has been moved to bm_core.

victorsowa12: oh funny, I must've re-included them when merging experiment/bm_core back into this branch. I'll remove this, bcmp_resource_discovery and bcmp_topology

matt001k (approved):

PR #187, opened 2024-10-11
Prevent Watchdog From Tripping At First Aligned Interval Reading In Power Controller

matt001k:
Well it looks like out time is up (pun intended).
The epoch time had gotten too large that when we started to align interval start times for the power controller starting from 0 in variable `lastIntervalStartS`, we began tripping the wdt.
This probably should get released ASAP, because there is a potential that many units in the field would be experiencing cyclic WDT resets on boot (I saw we have an upcoming release in our iteration).

### DIFF
--- a/src/apps/bridge/bridgePowerController.cpp
+++ b/src/apps/bridge/bridgePowerController.cpp
@@ -362,6 +362,10 @@ uint32_t BridgePowerController::getCurrentTimeS() {
 uint32_t BridgePowerController::_alignNextInterval(uint32_t nowEpochS,
                                                    uint32_t lastIntervalStartS,
                                                    uint32_t sampleIntervalS) {
+  if (!lastIntervalStartS) {
+    // Prevent many loops from occurring and tripping watchdog
+    lastIntervalStartS = nowEpochS - (nowEpochS % sampleIntervalS);
+  }
   uint32_t alignedEpoch = lastIntervalStartS + sampleIntervalS;
   while (alignedEpoch < nowEpochS) {
     // If the aligned epoch is in the past, the timebase must have just jumped forward.


src/apps/bridge/bridgePowerController.cpp
```diff
@@ -362,7 +362,12 @@ uint32_t BridgePowerController::getCurrentTimeS() {
 uint32_t BridgePowerController::_alignNextInterval(uint32_t nowEpochS,
                                                    uint32_t lastIntervalStartS,
                                                    uint32_t sampleIntervalS) {
-  uint32_t alignedEpoch = lastIntervalStartS + sampleIntervalS;
+  uint32_t alignedEpoch = 0;
```
towynlin: Minor thing but could you remove the alignedEpoch = 0 line and add the uint32_t at the beginning of line 370 below. Since alignedEpoch is only needed on 371 and not within the if clause, that would make this PR is only the addition of the check for lastIntervalStartS == 0.

towynlin:
I'm leaving my board stack plugged in for a while, but I haven't observed this on it yet, thankfully.

The while loop was introduced in https://github.com/bristlemouth/bm_protocol/pull/73 which was first released in v0.10.0. So if this does turn out to be a wider issue, that's the impacted group — smart moorings on v0.10.0 or later.

towynlin:
As discussed IRL, number of while loop iterations depends on sample interval. Shorter sample intervals lead to more time spinning in that loop.

victorsowa12:
I'm trying to reproduce aswell and not hitting it. What are the configs on your bridge? The sampleIntervalMs should be what is increasing/decreasing the length of the loop.

matt001k:
@towynlin the default value for this is 5 minutes. This bug is dependent on the `sampleIntervalMs`. Before merging I will see if there is a way to get rid of the while loop here to mitigate this issue (now that we know it is not hair on fire and we have a way to fix it if it appears).

victorsowa12:
I'm think the default value for sample interval is 30 minutes, unless we set it differently during production.

towynlin (approved):

victorsowa12 (approved):

PR #188, opened 2024-10-12
CMake Function Integration

matt001k:
This is to go along with https://github.com/bristlemouth/bm_core/pull/15


src/apps/bm_devkit/bmdk_common/bm_config.h
```diff
@@ -1,42 +1,24 @@
...
-    void * config;
-} bm_netdev_config_t;
-
-/* Define for actual netdev instance count here, as some of the later code currently iterates devices
-   using BM_NETDEV_MAX_DEVICES (which specifies the number of enum entries, not the actual number of devices).
-   If you added a new enum entry, but didn't change the instance array, the current code would crash. */
-#define BM_NETDEV_COUNT (2)
-
-extern const bm_netdev_config_t bm_netdev_config[BM_NETDEV_COUNT];
-
```
matt001k, line 39: These moved to bm_l2, I believe they will fit there better with @towynlin work on the netif.
I will be updating some of the L2 code to pass in the netif we plan to use rather than having it hardcoded in the code. (Next PR).

victorsowa12 (approved):

PR #190, opened 2024-10-14
Pulls Out Many ADIN Related Items

matt001k:
Allows the l2 layer to standalone without ADIN related code
Adds new API to get number of configured devices and replaces items that utilized the get_port API
Formatting changes

### DIFF
--- a/src/apps/bridge/bridgePowerController.cpp
+++ b/src/apps/bridge/bridgePowerController.cpp
@@ -1,6 +1,7 @@
 #include "bridgePowerController.h"
 #include "FreeRTOS.h"
 #include "app_pub_sub.h"
+#include "app_util.h"
 #include "bm_l2.h"
 #include "bm_serial.h"
 #include "bridgeLog.h"
@@ -9,7 +10,6 @@
 #include "stm32_rtc.h"
 #include "task.h"
 #include "task_priorities.h"
-#include "app_util.h"
 #include <cinttypes>
 #include <stdio.h>
 #ifdef RAW_PRESSURE_ENABLE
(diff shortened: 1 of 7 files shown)


src/apps/bridge/bridgePowerController.cpp
```diff
@@ -312,13 +312,12 @@ void BridgePowerController::powerControllerRun(void *arg) {
 
 bool BridgePowerController::getAdinDevice() {
   bool rval = false;
-  for (uint32_t port = 0; port < bm_l2_get_num_ports(); port++) {
+  for (uint32_t port = 0; port < bm_l2_get_num_devices(); port++) {
```
matt001k: This API had to be changed because it was not accurrate, we are getting the handle for a device not the handle for a specific port.

matt001k:
The main goal here was to remove all ADIN related items, so that a network device configuration can be called in as a function parameter in the L2 layer. This makes the module more scalable and fits our scheme of how we actually want to initialize bristlemouth.

src/apps/bridge/bridgePowerController.cpp
victorsowa12: 👍, just putting down comment from meeting to switch the "port" variable to "device" or something more accurate

victorsowa12 (approved):

src/apps/bridge/bridgePowerController.cpp
matt001k: Changed!

matt001k:
@towynlin any reservations on this PR?

towynlin (approved):
Super quick scan review, but looks good to me! 👍

PR #191, opened 2024-10-16
Abstracts LWIP API From Bristlemouth Middleware

matt001k:
Adds further API to bm_ip.h in order to remove lwip dependencies
Replaces FreeRTOS calls with bm_os calls.


src/lib/middleware/bm_pubsub.h
```diff
@@ -1,30 +1,31 @@
 #pragma once
 
+#include "bm_common_pub_sub.h"
 #include <stdbool.h>
 #include <stdint.h>
-#include "lwip/netif.h"
-#include "lwip/pbuf.h"
-#include "lwip/udp.h"
```
matt001k, line 7: Only non-formatting changes here.

src/lib/middleware/bm_service.cpp
matt001k, line 1: Formatting only changes, can revert to clean this up and place it in next PR when this file get refactored.

src/lib/middleware/middleware.cpp
```diff
@@ -1,191 +1,155 @@
...
-#include "util.h"
+#include "bcmp.h"
+#include "bm_ip.h"
+#include "bm_os.h"
 }
 #include <string.h>
 
-#define NET_QUEUE_LEN 64
-
-#define MAX_PAYLOAD_LEN (1500 - sizeof(struct ip6_hdr) - sizeof(struct udp_hdr))
```
matt001k, line 20: This is partially defined in `bcmp.h`. So this change utilized the macro there.

src/lib/middleware/middleware.cpp
```diff
@@ -1,191 +1,155 @@
...
+#define middleware_task_size 512
+#define net_queue_len 64
+#define udp_header_size 8
```
matt001k, line 13: UDP header size is always 8 bytes

PR #194, opened 2024-10-22
Adds Load Cell Configuration Parameters

matt001k:
- Changes where load cell is initialized at (app_main) due to the need of having configuration parameter class object available
- Will only initialize load cell on a loadcell app build
- Adds configuration parameters for calibration factor and zero offset. Default values at 800.0 and 80000 respectively.

Screenshot of setting the calibration factor:
image

Screenshot of setting the zero offset:
image

Screenshot of getting the calibration factor after saving:
image

Screenshot of getting the zero offset after saving:
image

Below is a snippet of the load cell running with some "weight" (me squishing the loadcell) with configuration values:
```
Load cell sample called
adc0 - adc2:            0               123             84
adc0 - adc2:            0               122             191
adc0 - adc2:            0               120             130
adc0 - adc2:            0               120             77
adc0 - adc2:            0               120             108
adc0 - adc2:            0               118             201
adc0 - adc2:            0               121             20
adc0 - adc2:            0               125             105
adc0 - adc2:            0               124             114
2303285 | reading: 31572
byte zero 0
byte one 0
byte two 0
2303287 | weight: 26.158001
2303287 | calFactor: 1000.000000
2303287 | zeroOffset: 5000

```

### DIFF
--- a/src/apps/bm_devkit/bm_loadcell/user_code/user_code.cpp
+++ b/src/apps/bm_devkit/bm_loadcell/user_code/user_code.cpp
@@ -4,9 +4,10 @@
 #include "bm_pubsub.h"
 #include "bristlefin.h"
 #include "bsp.h"
+#include "configuration.h"
 #include "debug.h"
+#include "loadCellSampler.h"
 #include "lwip/inet.h"
-#include "nau7802.h"
 #include "payload_uart.h"
 #include "sensors.h"
 #include "stm32_rtc.h"
@@ -14,16 +15,15 @@
 #include "uptime.h"
 #include "usart.h"
 #include "util.h"
+#include <string.h>
 
 #define LED_ON_TIME_MS 20
 #define LED_PERIOD_MS 1000
 
-void loadCellSamplerInit(NAU7802 *sensor);
-
-NAU7802 loadCell(&i2c1, NAU7802_ADDR);
-
 // A timer variable we can set to trigger a pulse on LED2 when we get payload serial data
 static int32_t ledLinePulse = -1;
+static NAU7802 load_cell(&i2c1, NAU7802_ADDR);
+extern cfg::Configuration *systemConfigurationPartition;
 // This function is called from the payload UART library in src/lib/common/payload_uart.cpp::processLine function.
 //  Every time the uart receives the configured termination character ('\0' character by default),
 //  It will:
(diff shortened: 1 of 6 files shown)


src/apps/bm_devkit/bm_loadcell/user_code/user_code.cpp
```diff
@@ -79,16 +72,14 @@ void loop(void) {
...
+  if (!led1State && ((u_int32_t)uptimeGetMs() - ledPulseTimer >= LED_PERIOD_MS)) {
     bristlefin.setLed(1, Bristlefin::LED_GREEN);
     ledOnTimer = uptimeGetMs();
     ledPulseTimer += LED_PERIOD_MS;
     led1State = true;
   }
   // If LED1 has been on for LED_ON_TIME_MS milliseconds, turn it off.
-  else if (led1State &&
-           ((u_int32_t)uptimeGetMs() - ledOnTimer >= LED_ON_TIME_MS)) {
+  else if (led1State && ((u_int32_t)uptimeGetMs() - ledOnTimer >= LED_ON_TIME_MS)) {
```
matt001k, line 104: Formatting changes

src/apps/bm_devkit/bmdk_common/app_main.cpp
```diff
@@ -395,6 +394,27 @@ static void defaultTask(void *parameters) {
...
+                                         load_cell_cfg.calibration_factor);
+  }
+  if (!debug_configuration_system.getConfig("loadCellZeroOffset", strlen("loadCellZeroOffset"),
+                                            load_cell_cfg.zero_offset)) {
+    debug_configuration_system.setConfig("loadCellZeroOffset", strlen("loadCellZeroOffset"),
+                                         load_cell_cfg.zero_offset);
+  }
+  // Adds the load cell as a sensor for periodic sampling.
+  loadCellSamplerInit(&load_cell, load_cell_cfg);
+#endif
```
matt001k: Load cell is now initialized here rather than in user code, this will only occur if the app has a project definition for `LOADCELL`

src/apps/bm_devkit/bmdk_common/app_main.cpp
```diff
@@ -395,6 +394,27 @@ static void defaultTask(void *parameters) {
...
```
matt001k: Load cell is now initialized here based on project define.

matt001k:
AAddressing app_failure now check failure now :)

victorsowa12 (commented):
I would say the current trend is to just use the globally defined configuration partition from app_main `cfg::Configuration *systemConfigurationPartition`  in the code that needs the config partition(the sys one may need to be added to the bmdk main). This isn't the best way to do it, but I think we should keep doing that until we have a good method to get the config partition without needing the global. I think that would help to minimize app_main init code, keep the load cell init in the load cell app specific code, prevent another thing being wrapped by a #define, and keep the config usage examples somewhat consistent for people that are looking to use configs.

matt001k:
@victorsowa12 agreed I like this approach better.

towynlin (approved):

src/apps/bristleback_apps/loadcell/CMakeLists.txt
```diff
@@ -273,6 +273,7 @@ set(APP_LIBS
 set(APP_DEFINES
     APP_NAME="${APP_NAME}"
     CFG_TUSB_MCU=OPT_MCU_STM32U5
+    LOADCELL=1
```
victorsowa12: I believe this is no longer needed

victorsowa12 (approved):
👍👍👍

PR #195, opened 2024-10-23
Uma/sdi12 sonde

uma-kat:
SDI12 protocol for EXO3s sonde


src/apps/bm_devkit/bmdk_common/app_main.cpp
```diff
@@ -396,7 +396,7 @@ static void defaultTask(void *parameters) {
                                  .sensorsPollIntervalMs = sys_cfg_sensorsPollIntervalMs};
   sensorSamplerInit(&sensorConfig);
   // must call sensorsInit after sensorSamplerInit
-  sensorsInit();
+//  sensorsInit();
```
victorsowa12: Should uncomment this. We don't want to turn off the sensors for all the other apps. Unless this is added somewhere else, which I am not currently finding.
```suggestion
sensorsInit();
```

src/apps/bm_devkit/sdi12_example/exo3_sdi12_decoder.py
```diff
@@ -0,0 +1,128 @@
...
+        # Read the next 11 bytes representing detection data.
+        detect_data = bitstream.read('bytes:40')
+
+        # Convert detection data from bytes to a structured format.
+        detection_data = hex_to_struct(detect_data, detect_struct_description)
+
+        # Print the unpacked detection data.
+        print(f"- Detection data:")
+        for key in detection_data:
+            print(f"\t{key}: {detection_data[key]}")
```
victorsowa12: When I ran the script I noticed that it printed a bunch of decimal places.
Screenshot 2024-11-11 at 11 25 51 AM

Is it valuable at all to limit to the same number of sig figs as what was sent by the sensor?

src/apps/bm_devkit/sdi12_example/exo3_sdi12_decoder.py
```diff
@@ -0,0 +1,128 @@
+from bitstring import BitStream
```
victorsowa12, line 1: I noticed that my bristlemouth conda env didn't have `bitstring` in it, and I needed to activate another environment to run this script. Could you add this to the environment.yml please?
```
 - pip:
      - bitstring==4.2
```

src/lib/common/payload_uart.cpp
```diff
@@ -249,6 +249,49 @@ void setBaud(uint32_t new_baud_rate) {
                         LL_LPUART_PRESCALER_DIV64, new_baud_rate);
 }
 
+void setEvenParity(void){
+  LL_LPUART_Disable(static_cast<USART_TypeDef *>(uart_handle.device));
+  LL_LPUART_SetParity(static_cast<USART_TypeDef *>(uart_handle.device),
+                      LL_LPUART_PARITY_EVEN);
+  LL_LPUART_Enable(static_cast<USART_TypeDef *>(uart_handle.device));
+}
```
victorsowa12, line 257: 🤘

src/apps/bm_devkit/sdi12_example/user_code/user_code.h
```diff
@@ -0,0 +1,9 @@
+#include "serial.h"
+#include "bsp.h"
+#include "debug_configuration.h"
+
+#pragma once
+
+void setup(void);
+void loop(void);
+void transmit_samples(void);
```
victorsowa12: if this is only called in `user_code.cpp` you don't have to define it here. You can just define it in `user_code.cpp` near the top (which it already is) and make it `static void transmit_samples(void)` so we know it is only called from within the .cpp file.

src/apps/bm_devkit/sdi12_example/user_code/exo3s_sensor.cpp
```diff
@@ -0,0 +1,256 @@
...
+      rxBuffer[i] = rxBuffer[i] & 0x7F;
+    }
+    return true;
+  }
+  return false;
+}
+
+
+void SondeEXO3sSensor::sdi_cmd(int cmd) {
+  char result = sdiSuccess;
```
victorsowa12: `sdiSuccess` is #defined to 0 and later this `result` variable is set by `result = sdi_receive();` which returns a bool where true = success. This is a bit confusing. I think it would be clearer to just have result be a bool the entire time and change this line to:
```
bool result = false;
```
thus getting rid of the use of `sdiSuccess`.

src/apps/bm_devkit/sdi12_example/user_code/exo3s_sensor.h
```diff
@@ -0,0 +1,103 @@
...
+#define sdiFail 1
+#define sdiTimeout 2		// timeout condition
+#define sdiParityError 3 	// parity error in reply
+#define sdiReplyError 4		//
+#define breakTimeMin 15		// spec is 12 but a bit extra does no harm
+#define markTimeMin 9		// actually 8.33 but I am not OCD
+#define maxSensorResponseTime 15	// time after the command is sent
+#define maxTotalResponseTime 900	// time for the longest response which could be expectd
+#define characterDelay 2	// maximum time to wait for the next character before a fault flagged
+#define delayAfterTransmit 2	// a short delay put in after the last character is sent to allow for switching
```
victorsowa12: Besides the `breakTimeMin` and `markTimeMin` none of these are used. I think it may be valuable to leave them with a comment saying: "these are not used, but are part of the SDI-12 spec and are here for future improvements" or something along those lines.

src/apps/bm_devkit/sdi12_example/user_code/exo3s_sensor.h
```diff
@@ -0,0 +1,103 @@
...
+#include <stdint.h>
+#include "OrderedSeparatorLineParser.h"
+#include "Exo3LineParser.h"
+
+#define sizeBuffer 256
+#define sdiSuccess 0
+#define sdiFail 1
+#define sdiTimeout 2		// timeout condition
+#define sdiParityError 3 	// parity error in reply
+#define sdiReplyError 4		//
```
victorsowa12: Besides `sizeBuffer` these #defines are not used, so I think it would be best to remove them unless we want to use them to improve the error reporting.

(sdiSuccess is used, but I think it can replaced by just using a bool, see my other comment on that one)

src/apps/bm_devkit/sdi12_example/user_code/exo3s_sensor.cpp
victorsowa12: Or alternatively we could have `sdi_recieve();` return an `sdiErr` enum variable like:
```
typedef enum {
sdiSuccess,
sdiFail,
sdiParityError,
} sdiErr;
```

src/lib/bristlefin/bristlefin.cpp
```diff
@@ -85,7 +93,6 @@ void Bristlefin::disable3V() {
   IOWrite(&BF_IMU_RST, 0); // https://github.com/wavespotter/bristlemouth/issues/422 - Drive IMU RST low to avoid consuming xtra power.
   saved_mux_current_channel_ = getMuxChannel();
   setMuxChannel(TCA::CH_NONE); // Lowest power state for I2C mux
-  IOWrite(&BF_3V3_EN, 0); // 1 enables, 0 disables. Needed for I2C and I/O control.
```
victorsowa12: I think this should not be deleted?

victorsowa12 (commented):
Overall looks really good 😎! Left a couple comments

src/apps/bm_devkit/bmdk_common/app_main.cpp
evanShap: ➕

src/apps/bm_devkit/sdi12_example/CMakeLists.txt
```diff
@@ -0,0 +1,336 @@
...
+
+#
+# MCUBoot files/includes
+#
+set(MCUBOOT_DIR ${SRC_DIR}/third_party/mcuboot)
+set(MCUBOOT_FILES
+    ${MCUBOOT_DIR}/boot/bootutil/src/boot_record.c
+    ${MCUBOOT_DIR}/boot/bootutil/src/bootutil_misc.c
+    ${MCUBOOT_DIR}/boot/bootutil/src/bootutil_public.c
+    # ${MCUBOOT_DIR}/boot/bootutil/src/caps.c
```
evanShap, line 167: @uma-kat, did you comment these out, or were these commented out lines copy-pasted from a different cmake file?

src/lib/bristlefin/bristlefin.cpp
```diff
@@ -75,6 +75,14 @@ void Bristlefin::disableVout() {
   IOWrite(&BF_PL_BUCK_EN, 1); // 0 enables, 1 disables. Vout
 }
 
+void Bristlefin::sdi12Tx(){
+  IOWrite(&BF_SDI12_OE, 0); // reset to enable TX of data
+}
+
+void Bristlefin::sdi12Rx(){
+  IOWrite(&BF_SDI12_OE, 1); // set to disable TX and enable RX of data?
```
evanShap: @uma-kat please resolve the "?" on this one - does or doesn't it enable Rx?

src/lib/bristlefin/bristlefin.cpp
evanShap: ➕

src/lib/bristlefin/bristlefin.h
```diff
@@ -46,6 +46,10 @@ class Bristlefin {
   void enable3V();
   void disable3V();
 
+  // sdi
```
evanShap: @uma-kat - should make this more verbose, something like:
`// functions to switch between SDI12 Tx and Rx`

evanShap (changes requested):
@uma-kat looks great!
- I concur with all of @victorsowa12 's requested changes.
- I added a couple minor requests to cleanup some code comments.

mbella-sofar (commented):
LGTM

ENV_SETUP.md
```diff
@@ -245,6 +245,16 @@ To see fancy colored output with test details, run:
...
+To run an individual test suite, from the **test-build** directory run eg.:
+```
+$ ./test/src/common/differenceSignal
+```
+
+
+### Adding unit tests
+1. Add your test suite cpp file in **test/src**.
+2. Add a section for the new test suite to **test/src/common/CMakeLists.txt** with: `add_executable`, `target_include_directories`, `target_sources`, `target_link_libraries`, and `add_test`.
+
```
matt001k, line 257: Nice addition!

src/apps/bm_devkit/sdi12_example/user_code/exo3s_sensor.h
```diff
@@ -0,0 +1,103 @@
...
+#define breakTimeMin 15		// spec is 12 but a bit extra does no harm
+#define markTimeMin 9		// actually 8.33 but I am not OCD
```
matt001k: Can these be names to include `ms` somewhere in the name?
`breakTimeMin` could be confused with minutes,
maybe like `breakTimeMs` is sufficient enough.

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
```diff
@@ -0,0 +1,196 @@
...
+         "wiper pos: %.3f V, "
+         "depth: %.3f m, "
+         "power supply: %.3f V\n", sens.pH_mV, sens.dis_oxy, sens.dis_oxy_mg, sens.turbidity, sens.wiper_pos, sens.depth, sens.power);
+
+  current_sample_index++;
+  if (current_sample_index == sample_count){
+    transmit_samples();
+    current_sample_index = 0;
+  }
+  vTaskDelay(pdMS_TO_TICKS(2000));
```
towynlin: The LED code below is non-blocking, whereas this delay blocks `loop`. That means that LED1 won't actually pulse green every second as the code implies. Also LED2 won't pulse green on payload input. Both will probably pulse every 2 seconds.

It's fine as is. However, if you have time, it would be an improvement to make this sonde interaction also non-blocking. The steps would be:
- Delete the call to `vTaskDelay`.
- Extract lines 98 to 156 (the whole sonde interaction) into its own function named something like `static void sample_exo3_sonde(void)`
- Add a `#define EXO3_READING_PERIOD_MS 2000` (or similar name) up around line 20.
- At the top of `loop` do something like this:

```C++
static uint64_t last_exo3_reading_uptime = uptimeGetMs();
if (uptimeGetMs() - last_exo3_reading_uptime >= EXO3_READING_PERIOD_MS) {
  last_exo3_reading_uptime = uptimeGetMs();
  sample_exo3_sonde();
}
```

towynlin (commented):
Looks Great @uma-kat! 🙌 

Also a big 👍 to everything @victorsowa12 said. Thanks for the very thorough review!

I only had one additional comment. It's not required, but if you have time, it would be an improvement to the timing of how the loop works. Cheers!

src/apps/bm_devkit/sdi12_example/user_code/exo3s_sensor.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+  switch (cmd) {
```
matt001k, line 126: Can these cases have an enum associated with them? It is hard to understand what each number indicates,
something like:
```
enum {
  SDI_ADDRESS_INQUIRY_CMD,
  SDI_INDENTIFICATION_CMD,
  SDI_START_MEASUREMENT_CMD,
};
```

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
```diff
@@ -0,0 +1,196 @@
...
+    printf("%llut - %s | Failed to send Spotter transmit data request\n", uptimeGetMs(), rtcTimeBuffer);
+  }
+
+  // resetting the buffer
+  memset(samples, 0, sizeof(samples));
+}
+
+void setup(void) {
+  // samples per send
+  userConfigurationPartition->getConfig("sampleCount", strlen("sampleCount"), sample_count);
```
matt001k, line 75: Should we make sure that we limit `sample_count` here to `MAX_TX_BUFFER_SIZE`?
Something on line 78 like:
```
if (sample_count * SAMPLE_SIZE > MAX_TX_BUFFER_SIZE) {
  printf("sampleCount value is larger than max buffer size: %d, limiting sampleCount to: %d", MAX_TX_BUFFER_SIZE, MAX_TX_BUFFER_SIZE / SAMPLE_SIZE);
  sample_count = MAX_TX_BUFFER_SIZE / SAMPLE_SIZE;
}
```

matt001k (dismissed):

src/apps/bm_devkit/sdi12_example/CMakeLists.txt
uma-kat: @evanShap I don't recollect editing this `CmakeLists.txt` file. Could these have been auto generated and added here? :o

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
uma-kat: @towynlin I shall delete the 2 second delay. I had added that initially for debug purposes and is not really needed for the sensor to do its thing! Thanks for catching that

towynlin: Got it. If we don't delay, does the sensor only provide readings on a certain cadence or would that gather lots of readings really quickly, leading to our sending too much data via `spotter_tx_data`?

src/lib/bristlefin/bristlefin.cpp
uma-kat: I shall remove the `?` since it does enable Rx

ENV_SETUP.md
uma-kat: That's Evan's addition!

src/apps/bm_devkit/sdi12_example/user_code/exo3s_sensor.cpp
uma-kat: I can see that! I'll make these an enum 👍

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
uma-kat: The sensor does not rely on the delay. It takes a long time to make a measurement and send it over (62 SECONDS). So we have about 40 bytes of data each time we inquire a measurement. I have grouped 3 such measurements (samples) together and tx it to the spotter along with some RTC timestamp. 
Evan mentioned we have a 311 bytes limit for `spotter_tx_data`. So 29 header bytes for the `BMBinData1` message type + 40 *3 bytes sensor data, 149 bytes  < 311 bytes!

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
```diff
@@ -57,7 +57,7 @@ void transmit_samples(){
   tx_hex[i * 2] = '\0'; // Null-terminate the string
 
   // print to console
-  printf("[exo3-sonde-sdi12] buffer to send | tick: %llu, rtc: %s, buff: %s\n", uptimeGetMs(), rtcTimeBuffer, tx_hex);
+//  printf("[exo3-sonde-sdi12] buffer to send | tick: %llu, rtc: %s, buff: %s\n", uptimeGetMs(), rtcTimeBuffer, tx_hex);
```
matt001k: Was this intentional?

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
```diff
@@ -154,7 +155,7 @@ void loop(void) {
     transmit_samples();
     current_sample_index = 0;
   }
-  vTaskDelay(pdMS_TO_TICKS(2000));
+//  vTaskDelay(pdMS_TO_TICKS(2000));
```
matt001k: Also this? I feel like if its not needed, maybe it is best to remove to avoid clutter and confusion.

matt001k (approved):
Code looks good!

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
uma-kat: Yes! It is. I shall delete it to remove confusion

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
uma-kat: I shall delete it

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
uma-kat: Forgot to add - the sensor does a measurement and sends it over only after we request it to (sending a measure command), and does not measure new data automatically.

towynlin (approved):

src/apps/bm_devkit/sdi12_example/user_code/user_code.cpp
matt001k: I believe the logic was correct before? Moving it below the if statement will not actually limit the sample count

uma-kat: Wouldn't having this check right before `userConfigurationPartition->get...sample_count);` adjust the sample_count value if it's too high? 
👀 Could you explain why you say it will not actually limit the sample_count?

matt001k: Well we are going to read/assign the `sample_count` in from the user configuration partition in the `getConfig` API as `sample_count` is passed in by reference.
If the user configuration partition sets this value larger than `MAX_TX_BUFFER_SIZE / SAMPLE_SIZE` then we run into the risk of a buffer overflow.

uma-kat: Gotcha! I reverted it back. Thanks for explaining.

src/apps/bm_devkit/sdi12_example/CMakeLists.txt
uma-kat: @evanShap I copied the CMakeLists.txt from serial_payload app. Lmk if I need to change this, otherwise could approve this PR if it looks alright?

victorsowa12 (approved):
🚀 LGTM!

evanShap (approved):

PR #196, opened 2024-10-24
Supported services port

matt001k:
This PR is opened to show the changes made to source files to make them C compatible and bm_core compatible. Next PR will be the migration of these files to `bm_core`.


src/lib/debug/debug_bm_service.cpp
```diff
@@ -27,7 +27,7 @@ static const CLI_Command_Definition_t cmdBmService = {
     "bmsrv:\n"
     " * bmsrv req <service> <data> <timeout s>\n"
     " * bmsrv req sysinfo <node id> <timeout s>\n"
-    " * bmsrv req cfgmap <node id> <1(hw)/2(sys)/3(usr)> <timeout s>\n",
+    " * bmsrv req cfgmap <node id> <1(sys)/2(hw)/3(usr)> <timeout s>\n",
```
matt001k, line 30: This did not align with what is actually happening when running this command...

src/lib/sys/configuration.cpp
matt001k, line 1: This API was moved to bm_core.

test/src/common/configuration_ut.cpp
matt001k, line 1: This unit test was moved to bm_core

src/lib/debug/debug_bm_service.cpp
towynlin: Good eye!

towynlin (approved):

PR #198, opened 2024-10-25
Migrates Services Files Over To bm_core

matt001k:
Places services files into bm_core


src/apps/bm_devkit/hello_world/CMakeLists.txt
```diff
@@ -268,6 +263,7 @@ set(BM_CORE_LWIP_INCLUDES
     ${SRC_DIR}/lib/lwip
     ${LWIP_INCLUDE_DIRS}
     ${BCMP_INCLUDES}
+    ${BRISTLEMOUTH_INCLUDES}
```
matt001k, line 266: This is needed right now because of bm_core's reliance on `bm_common_messages` repository.
I think we should discuss how to handle this in our Monday meeting.
It should be a submodule of `bm_core` but it is full of cpp files.

victorsowa12 (approved):

PR #203, opened 2024-10-30
Ports Bridge App To Use bm_core

matt001k:
Updates bm_serial and opens a branch for experiment/bm_core in bm_serial
Adjusts API to match what has been changed in bm_core


src/lib/bm_ncp/ncp_config.cpp
```diff
@@ -1,271 +1,292 @@
...
-            if (!bcmp_config_get(node_id, partition, key_len, key, err, _cfg_get_bcmp_cb)) {
-                bridgeLogPrint(BRIDGE_CFG, BM_COMMON_LOG_LEVEL_ERROR, USE_HEADER, "Failed to issue bcmp cfg get, err: %" PRIi8 "\n", err);
-                break;
-            }
-        }
-        rval = true;
-    } while(0);
-    if (valueBuf != NULL) {
-        vPortFree(valueBuf);
+bool ncp_cfg_get_cb(uint64_t node_id, bm_common_config_partition_e partition, size_t key_len,
```
towynlin, line 24: It looks like the changes to this file are mostly clang-format, with some include file name changing and some bool-to-BmErr return values. Anything else significant to review?

src/lib/bm_ncp/ncp_uart.cpp
```diff
@@ -224,16 +225,15 @@ static bool bcmp_info_request_cb(uint64_t node_id) {
     }
   } else {
     // send back the info for the node_id
-    bcmp_request_info(node_id, &multicast_global_addr, bcmp_info_reply_cb, NULL);
```
towynlin, line 227: What's the story of removing the final parameter?

src/lib/bm_serial
towynlin, line 1: The bm_serial change looked like just clang-format and an `add_subdirectory` for common messages. Is that it? Say more about why bm_serial needed to change.

towynlin (approved):
Looks good. A couple questions.

src/lib/bm_ncp/ncp_config.cpp
matt001k: Nope that is accurate :)

src/lib/bm_serial
matt001k: It was referencing the old `src/lib/bm_common_messages` in it, and actually was not referencing the one that was inside of the submodule itself `bm_serial/bm_common_messages`

src/lib/bm_ncp/ncp_uart.cpp
matt001k: This function always just had 3 parameters? Somehow we were able to pass 4 into it when it was a C++ function? Take a look at it from develop: https://github.com/bristlemouth/bm_protocol/blob/05b953254685c1463fc07c871c556f5f3cfad04f/src/lib/bcmp/bcmp_info.cpp#L33

matt001k: Cancel, that, somewhere in the lands between develop and where we are now this file somehow got an extra argument in these function calls? Let me look through bm_core history to see why.

matt001k: Looking at a diff of the history of that file in bm_core, I cannot see where an extra parameter was added to the info bcmp request. Maybe the NULL argument was added here by accident some time ago and this file was never built until now.

towynlin: yowsa, ok, thanks for cleaning it up

matt001k: Commit 4d77bb13 by me in bm_protocol added this in. Darn Git Blame.

PR #205, opened 2024-10-31
CMake Presets!

towynlin:
Since I blew away my previous IDE settings while debugging and was reconstructing them, I learned about a new feature — CMake Presets! As of this PR, you can now run the following commands and not have to remember all the magical incantations! I made the presets build in a new `preset-builds` folder so we don't overwrite existing build folders that people may have.

```
cmake --list-presets
cmake --preset bootloader
cmake --build --preset bootloader
cmake --preset hello-world
cmake --build --preset hello-world
cmake --preset bridge
cmake --build --preset bridge
cmake --preset unit-tests
cmake --build --preset unit-tests
ctest --output-on-failure --test-dir preset-builds/unit-tests
```

Running `cmake --preset hello-world` for example *configures* the build, creating the build folder if it does not exist. After that you can just run `cmake --build --preset hello-world` each time you want to build.

This all integrates super well in VS Code too! Just open the side panel for the CMake extension and choose (near the top) the configure and build presets you want. Then F7 builds your selected preset.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ tools/bin/mpy-cross
 .vscode
 cli-cmake-debug-build/
 *build/
+preset-builds
 
 src/bsp/pluteus_v3.1/*.s
 src/bsp/pluteus_v3.1/*.ld
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,60 @@
+{
+    "version": 8,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 29,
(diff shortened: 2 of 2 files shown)


towynlin:
Ah, build failing waiting on @matt001k's work to update ALL THE THINGS to use `bm_core`. This PR doesn't touch any of the existing source though.

matt001k:
Working on it! Nice work here I like TIL. This is 🔥

matt001k (approved):

victorsowa12 (approved):

PR #207, opened 2024-11-01
Refactors Configuration Module To Be Utilized In bm_core

matt001k:
Changes the way the functions operate in order to be able to make it compatible with C
Adds functions the user must implement for reading/writing to flash and restarting the system
Updates bm_devkit apps to utilize this new API

This PR is for visibility and getting your opinions on the changes to the configuration module.
I went with a callback approach here, but it can easily be made into something similar like what @victorsowa12 did with the generic header files (more for convenience at this stage).
If you want to pursue with this solution I will make sure to get everything building and working as expected.
Currently I have it running for the `hello_world` application and tested out the core functionality of the config module via the CLI.

Biggest problem this has here, is that it changes the API used to access configurations, which is n issue for `bm_protocol`, if we want to prevent breaking API changes, I suggest that we create a wrapper file against these new bm_core calls in order to prevent this scenario.

Let me know your thoughts!


src/apps/bm_devkit/bmdk_common/app_main.cpp
```diff
@@ -138,9 +139,9 @@ SerialHandle_t usbPcap = {
 };
 
 // TODO - make a getter API for these
-cfg::Configuration *userConfigurationPartition = NULL;
-cfg::Configuration *systemConfigurationPartition = NULL;
-cfg::Configuration *hardwareConfigurationPartition = NULL;
+static NvmPartition *userConfigurationPartition = NULL;
+static NvmPartition *systemConfigurationPartition = NULL;
+static NvmPartition *hardwareConfigurationPartition = NULL;
```
matt001k: Was able to get rid of these global variables. Now they are just local to this file for the callback functions. As well as now pointers to NVM partitions for the callbacks.

src/apps/bm_devkit/bmdk_common/app_main.cpp
```diff
@@ -300,6 +301,50 @@ static const DebugGpio_t debugGpioPins[] = {
...
+    ret = hardwareConfigurationPartition->write(offset, buffer, length, timeout_ms);
+    break;
+  default:
+    break;
+  }
+
+  return ret;
+}
+
+static void config_reset_handler(void) { resetSystem(RESET_REASON_CONFIG); }
```
matt001k: New callbacks 👽

src/apps/bm_devkit/bmdk_common/app_main.cpp
```diff
@@ -375,24 +420,19 @@ static void defaultTask(void *parameters) {
...
+
+  get_config_uint(BM_CFG_PARTITION_SYSTEM, "sensorsPollIntervalMs",
+                  strlen("sensorsPollIntervalMs"), sys_cfg_sensorsPollIntervalMs);
+  get_config_uint(BM_CFG_PARTITION_SYSTEM, "sensorsCheckIntervalS",
+                  strlen("sensorsCheckIntervalS"), sys_cfg_sensorsCheckIntervalS);
   NvmPartition debug_cli_partition(debugW25, cli_configuration);
   NvmPartition dfu_partition(debugW25, dfu_configuration);
   dfu_partition_global = &dfu_partition;
-  debugConfigurationInit(&debug_configuration_user, &debug_configuration_hardware,
-                         &debug_configuration_system);
```
matt001k, line 395: A lot of functionality that relied on passing around the configuration objects to different modules can now be removed.

src/lib/sys/bm_config_wrapper.cpp
matt001k, line 1: This file would get removed when migrated to bm_core rather than relying on an object, we now have the 3 partitions declared here. It is up to the integrator to write the 3 callback functions to handle where the configs are read/written to.

src/lib/sys/configuration.cpp
matt001k, line 1: Meat and potatoes

test/src/common/configuration_ut.cpp
```diff
@@ -150,152 +187,127 @@ TEST_F(ConfigurationTest, BasicTest)
...
+//TEST_F(ConfigurationTest, TooLittleStorage)
+//{
+//  const ext_flash_partition_t test_configuration = {
+//      .fa_off = 4096,
+//      .fa_size = 10000,
+//  };
+//  uint8_t small_storage;
+//  NvmPartition testPartition(_storage, test_configuration);
+//  EXPECT_DEATH(Configuration config(testPartition,&small_storage,sizeof(small_storage)), "");
+//}
```
matt001k: This will have to get sorted...

src/lib/sys/configuration.h
```diff
@@ -42,41 +33,31 @@ typedef struct ConfigPartition {
...
-    bool prepareCborEncoder(const char * key, size_t key_len, CborEncoder &encoder, uint8_t &keyIdx, bool &keyExists);
-    bool loadAndVerifyNvmConfig(void);
-
-    static constexpr uint32_t CONFIG_START_OFFSET_IN_BYTES = 0;
-    static constexpr uint32_t CONFIG_LOAD_TIMEOUT_MS = 5000;
+typedef bool (*ConfigRead)(BmConfigPartition partition, uint32_t offset, uint8_t *buffer, size_t length, uint32_t timeout_ms);
+typedef bool (*ConfigWrite)(BmConfigPartition partition, uint32_t offset, uint8_t *buffer, size_t length, uint32_t timeout_ms);
+typedef void (*ConfigRestart)(void);
+
+void config_init(ConfigRead read, ConfigWrite write, ConfigRestart restart);
```
towynlin: The only change I'd recommend is passing a struct like your `ConfigCb` that holds the 3 callbacks rather than 3 separate params, because it would allow us to add and remove callbacks in the future without changing the signature of the function.

towynlin (approved):
Looks good to me! Had one suggestion.

victorsowa12 (approved):

src/lib/sys/configuration.h
matt001k: Yea I agree! Good callout!

src/apps/bm_devkit/bmdk_common/app_main.cpp
matt001k: Sadly these have to be globals now again with the config wrapper updates

src/lib/sys/configuration.h
matt001k: This API now takes 0 arguments due to not using callbacks anymore.

src/lib/sys/bm_config_wrapper.cpp
matt001k: Now that we are not using callbacks, this file will stay!

victorsowa12 (approved):

PR #211, opened 2024-11-14
Update BSP For Bridge v7 USB Detection

matt001k:
Swap Green LED pin and USB detect pin


matt001k:
After some discussion, if both of these pins are inputs, we do not run into the risk older/newer versions of the board's USB port not working. The following is a rundown of what would happen:

```
old board PA1 to green led
new board PA1 to VUSB

old fw PA1 output low
new fw PA1 input
unlikely pathology: old boards with new fw if jumper connected leak current through led

----

old board PH0 to VUSB
new board PH0 to green led

old fw PH0 input
new fw PH0 input
unlikely pathology: new boards with new fw if jumper connected leak current through led
```

victorsowa12 (approved):

src/apps/bridge/app_main.cpp
```diff
@@ -373,6 +380,16 @@ static void defaultTask(void *parameters) {
                                        sys_cfg_sensorsCheckIntervalS);
   NvmPartition debug_cli_partition(debugW25, cli_configuration);
   NvmPartition dfu_partition(debugW25, dfu_configuration);
+  uint32_t hw_version = 0;
+  if (!debug_configuration_hardware.getConfig(
+          AppConfig::HARDWARE_VERSION, strlen(AppConfig::HARDWARE_VERSION), hw_version)) {
+    //TODO: next commit place some logic in here to activate serial terminal
```
matt001k: There needs to be logic to activate USB here. Since we want to default to older version of the hardware, that poses a problem for v7 boards for when we want to set the revision version in the hwCfg.
Thoughts on how we should do this?
If we set the cb in `usbInit` to `NULL` the usb port will always be open.

towynlin (approved):

PR #213, opened 2024-11-19
Serial Payload Example supports SDI-12

evanShap:
- need to figure out where to put the SDI-12 specific code - BF? PLUART? New file? BF makes the most sense, but currently isn't aware of PLUART.
- need to implement configurable Tx line term in PLUART Debug CLI.


src/apps/bm_devkit/serial_payload_example/user_code/user_code.cpp
```diff
@@ -33,9 +36,53 @@ static u_int32_t uart_mode_config = DEFAULT_UART_MODE;
 // A buffer for our data from the payload uart
 char payload_buffer[2048];
 
+void sdi_wake(void) {
```
evanShap: @towynlin @victorsowa12 @matt001k - `sdi_wake` and `sdi_break_mark` are functions that are required for managing serial transactions with an SDI-12 sensor.

@uma-kat and I think these should probably go into the Bristlefin namespace as these are dealing with specific pins and the specific behavior of the SDI-12 transceiver on the BF board, but that would bring PLUART into bristlefin.h.

Please let us know if you think these functions should go somewhere else - PLUART directly? new file that includes BF and SDI12?...

evanShap: We could also abstract out some of the logic in setup like `bristlefin::setup_rs485_hd()`, `bristlefin::setup_sdi12()`, which would be nice.

victorsowa12: I don't have any issues with it. Makes sense to put the SDI-12 stuff into the bristlefin namespace since bristlefin's are the only board that can use SDI-12. Also no problem for pulling in PLUART since it is the shared library for all board types.

src/apps/bm_devkit/serial_payload_example/user_code/user_code.cpp
```diff
@@ -58,12 +116,19 @@ void setup(void) {
...
   if (uart_mode_config == MODE_RS485_HD) {
     printf("Enabling RS485 Half-Duplex.\n");
     bristlefin.setRS485HalfDuplex();  // Set the RS485 transceiver to Half-Duplex.
     // Set up PLUART transactions for proper RS485 enable Tx / enable Rx
     PLUART::enableTransactions(Bristlefin::setRS485Tx, Bristlefin::setRS485Rx);
+  } else if (uart_mode_config == MODE_SDI12) {
+    printf("Enabling SDI-12 serial.\n");
+    // Set up PLUART transactions for proper SDI12 enable Tx / enable Rx
+    PLUART::enableTransactions(sdi_break_mark, Bristlefin::sdi12Rx);
   }
```
towynlin: Since all 3 of the other hardware control functions here are in the `Bristlefin` namespace, the first thing that occurs to me is that maybe it makes sense to move `sdi_break_mark` there as well.

However, all those other methods just involve calling `IOWrite` on various pins, whereas `sdi_break_mark` is entirely a sequence of `PLUART` calls with some additional timing. The same is true of `sdi_wake`.

The `Bristlefin` nameespace doesn't know about `PLUART`, and these SDI-12 sequences require too many calls to shoehorn it in there with some callbacks or duck typing. Putting the SDI-12 functions in the `Bristlefin` namespace would required adding a lot of dependencies that are otherwise unused there. That's a pretty clear sign that we should design it differently.

I think it makes the most sense to move `sdi_wake` and `sdi_break_mark` into the `PLUART` namespace. I'm pretty sure all the necessary dependencies are already there. That's my recommendation.

src/apps/bm_devkit/serial_payload_example/user_code/user_code.cpp
matt001k: I feel like these functions themselves should be located outside of this file. Analyzing the bristlefin class,  I feel like this would be the perfect location currently for the API  mentioned as this is what houses `sdi12Tx/sdi12Rx`. Personally I do not see an issue bringing in PLUART, and I doubt that it would have to be included in the header file.

Personally I believe it would be nice to see all of the code encapsulated in a single module, is there any reason why we could not use the `exo3s_sensor.h` API here? Maybe that file can be moved to somewhere in `src/lib/drivers` and then be used in this file directly?
Or is `sdi_wake` and `sdi_break_mark` generic for every type of SDI12 type device?
If that is the case, could we create a module called `sdi12.h` in `src/lib/drivers` and utilize that here?
Maybe as a future task configure that driver to be agnostic to the platform used such as the pins, etc.

evanShap: Thanks @victorsowa12 and @matt001k, `sdi_wake` and `sdi_break_mark` are generic for all SDI12 devices, and the implementation in these functions is specific to Mote + Bristlefin - not specific to the EXO3, and not abstract enough to work on a different platform + interface board.

As a first step I like migrating these to bristlefin.h and updating exo3_sensor.h to use those functions.

I'll give that a go and poke for re-review when it's ready for a look.

src/apps/bm_devkit/serial_payload_example/user_code/user_code.cpp
evanShap: @towynlin `sdi_break_mark` needs bristlefin.h. Do we want to add bristlefin to PLUART? A bristlefin board always has a Mote with a PLUART, but a Mote with a PLUART isn't always connected to a bristlefin board.

Or, do we put these SDI12-specific sequence functions in PLUART, but then have the caller responsible for passing in the SDI12-tx-enable function per whatever board lib it's using. Imagining we make another adapter board some day with an SDI-12 transceiver on it, this would require less copy-pasting than putting stuff in bristlefin.h. However, we're then bloating PLUART with code that isn't relevant to most systems that have a PLUART. All systems importing this file will have a Mote with a PLUART, but many of them won't have an SDI-12 transceiver. If we want to avoid that, we could alos just introduce an sdi12.h file.

towynlin: @evanShap  Ah! I didn't realize that PLUART didn't already have access to Bristlefin. Well, derp. Probably then a separate `sdi12.{c,h}` would be best then actually.

evanShap: Yea, I'm coming around to separate sdi12 and rs485 files being the most future-proof approach. Given that, a couple more questions:
- What is the preference style guide for implementing these - inlines in a single .h, implement in .c, implement in .cpp?
- I like the idea of including and wrapping these in bristlefin.h, rather than separately including these and bristlefin.h in the app. Reasoning:
  - Bristlefin HW _always_ has access to SDI12 and RS485, (even though a particular app might not use them).
  - This minimizes/simplifies the app code, which I think is pretty important given the level of experience of most of the users.
  - This still provides a clean + clear path for implementing different mote interface boards with some subset of RS485 and/or SDI-12:
    - 1. write a header for your carrier board, like bristlefin.h, but with your custom stuff.
    - 2. include and write wrappers for sdi12/rs485 as needed, but passing in the hooks for your carrier boards enable GPIOs.
    - 3. In your app, setup the PLUART transaction identically to how you do using bristlefin, just pass in a handle to your carrier board instead of the bristlefin.
@towynlin thoughts?

towynlin: Love it. 👍 Let the .h file be pretty minimal — just declare your public functions and any types that consuming code might need, then in a .cpp file write your implementations. (I only mentioned .c above because that's what we've been doing in `bm_core`, but here C++ is more likely the way to go.)

src/apps/bm_devkit/serial_payload_example/user_code/user_code.cpp
evanShap: @matt001k @uma-kat - finally getting around to this. Moving the SDI-12 functions to bristlefin was dead simple, and I've manually tested it.

src/apps/bm_devkit/serial_payload_example/user_code/user_code.cpp
evanShap: @towynlin - we'll probably need to save the above proposed abstraction for someone else. This has been back-burnered for 2 months, and I don't expect to have time to address it in the near term. 😿
I don't think it needs to gate this PR though, what we have now is functional and consistent with the pattern we already have established?

evanShap:
This one is finally ready for review! I realized we **don't** need to implement configurable PLUART write line termination for this - we can just tell users to add their own '!'. I updated the doc to reflect that @uma-kat .

matt001k (commented):
Updates look good!

matt001k (approved):

PR #219, opened 2024-11-25
Update Bridge To Inform Spotter Of Its Hardware Revision

matt001k:
Allows bridge to inform the spotter of its hardware revision

This is to go along with:
https://github.com/bristlemouth/bm_serial/pull/23

Some updates needed to be made to bm_serial to obtain the NodeID of the bridge from spotter.
I added this capability because sometimes the initial topology report generation would not be seen by spotter and we would have to wait 60 seconds for the next one. The topology report is how we were obtaining the NodeID of the bridge from spotter originally. But this streamlines the communication.


src/lib/common/device_info.h
```diff
@@ -13,15 +13,17 @@ const versionInfo_t *findVersionInfo(uint32_t addr, uint32_t len);
 
 bool fwIsEng(const versionInfo_t *info);
 bool fwIsDirty(const versionInfo_t *info);
-const uint32_t* getUID();
-const char * getUIDStr();
-const char * getFWVersionStr();
+const uint32_t *getUID();
+const char *getUIDStr();
+const char *getFWVersionStr();
```
towynlin: nit: as long as we're editing this, please add `void` to the parameter list

towynlin (approved):

src/lib/common/device_info.h
matt001k: Ok!

PR #221, opened 2024-11-27
Fix bridge with raw pressure

towynlin:
I noticed that a recent merge commit https://github.com/bristlemouth/bm_protocol/commit/56b7a9a8c320d96ee2d3aea0546995c8253e2e2a#diff-2c23633dd934ba3ff7f6be1ea6872d2f8a369c65dcc90173da6159723e943d67R438 had conflict markers in `bridge/app_main.cpp`.

CI was passing because that part of the bridge code was only evaluated when `RAW_PRESSURE_ENABLE` is defined, which was not a part of any of our CI builds. So, I added that as well as a cmake preset. There were other failures as well related to configs, which I cleaned up.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -71,6 +71,13 @@
             }
         },
         {
+            "name": "bridge-raw-pressure",
+            "inherits": "bridge",
+            "cacheVariables": {
+                "RAW_PRESSURE_ENABLE": "1"
+            }
+        },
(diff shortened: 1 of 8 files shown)


matt001k (approved):
Nice catch 💯 🥇 🙏

victorsowa12 (approved):

PR #224, opened 2024-12-04
Fix/remove dfu confirm from generic

victorsowa12:
Remove the dfu confirm checks from the dfu generic/wrapper files.

Waiting on https://github.com/bristlemouth/bm_core/pull/63, then need to point bm_core back to main!


src/lib/drivers/bm_dfu_wrapper.cpp
```diff
@@ -65,20 +64,6 @@ uint32_t bm_dfu_client_flash_area_get_size(const void *flash_area) {
...
-  return val == 1;
-}
-
-void bm_dfu_client_confirm_enable(bool en) {
-  uint32_t val = en;
-  set_config_uint(BM_CFG_PARTITION_SYSTEM, dfu_confirm_config_key,
-                  strlen(dfu_confirm_config_key), val);
-  save_config(BM_CFG_PARTITION_SYSTEM, true);
-}
-
```
victorsowa12, line 81: these have been moved to the dfu_client.c file in `bm_core`. See https://github.com/bristlemouth/bm_core/pull/63

test/header_overrides/bm_config.h
victorsowa12, line 1: This was the old header override of the old bm_config file. These `#ifndef __BM_CONFIG_H__
#define __BM_CONFIG_H__` were causing the new `bm_config.h` functions from not being available in the unit tests, which they are now needed. Specificially `bm_debug` is needed and it was not being included. Now it is.

src/apps/bridge/CMakeLists.txt
```diff
@@ -289,7 +289,7 @@ include(${SRC_DIR}/lib/bm_core/cmake/bm_core.cmake)
 setup_bm_ip_stack(LWIP "${BM_CORE_LWIP_INCLUDES}")
 setup_bm_os(FREERTOS "${BM_CORE_FREERTOS_INCLUDES}")
 add_subdirectory(${SRC_DIR}/lib/bm_core bmcore)
-target_include_directories(bmcore PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../bmdk_common)
+target_include_directories(bmcore PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
```
victorsowa12, line 292: While looking into why `bm_debug` was not being included in the unit test builds I found this. I don't think the bridge should be referencing bmdk common files (nor is this the correct path to access it).

TODO - test bridge with this change.

victorsowa12: Letting a unit run overnight and will observe for any crashes in the morning.

victorsowa12: No crashes over night, sampling aggregation and logging all worked!

towynlin (approved):

PR #231, opened 2024-12-10
Extracts essential Bristlemouth functionality into bm_core submodule

towynlin:
## What changed?

A lot! The intent has been to make a very large refactor, with no breaking changes, to extract the core of Bristlemouth into a separate `bm_core` repo which builds as a static library and is suitable for use by hardware integrators to add Bristlemouth capability to their devices.

This was not feasible with the prior structure. Now, the following dependencies have been made optional, with a clear path to replacing them as needed.

- Operating System (FreeRTOS is the only current implementation)
- IP Stack (lwip is the only current implementation)
- Single-Pair Ethernet chip driver (ADIN2111 is the only current implementation)

There is a lot more documentation, and there are more unit tests. Automated hardware-in-the-loop tests are in the works.

With `bm_core` containing the essentials of the Bristlemouth protocol, this repo is in the process of transitioning into a set of example applications, with some common build infrastructure, that each uses `bm_core`. There are also useful libraries here as well as board support packages for the extant dev kit hardware.

Bristlemouth is now much more accessible to organizations with existing hardware and firmware that might consider contributing to and benefiting from the growing ecosystem of Bristlemouth-capable devices. It is flexible enough to run on hardware and software stacks different from the original dev kit hardware and firmware.

We have already done a great deal of integration testing to ensure compatibility, so this is a final review before we transition to `bm_core` becoming the base for future releases.

## Checklist


victorsowa12 (approved):
🎉

matt001k (approved):
🥳

evanShap (approved):
🚢

PR #232, opened 2024-12-11
Dfu hil

matt001k:
Adds BCMP commands for DFU HIL test.


victorsowa12 (approved):
I really like this! I'm a little worried about potential confusion between the `bm dfu` commands and the `dfu start` commands. Nothing that some good docs and maybe adding something to the `dfu start` help string to specify that its updating from the external flash can't help with.

PR #233, opened 2024-12-11
Migrate FreeRTOS dependencies to bm_core in bmdk app main

towynlin:
There are more than 200 references to FreeRTOS.h in this repo that should migrate to use functions from `bm_os.h`. This is one example of how they can be migrated.


matt001k:
Ohhhhhhh looks pretty 🌹

victorsowa12 (approved):
🤘

PR #234, opened 2024-12-13
Reduce The Developer Load For Application CMakeLists.txt Integration

matt001k:
Encapsulates library and third_party source and header file dependencies into individual CMakeLists.txt files

Build bm_core in src/CMakeLists.txt

Moves micropython dependency to src/CMakelists.txt

Cleans up all app CMakeLists.txt files


src/apps/bringup_bridge/task_priorities.h
```diff
@@ -7,6 +7,8 @@
 
 #define DEFAULT_BOOT_TASK_PRIORITY 16
 
+#define STRESS_TASK_PRIORITY 1
+#define SENSOR_SAMPLER_TASK_PRIORITY 1
```
victorsowa12: are these tasks used by this app? 

Also I think the priorities are different in the other apps, so it would be good to keep them consistent.

Small nit: keeping them in numerical order

victorsowa12 (approved):
Awesome! Had a small question + nit about the task priority changes, but love all the deletion!

src/CMakeLists.txt
```diff
@@ -321,6 +323,23 @@ add_link_options(
...
+if(NOT APP MATCHES "bootloader")
+add_subdirectory(lib)
+add_subdirectory(third_party)
+endif()
+
+add_compile_definitions(
+	${BSP_DEFINES}
+    ${LIB_DEFINES}
+	${APP_DEFINES}
+	${LINKER_DEFINES}
```
towynlin, line 336: nit: indentation

src/CMakeLists.txt
```diff
@@ -375,17 +394,96 @@ endif() # if (APP STREQUAL "bootloader")
 
 target_sources(${EXECUTABLE} PRIVATE
 	${BSP_FILES}
+    ${LIB_FILES}
+    ${THIRD_PARTY_FILES}
 	${APP_FILES}
```
towynlin, line 399: nit: indentation

src/CMakeLists.txt
```diff
@@ -375,17 +394,96 @@ endif() # if (APP STREQUAL "bootloader")
...
 	${KEY_FILES}
 	${CMAKE_CURRENT_BINARY_DIR}/version.c)
 
 target_include_directories(${EXECUTABLE} PRIVATE
 	${BSP_INCLUDES}
+    ${LIB_INCLUDES}
+    ${APP_DIR_COMMON}
+    ${THIRD_PARTY_INCLUDES}
 	${APP_INCLUDES}
```
towynlin, line 408: nit: indentation

src/CMakeLists.txt
```diff
@@ -375,17 +394,96 @@ endif() # if (APP STREQUAL "bootloader")
...
+set_source_files_properties(
+    ${LIB_DIR}/micropython/micropython_freertos.c
+
+    DIRECTORY ${SRC_DIR}
+
+    PROPERTIES
+    COMPILE_FLAGS -Wno-unused-parameter
+    )
+
+endif()
```
towynlin, line 468: nit: add a `USE_MICROPYTHON` comment since this is dozens of lines away from the if

src/apps/bridge/memfault_platform_config.h
```diff
@@ -19,8 +19,3 @@
 #define MEMFAULT_METRICS_HEARTBEAT_INTERVAL_SECS (6 * 60 * 60)
 
 #define MEMFAULT_WATCHDOG_SW_TIMEOUT_SECS 30
-
-// If you want to enable heap stats collection while debugging, uncomment the following lines
-// #define MEMFAULT_COREDUMP_COLLECT_HEAP_STATS 1
-// #define MEMFAULT_FREERTOS_PORT_HEAP_STATS_ENABLE 1
-// #define MEMFAULT_COREDUMP_HEAP_STATS_LOCK_ENABLE 0
```
towynlin: Let's keep these commented lines in! It's tough to discover all these variables when this kind of debugging comes up. They're also a good reminder that this functionality is available.

src/lib/CMakeLists.txt
```diff
@@ -0,0 +1,184 @@
...
+        ${BM_NCP_FILES}
+    )
+endif()
+
+if(APP_NAME MATCHES "bringup")
+    list(APPEND LIB_FILES
+        ${LIB_DIR}/debug/debug_uart.c
+        ${LIB_DIR}/debug/debug_adin_raw.c
+    )
+endif()
```
towynlin, line 114: Haha! Just highlighting as an interesting hack. I see you moved some files into the bridge app folder, but I guess these generally useful, so should stay in lib, even if they're only used in a couple apps. 👏

towynlin: Ah! The BMDK one below is similar too.

src/lib/common/Exo3LineParser.cpp
```diff
@@ -5,6 +5,10 @@
 #include "Exo3LineParser.h"
 #include "string.h"
 
+#ifndef __clang__
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+#endif
```
towynlin: Could you explain this addition?

towynlin (commented):
Mostly nits. One place where I'd love to understand why you added a compiler-specific directive.

src/apps/bringup_bridge/task_priorities.h
matt001k: Yes I can go ahead and change that!
This app does not use it butttttt with the addition of the lib source files, these two tasks are defined within there :)

src/apps/bridge/memfault_platform_config.h
matt001k: I think I accidently removed that file and then copied another directories `memfault_platform_config.h` file back into this directory.

src/lib/common/Exo3LineParser.cpp
matt001k: This was failing on our builds. We currently are not building the SDI12 app (which we probably should and I will add that to another commit), but I cannot see how that build was passing due to this error failing compilation.
This warning is something that is not very serious as long as you are aware of what you are doing (which looking at the logic it makes sense), see [this](https://stackoverflow.com/questions/50198319/gcc-8-wstringop-truncation-what-is-the-good-practice) for a rundown.

This was occurring at line 44, because we assign token_len at line 34.

matt001k: I will dig in deeper to find out if this was building on develop and why.

towynlin: Nice. Thanks. I think that the warning is valuable and that we should fix the code rather than suppress the warning.

It's interesting that the compiler can't connect the dots to see that the final copied character will always be a null terminator. We can add insurance and help the compiler by inserting:

    token_copy[token_len + 1] = '\0';

immediately after the strncpy. I think then the warning should go away.

matt001k: Addressed this issue with a memcpy here.
The way that the `strncpy` was intended to be originally used was as a `memcpy` as `strncpy` will append a zero to the destination if the source is less than the length parameter.

PR #237, opened 2024-12-31
CICD HIL Test On Local Runner

matt001k:
## What changed?
Adds script to run HIL tests/update a node to the latest FW push.
When a PR is opened, the hello_world app will be flashed onto that mote with the updated code.
That newly flashed mote will then run the HIL tests.
Adds `did_i_break_something.py` script command line option to build all apps and store them at `path/to/bm_protocol/build` rather than in a temporary directory. This allows for apps to be flashed from local runner.

TODO:
Upon releases, we should flash all motes to the latest release (automated with DFU). Then PRs opened after the release will always be run against motes on the latest release.

Automated HIL tests on every PR! WITH LOGS!

The python script, `bm_hil.py`. See if there are any improvements that can be made.
Currently we update the mote through dfu-util, but should we?
Or should we flash over the ST-Link.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,14 +28,12 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v3
-
       - uses: conda-incubator/setup-miniconda@v2
         with:
           miniconda-version: "latest"
           activate-environment: bristlemouth
           environment-file: environment_ci.yml
           auto-activate-base: false
-
       - name: Checkout submodules
         run: |
           git submodule update --init --recursive src/lib/bm_core
@@ -50,6 +48,7 @@ jobs:
           git submodule update --init src/third_party/tinyusb
           git submodule update --init src/third_party/BME280_driver
           git submodule update --init test/third_party/fff
-
       - name: Run Tests
-        run: python3 tools/scripts/test/did_i_break_something.py tools/scripts/test/configs
+        run: python3 tools/scripts/test/did_i_break_something.py tools/scripts/test/configs --store
+      - name: Run HIL
+        run: python tools/scripts/cicd/bm_hil.py ./build/hello_world/src/bm_mote_v1.0-hello_world.elf.dfu.bin C000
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -16,9 +16,9 @@ jobs:
   build:
     # The type of runner that the job will run on
     runs-on:
-      # - self-hosted
+      - self-hosted
       # Uncomment to run on github's infrastructure (and comment out self-hosted above!)
```
towynlin, line 20: Update comment before merge 😄

tools/scripts/cicd/logging_helper.py
```diff
@@ -0,0 +1,55 @@
...
+    ERROR = 40
+    WARNING = 30
+    INFO = 20
+    DEBUG = 10
+
+    def __init__(self, sub_dir: str, name: str, stdout: bool = False):
+        home = Path.home()
+        directory = str(home) + "/" + self.LOG_DIR + "/" + sub_dir
+        os.makedirs(directory, mode=0o777, exist_ok=True)
+        filename = directory + "/" + name + "_" + str(int(time())) + ".log"
```
towynlin: Thanks for adding time to the log file name. 👍

On the machine right now I see these files: `hil_test_1735594072.log`, `hil_test_1735669825.log`, `hil_test_1735670427.log`, `hil_test_1735671351.log`. I'm imagining seeing something wrong on github and wanting to find the right log file. Maybe to make debugging easier and faster, I'd suggest also adding short git sha of the commit we're testing to the file name after the timestamp. Would that be reasonably easy? If not in the filename, then near the top of the log file content could be good too.

towynlin (approved):
Looks good to me! Had a couple minor suggestions.

towynlin:
Oh, and yeah to respond to your comment — I do think dfu-util is a better choice right now, mainly because of fewer wires in the setup. Also I've seen some inconsistent behavior when using openocd and the stlink to flash some non-standard setups lately. I probably have some assumptions about how it should work that aren't correct. If we find that we have dfu failures often then we could change it to swd flashing in the future and see how that goes.

tools/scripts/cicd/logging_helper.py
matt001k: That is a real good idea I will look at adding it!

victorsowa12 (approved):

PR #239, opened 2025-01-03
Add Borealis Handling Placeholder In Bridge App

matt001k:
## What changed?
Creates `borealisSensor` module to send individual readings to spotter from a subscribed topic,
this currently houses a placeholder for the decode function that is yet to be implemented.

Updates `sensorController` with ability to add borealis nodes to list of subscribers.

Adds generic functions that can be applied to all supported sensor nodes.
This is to reduce copy-pasta in many `xxxxxSensor.cpp` files.
These functions perform the following:

- update a sensor node's positioning in the topology
- send individual logs to spotter

I did not replace all of the code from other sensors with these generic functions,
as I felt that was not in the scope of this work and might create unnecessary clutter.

Adds support for new Borealis sensor to bridge app.
Adds API to reduce code use in the future for new sensors in the bridge app.

Mainly the two new source files:
- `borealisSensor.cpp`
- `abstractSensor.cpp`

I tried my best to utilize all `bm_core` API that I could,
`bm_malloc`, `bm_free`, `bm_get_tick_count`, etc.,
but, there is no API for creating mutexes in `bm_os.h`, is this something we want to add to avoid FreeRTOS pollution in our pretty os abstraction?

## Checklist


src/apps/bridge/sensorController.cpp
```diff
@@ -109,6 +111,22 @@ void sensorControllerInit(BridgePowerController *power_controller) {
...
+    bridgeLogPrint(
+        BRIDGE_CFG, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
+        "Failed to get borealis reading period from config, using default value and writing "
+        "to config: %" PRIu32 "ms\n",
+        _ctx.borealis_reading_period_ms);
+    set_config_uint(BM_CFG_PARTITION_SYSTEM, AppConfig::BOREALIS_READING_PERIOD_MS,
+                    strlen(AppConfig::BOREALIS_READING_PERIOD_MS),
+                    _ctx.borealis_reading_period_ms);
+    save = true;
+  }
```
matt001k: Would we want to get this configuration parameter (the period at which borealis will publish new sensor data) from the node over Bristlemouth?
Other sensor types have this configuration as well, but nowhere is it used besides calculating the aggregation period  of those nodes.

src/apps/bridge/sensor_drivers/softSensor.cpp
```diff
@@ -59,7 +59,8 @@ void SoftSensor::softSubCallback(uint64_t node_id, const char *topic, uint16_t t
           printf("Updating soft %016" PRIx64 " node position, current_time = %" PRIu32
                  ", last_time = %" PRIu32 ", reading count: %" PRIu32 "\n",
                  node_id, current_timestamp, soft->last_timestamp, soft->reading_count);
-          soft->node_position = topology_sampler_get_node_position(node_id, pdTICKS_TO_MS(5000));
+          soft->node_position =
+              topology_sampler_get_node_position(node_id, pdTICKS_TO_MS(5000));
```
victorsowa12: Did you want to remove the pdTICK_TO_MS here like you did in the rbr code?

src/apps/bridge/sensorController.cpp
victorsowa12: I think we talked about this in the past and it was an "eventually we may want this". I like the idea and maybe now we have time to do it!

So in the other sensors this value is used to calculate the AVERAGER_MAX_SAMPLES which is used by their respective createSub function to allocate enough space for their aggregation buffers. The reason it is there is because sensors can have other sampling rates. For example, if the RBR is sampling at 1Hz and then we change that to 2Hz, the buffers that are the size to handle 1Hz wouldn't be able to handle the 2Hz rate of data coming from that node. So having these configs allows the bridge to allocate the appropriate sized buffers. Moving that config to the node that has the RBR makes sense to me since it would more easily allow someone to take their RBR from one Spotter to another without needing to change configs on the bridge (assuming their RBR has a non-standard sampling rate). However, I think that right now not too many people are changing the sampling rates of the sensors themselves and therefore it was safe for us to use/store the config on the bridge. One downside of both of these approaches is that the sensors and the nodes are interchangeable, so you could have an RBR node and swap the RBR and all of a sudden have one that is sampling at a different rate than what is stored in the config. So perhaps some level of auto-detection would be nice 😎!

I don't know if a moving average was considered, but that is another possible solution that would help us eliminate the need for huge buffers and therfore the need to know this reading period config. However, I'm not 100% sure on the accuracy implications of a moving average.

src/apps/bridge/sensor_drivers/softSensor.cpp
victorsowa12: Or we can use the `bm_ticks_to_ms` instead!

victorsowa12 (approved):
Looks good! Had some ideas around the config for reading period.

src/apps/bridge/sensor_drivers/softSensor.cpp
matt001k: `topology_sampler_get_node_position` is calling `pdTICKS_TO_MS` internally, so this is essentially running twice!

src/apps/bridge/sensorController.cpp
towynlin: Agreed that now's a good time to make the sensor node the source of truth and avoid the duplication of the config on the bridge. The bridge should still have a default value in code that it uses if the config we request isn't set on the sensor node, and aggregation should still behave sensibly and degrade gracefully if the sensor node doesn't do what the config suggests it should — like Victor was saying if someone swaps an RBR sensor into an existing node where the sensor is internally set to a different reading rate.

Let's plan to do this now on first pass for the new Borealis and DO nodes. I'll write up a separate story to migrate the existing nodes (soft, rbr, current meter, turbidity). Also I'll write up a separate story for bufferless aggregation using Kahan summation and Welford variance. We've talked about these many times, but it hasn't been necessary yet thanks to the STM32U5's luxurious SRAM.

towynlin (approved):

src/apps/bridge/sensorController.cpp
matt001k: @towynlin I added in the logic to obtain the sampling period for the Borealis sensor over the Bristlemouth network in my latest commit. It utilizes a new API created in `bm_core` found in the following [PR](https://github.com/bristlemouth/bm_core/pull/71) to decode config value messages.

towynlin (approved):

PR #241, opened 2025-01-03
Handle SPI Callbacks Asynchronous In Adin Callstack

matt001k:
## What changed?
Previously over at Sofar, there has been call stack traced from the Adin driver code reaching over 60 calls deep in the L2 task!
Handling the callbacks in a separate task rather than the calling task of `HAL_SpiReadWrite` will help reduce the size of the task's call stack.
The following is a peek at what the callback stack will look like when being handled from a separate task:

image

Below is what it has looked like previously:

image
image

(With even more calls to go if I continue debugging the total trace!)

The provided solution waits on a recursive notification in the new *ADIN SPI Task*.
The notification is given every time a successful SPI transfer occurs.
Then the callback set by the ADIN driver is invoked in the *ADIN SPI Task*!

Following the SPI transaction flow through `protected_spi.c`, there is no need to trigger this notification from an interrupt, or through any other source. The notification will occur after the transaction has been completed every time in the provided solution.

Reduce the risk of a stack overflow!

Look at resource allocation, I provided 512 bytes to the new task's stack, but that might be overkill.
Also, from the Adin driver, `HAL_Init_Hook` is never called, but is available in the `adi_hal.h` file.
To initialize this new task,
`HAL_Init_Hook` is utilized but called in the `bristlemouth_client.cpp` file.
Is this desired? I feel like it creates confusion.

## Checklist

### DIFF
--- a/src/lib/bm_integration/bristlemouth_client.cpp
+++ b/src/lib/bm_integration/bristlemouth_client.cpp
@@ -9,6 +9,7 @@
 #include "queue.h"
 #include "task.h"
 
+#include "adi_hal.h"
 #include "bcmp_cli.h"
 #include "bm_ports.h"
 #include "bsp.h"
@@ -51,6 +52,7 @@ static void adin_power_callback(bool on) {
 
 void bcl_init(void) {
   IORegisterCallback(adin_pins.interrupt, network_device_interrupt, NULL);
+  HAL_Init_Hook();
   config_init();
   //TODO: This should all be moved to user code
   uint8_t major, minor, patch;
--- a/src/lib/drivers/adin2111/src/adi_hal.c
+++ b/src/lib/drivers/adin2111/src/adi_hal.c
@@ -1,13 +1,32 @@
 #include "adi_hal.h"
+#include "FreeRTOS.h"
 #include "app_util.h"
+#include "bm_os.h"
 #include "bsp.h"
 #include "protected_spi.h"
+#include "semphr.h"
+#include "util.h"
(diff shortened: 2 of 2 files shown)


towynlin:
Looks like a good solution. 👍 

I'm going to talk out here a possible alternative that moves this extra task into bm_core so that future implementors don't have to work too hard to get this right. This adi_hal.c will always be required though, and this is a reasonably easy model to follow.

I guess it could work like this... Instead of calling the callback directly in adi_hal.c, we'd call some new public function from bm_core (maybe bristlemouth.h or network_device.h) and then there inside bm_core we'd notify another task to call the saved callback.

Meh, sounds like different hoops to jump through, not necessarily fewer or better hoops.

Yeah, I think I like this better because the solution will always involve a combination of elements that are beyond the purview of bm_core — the MCU-specific HAL calls to support the network device, the OS's implementation of task notifications/IPC.

Regarding resources, even though we might be able to use a smaller stack, I see that 512 is what we use by default for all the tasks unless we know they're resource-intensive. Keep it.

Regarding task priority, I see we have `ADIN_SPI_TASK_PRIORITY` defined to 15 in several places, which might be a remnant from before bm_core simplifications. I think that's probably a better priority than 30. I find at least one interrupt handler set to 20. Please use 15, and it would be best to put that priority in context with the others in a task_priority.h.

towynlin (approved):

victorsowa12 (approved):
Looks good to me!

PR #246, opened 2025-01-13
Ability To Build External Project

matt001k:
## What changed?
Updated CMake logic added to build a project located outside of the `bm_protocol` repository.
Adds an example directory that can be copied over directly to this new application space.
Updated documentation explaining to developers how to create external projects.

This allows developers using dev kits to build projects outside of the repository,
allowing for a cleaner experience.
The additional example directory adds an easy jump on spot that developers can copy over directly to this new application space.

I can picture a future where we have an arduino like IDE that developers can utilize which creates projects and configures projects all behind the scenes without the developer understanding the intricacies of CMake.
This is a step toward that future!

Documentation and testing.
Is the documentation clear?

Try building a project on your own to test!
I was doing some hydrophone testing myself with some of the new logic added to the bridge to ingest mock hydrophone messages via an external application!

The commands necessary to create a new project will be as so:

```
cd path/to/bm_protocol/
cd ..
mkdir bristlemouth_external
cd bristlemouth_external
cp -r ../bm_protocol/src/apps/bm_devkit/example/* ./
mkdir build && cd build
cmake ../../bm_protocol -DCMAKE_TOOLCHAIN_FILE=../../bm_protocol/cmake/arm-none-eabi-gcc.cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_BOOTLOADER=1 -DCMAKE_APP_TYPE=BMDK -DBSP=bm_mote_v1.0 -DAPP=path/to/bristlemouth_external -DCMAKE_VERBOSE_MAKEFILE=OFF
make -j
```

Note not all commands can be copied directly into the terminal as they require some specific paths, such as the `cmake` command.

## Checklist


towynlin (approved):
A solid step forward! Thank you!

My wish for how this could work in the future is that I could create a CMakePresets.json in the external app, which would require the initial source directory argument to the `cmake` command to be the external directory instead of bm_protocol. As this is, I have to add a new preset to bm_protocol, and essentially edit code in one window (the external app) while I build in another window (bm_protocol).

On the command line this is easy to achieve, but it doesn't integrate as well into VS Code at least and probably some other IDEs.

In any case, approved! It's so nice to allow Bristlemouth developers to keep their code in a separate repo!

matt001k:
@towynlin that sounds like a good next step! I like the idea of it a lot.

victorsowa12 (approved):

PR #247, opened 2025-01-15
Update The BSP For The Bootloader To Utilize The Actual Chipset On Motes/Bridge

matt001k:
## What changed?
Previously the BSP was referencing the NUCLEO dev kit STM32U575 which has 144 pins.
Now, reference the STM32U575ZIT6Q and remove references that are attempting to access GPIO on LEDs that do not exist.

References the true chipset that is used on the board.
If we ever need to update a pin on the bootloader BSP, this allows us to properly access the ports/pins that are on the physical hardware.

Changes are pretty trivial.
But testing is a must.
I was testing DFU updates from another node on the network through the HIL script for DFU updates.
In order to run this test, run the following commands from the root of `bm_protocol`:

```
cd src/lib/bm_core/test/scripts
pytest -s ./test_dfu.py --port /dev/cu.usbmodem000044CF8DF41 --baud 115200 --file ../../../../../build/loadcell/src/bm_mote_bristleback_v1_0-loadcell-dbg.elf.dfu.bin --node_id 0x71b84a66f73de0a7
```

Replace the arguments for the `port`, `file` and `node_id` with the port of the node you are connected to over USB, the file you are attempting to flash to the mote running the new bootloader and the ID of the node running the updated bootloader.

## Checklist


src/lib/bm_core
victorsowa12, line 1: Is this meant to be in this PR?

victorsowa12: Oh it will probably go away if develop is merged into this branch

victorsowa12 (approved):
Looks good to me!

PR #249, opened 2025-01-17
Disable Unused Ports Based Upon Configuration

matt001k:
## What changed?
Adds module to monitor the ports on a node.
If data has been received on a port, the port will stay online,
if not, disable the port, unlocking power savings.
The only way a node can utilize this is if a configuration parameter is set, which determines how long to wait until the ports are checked to see if there has been communication on each port.

This work also turns off power to port 2 on the bridge with the new L2 API `bm_l2_netif_enable_disable_port`.

This saves power on nodes by disabling open port(s) on the network device (ADIN2111).
When monitoring the power on an upstream node (via the serial console), after the an unused port is disabled there is a current savings of ~15mA.
See below for the console output:

F9B33FDA-BCFD-4CA8-AA26-528248D2586E_1_201_a

### Power Savings
The current drops from **0.013250A to 0.011750A**,
which would indicate a **36mW** power savings.

## Problems
Unfortunately on the ADIN2111, port 1 cannot be placed in software power-down mode individually, as it will bring down port 2 as well.
See [page 25 of the ADIN2111 datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/adin2111.pdf) for an in depth explanation on Software Power-Down Mode:

image

To combat this, logic has been added to notify spotter's console/log to a file that the device cannot turn off port 1 if it is the port attempting to be shut down.

Should the port monitor be something `bm_core` offers?
It could be useful in other applications.
Besides lines 49-59 this file is very generic and the very mote specific code can be brought out.
Let me know your thoughts!

## Checklist


src/lib/common/port_monitor.c
```diff
@@ -0,0 +1,107 @@
...
+#include "bm_config.h"
+#include "bm_os.h"
+#include "configuration.h"
+#include "device.h"
+#include "l2.h"
+#include "spotter.h"
+#include <inttypes.h>
+#include <string.h>
+
+#define port_monitor_time_ms_key "disableUnusedPortsTimeMs"
```
matt001k, line 11: I named the configuration key `disableUnusedPortsTimeMs` as to keep it very generic, which is not what was recommended (`adinDisableUnusedPortsMs`). I can change this easily, but if we want to make this something generic that `bm_core` uses, this might be a better name.

towynlin: Love it. Thank you. Please update the spec to match what you chose here, removing "adin" so that the same config can in the future apply to any SPE chip.

src/lib/common/port_monitor.c
```diff
@@ -0,0 +1,107 @@
...
+
+/*!
+  @brief Link change callback to set ports which have come online
+
+  @details Will set a masked value of the ports which have come online.
+
+  @param port port that has had a link change
+  @param state state of the port
+ */
+static void port_monitor_link_change(uint8_t port, bool state) {
```
towynlin: Please rename this arg (and in the comment) to indicate what representation we're talking about: `port_num`, `port_index`, or `port_mask`. Given the conversion to a mask after subtracting one I assume it's a `port_num`.

src/lib/common/port_monitor.c
```diff
@@ -0,0 +1,107 @@
...
+    uint8_t port_num = i + 1;
+    if (enabled == 0) {
+      if (port_num != 1) {
+        const char *fmt_disable_str =
+            "Disabling network port %d on node: %" PRIx64 ", it never came online\n";
+        spotter_log(0, port_monitor_log, USE_TIMESTAMP, fmt_disable_str, port_num);
+        spotter_log_console(0, fmt_disable_str, node_id(), port_num);
+        bm_l2_netif_enable_disable_port(port_num, false);
+      } else {
+        const char *fmt_err_str = "Could not disable network port %d on node: %" PRIx64
```
towynlin, line 58: Please add a comment with a brief explanation including the page number in the datasheet. Future folks will be very confused looking at this code and wondering why we compare to magic number 1.

towynlin (approved):
Small suggestions for naming and additional comments.

PR #251, opened 2025-01-23
Feature/make sure do buff accounts for subsampling

victorsowa12:
## What changed?
The DO sensor will likely send a reading very soon after power on. Therefore if subsampling is enabled we need to account for extra power ons in our aggregation buffer sizes! This PR makes sure that we take the `sampleDuration/subSampleInterval` to get the number of readings we should account for in this situation.

This makes sure that we aggregate all potential data we get from the DO and that we do not drop any of the data!

The new addittions to the pmeDissolvedOxygenSensor file.

## Checklist

## Testing
The log print for when the cfg is the same as the default - this prevents us from re-initializing the buffers if we recieve the config that is the same default the bridge is expecting:
Screenshot 2025-01-23 at 3 10 09 PM

The buffer sizes being updated for a different config get (set to 700 on the node) when subsampling is disabled:
Screenshot 2025-01-23 at 3 17 38 PM

The buffer sizes being updated for a different config get (set to 700) when subsampling is enabled:
Screenshot 2025-01-23 at 3 16 50 PM


towynlin (approved):
I think it looks good — one change request — please change all instances of `sub_sample` to the single word `subsample`. 😄

PR #252, opened 2025-01-24
Fix DFU Flashing On HIL Test

matt001k:
## What changed?
Previously the DFU update code in the HIL test only took into account how ports were enumerated on Mac, the port that ends with a 1 is the console port we want to talk to.
This updated code takes into account other operating systems who will enumerate serial ports differently, such as linux

Allows DFU to operate properly on the runner machine.
Fails the test if the DFU update fails.

Examine the output from the HIL test and ensure that it is working properly.

## Checklist


victorsowa12 (approved):
Looks good!

PR #254, opened 2025-01-27
Enable STOP1 low-power mode on BM Mote LPUART apps.

sophiediehl:
## What changed?
Enabled STOP1 low-power mode on BM Mote apps that use LPUART for sensor comms.

Save ~20-30 mW of processor power! See before and after photos:

Before change: Mote consumes 138 mW. (Note: since joulescope only measures up to 15V, and Bristlemouth bus operates at 24V, we must drop the measured voltage in half using a resistor voltage divider. Thus, the real power is double what's shown in the screenshot.)
Screenshot 2024-11-18 at 5 39 17 PM

After change: Mote consumes 112 mW.
Screenshot 2024-11-18 at 5 35 19 PM

Did I modify the right apps and BSP's? Anything you'd expect to see as part of this change, that is missing?

## Checklist


towynlin:
The `bm_mote_spi_v1_0` bsp should be converted too. Have you measured the change in power draw? Some pics from the scope would be great to see!

src/lib/common/payload_uart.cpp
```diff
@@ -80,7 +80,8 @@ bool endTransaction(uint32_t wait_ms) {
   _transactionInProgress = false;
   // If there were no writes during the transaction, don't wait for writes to complete.
   if (!_writeDuringTransaction) {
-    if (_postTxFunction != nullptr) _postTxFunction(); // Safely call the function
+    if (_postTxFunction != nullptr)
+      _postTxFunction(); // Safely call the function
```
victorsowa12: nit: looks like this used to be on one line, but using some brackets would be appreciated now that clang-format made it two lines

src/lib/common/payload_uart.cpp
```diff
@@ -93,7 +94,8 @@ bool endTransaction(uint32_t wait_ms) {
     ret_value = false; // TODO - maybe better to assert(false) here?
   }
   // Call the post-transaction function after the final transmission is fully complete
-  if (_postTxFunction != nullptr) _postTxFunction();
+  if (_postTxFunction != nullptr)
+    _postTxFunction();
```
victorsowa12: same as above, use some brackets for the if statement

victorsowa12 (commented):
Just a couple of nits, but looks good! +1 to posting some power comparison pics/charts, those would be great to see!

src/lib/common/payload_uart.cpp
sophiediehl: Done

src/lib/common/payload_uart.cpp
sophiediehl: Done

towynlin (approved):
👏 👏 👏 ⚡ ⚡ ⚡ 🎉 🎉 🎉

victorsowa12 (approved):
🤘LGTM

PR #255, opened 2025-01-29
Handle Borealis Levels Message

matt001k:
## What changed?
Added Borealis levels message parsing to bridge app.
This is in addition to the spectrum message already implemented.
Also, adds the encode/decoder created by the folks over at AOS.

Another stepping stone in getting the Borealis hydrophone integrated into Bristlemouth!

Architecture of the `borealisSensor.cpp` file.
I tried to avoid creating a new file for this message and adding more functions/code that could be handled in this single file.
Tell me your thoughts.

## Checklist


src/apps/bridge/sensor_drivers/abstractSensor.h
```diff
@@ -11,9 +11,10 @@ typedef enum abstractSensorType : uint8_t {
   SENSOR_TYPE_SOFT = 2,
   SENSOR_TYPE_RBR_CODA = 3,
   SENSOR_TYPE_SEAPOINT_TURBIDITY = 4,
-  SENSOR_TYPE_BOREALIS = 5,
-  SENSOR_TYPE_PME_DO = 6,
-  SENSOR_TYPE_PME_WIPE = 7,
+  SENSOR_TYPE_BOREALIS_SPECTRUM = 5,
+  SENSOR_TYPE_BOREALIS_LEVELS = 6,
+  SENSOR_TYPE_PME_DO = 7,
+  SENSOR_TYPE_PME_WIPE = 8,
```
towynlin, line 17: I notice the PME items here changed numbers from 6-7 to 7-8. Could be fine but please consider whether it causes any incompatibility. If it doesn't matter, then I wonder why the specific numbers are assigned.

towynlin (approved):

src/apps/bridge/sensorController.cpp
```diff
@@ -318,10 +319,18 @@ static bool node_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
...
+          Borealis_t *borealis_spectrum_sub =
+              createBorealisSensorSub(SENSOR_TYPE_BOREALIS_SPECTRUM, reply.node_id);
+          if (borealis_spectrum_sub) {
+            Borealis_t *borealis_levels_sub =
+                createBorealisSensorSub(SENSOR_TYPE_BOREALIS_LEVELS, reply.node_id);
+            if (borealis_levels_sub) {
+              abstractSensorAddSensorSub(borealis_spectrum_sub);
+              abstractSensorAddSensorSub(borealis_levels_sub);
+            } else {
+              free(borealis_spectrum_sub);
```
victorsowa12: Not sure if this file has started to use bm_os yet, but lets use `bm_free` or `vPortFree`

src/apps/bridge/sensor_drivers/abstractSensor.h
victorsowa12: These are only used within the bridge code to coordinate sensor types between the report builder, topology sampler, sensor controller and sensor subscriptions so it should be fine to change the order.

src/apps/bridge/sensorController.cpp
matt001k: Good catch!

victorsowa12 (approved):

PR #258, opened 2025-02-11
Enable STOP1 low-power mode for SOFT app.

sophiediehl:
## What changed?
Uncommented the line in SOFT app, re-enabling the STOP1 low-power mode.

Save ~25 mW on the SOFT node! See joulescope before/after shots. _Note: since joulescope max voltage is 15V, while Bristlemouth bus is 24V, I used a resistor voltage divider to divide the input voltage in half. Thus, actual power is twice the measured power._
Before: 135 mW
Screenshot 2025-02-10 at 4 29 32 PM

After: 109 mW
Screenshot 2025-02-10 at 4 27 46 PM

Did I miss anything I should be adjusting, to allow SPI to safely operate when coming in and out of LPM?

## Checklist

### DIFF
--- a/src/apps/bm_soft_module/app_main.cpp
+++ b/src/apps/bm_soft_module/app_main.cpp
@@ -405,7 +405,7 @@ static void defaultTask(void *parameters) {
   user_code_start();
 
   // Re-enable low power mode
-  // lpmPeripheralInactive(LPM_BOOT);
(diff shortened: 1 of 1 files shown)


victorsowa12 (approved):

towynlin (approved):
Woot!

PR #259, opened 2025-02-12
Remove 10 ms delay between handling incoming bytes in serial_bridge

campbellaos:
This avoids incurring the `bm_delay(10)` within the caller of `loop()` between all incoming bytes, which would add latency and artificially limit the effective throughput to 100 byte/s regardless of baud rate

## What changed?
`if (PLUART::byteAvailable())` is replaced with `while (PLUART::byteAvailable())` within bm_devkit/serial_bridge/user_code/user_code.cpp

The serial_bridge app will be more reactive and not artificially limit the throughput to 100 byte/s.

Verify there were no other prior factors which make this change redundant

## Checklist

- [x] Independently test on hardware

### DIFF
--- a/src/apps/bm_devkit/serial_bridge/user_code/user_code.cpp
+++ b/src/apps/bm_devkit/serial_bridge/user_code/user_code.cpp
@@ -37,7 +37,7 @@ static void reset() {
 }
 
 void loop() {
-  if (PLUART::byteAvailable()) {
+  while (PLUART::byteAvailable()) {
     const uint8_t b = PLUART::readByte();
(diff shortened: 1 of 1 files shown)


campbellaos:
I have tested this change on the mote, with an app that sends it cobs packets as fast as the baud rate allows, and confirmed that in the absence of this change, "WARN payload uart user byte stream buffer full!" is hit almost immediately, whereas with this change, all messages are handled and no other logic is blocked.

matt001k (approved):

victorsowa12 (approved):

towynlin (approved):

PR #260, opened 2025-02-12
Improve Handshaking Between Bridge And Spotter

matt001k:
## What changed?
Adds locking semaphore to control when messages can be sent/received
If a message reception is in progress, lock transmission
Places all transmissions into a queue and handles this within a separate task

Reduces the amount of dropped serial packets between spotter and bridge.
This was tested with DFU updates from spotter utilizing the `bridge dfu` command.
A mote sending data at a high rate (200+ bytes every second) to the `SENS_IND` file location on spotter was used to overload the serial communication.

The following message was monitored for: `[BM_DFU] [ERROR] Chunk timeout, retrying`.

Previously before this update, this message would show up 1-4 times if a DFU was triggered within the first 30 seconds of spotter being turned on.

I used FreeRTOS API here, should I use bm_core RTOS API?
I used the same task priority  `NCP_TASK_PRIORITY` for the `ncpTxTask` task that is being used for the`ncpRXTask`. Should I create another task priority? I feel like these tasks go hand in hand and deserve the same priority but would love other thoughts.

## Checklist


towynlin (approved):
Awesome!

You said

> Previously before this update, this message would show up 1-4 times if a DFU was triggered within the first 30 seconds of spotter being turned on.

What happens now with this change? Zero occurrences? Or fewer at least?

- Keep the FreeRTOS usage. We can make a later task to migrate this whole file. I don't want to conflate the two types of changes.
- Using `NCP_TASK_PRIORITY` for both TX and RX seems fine, especially with the semaphore and the yield call. They're cooperative.

Why did you change the priority from 6 to 12?

matt001k:
@towynlin to answer your questions:

> What happens now with this change? Zero occurrences? Or fewer at least?

Within my initial tests (10 DFU updates), I saw zero occurrences!

> Why did you change the priority from 6 to 12?

An initial thought I had when starting this work was that the `ncpRXTask` was not being serviced quick enough and data in the `ncpRXBuff` was being filled and overwritten by new messages from Spotter. The increase in task priority did not solve dropping messages, but it also has not hurt anything. I can change it back if there are any objections!

towynlin:
The only priority switch that looks significant is whether NCP is above or below L2 TX. Since Spotter comms has been our main bottleneck, I'd recommend keeping NCP higher at 12.

victorsowa12 (approved):
Looks good!

PR #261, opened 2025-02-19
Baud rate negotiation

matt001k:
## What changed?
Adds baud rate negotiation between spotter and bridge.
Adds baud rate discovery to bridge.
The goal here is to increase the speed of the serial communication between spotter and bridge.
In order to preserve backwards/forwards compatibility between both spotter and bridge, this feature ensures that no matter what the version spotter and bridge are on, both devices can communicate to one another.

This increases the speed at which the bridge app can communicate to spotter.

## Feature Overview
In order to increase the baud rate and avoid breaking changes between spotter and bridge firmware, a baud rate negotiation tactic can be utilized to determine if one of the devices can communicate at a higher baud rate.

Baud rate negotiation requires the following:

- A requestor
- A replier

The requestor will request a baud rate the other device is to be on.

The replier will reply to that message before changing its baud rate to the requested baud rate to indicate it can handle the request properly.

The following flow charts depict how this operation is to work:

baud_rate_negotiation

For this to work properly, there needs to be one device in charge of negotiating the baud rate initially. Because spotter waits for bridge to begin communication, bridge is the best option.

There will be 3 acceptable baud rates on each device (based upon the original baud rate set between the devices and the available `bridge baudrate` CLI command from spotter):

- 57600
- 115200
- 921600

With the above information, bridge will attempt to run a baud rate discovery protocol upon boot.

Bridge will:

1. Set the baud rate to the initial baud in the acceptable baud rates (57600)
2. Send a request for a higher baud rate (921600)
    1. If spotter does not respond a timeout will occur on bridge, meaning the request is not replied to
    2. If the request is not replied to the next baud rate in the acceptable baud rates will be set and the request will be attempted again
3. Spotter will reply at the discovered baud rate
4. Update baud rate to what was requested (921600)

Spotter will:

1. Receive request from bridge
2. Send a reply in the original baud and increase the baud rate immediately after the message is sent

The following is a flow chart of how this operation works:

bridge_baud_discovery

The largest chunk of changes are in `ncp_uart.cpp`, this would be a good place to start. There is a lot of additional handshaking occurring with semaphores to ensure messages are not sent when there is a negotiation ongoing. If you can think of a better cleaner way to handle some of this handshaking please let me know!
Also please ask questions, there is a lot going on and I want to make sure all concerns are addressed. 

## Checklist


src/lib/bm_ncp/ncp_uart.cpp
```diff
@@ -446,6 +559,41 @@ static void ncpRXProcessor(void *parameters) {
...
+    // Send 0 to reset receiving buffer on spotter and reset our buffers just
+    // incase anything was sent at a different baud rate
+    ncpRXCurrBuff = 0;
+    ncpRXBuffIdx = 0;
+    bm_serial_tx(BM_SERIAL_ACK, &null_buf, sizeof(uint8_t));
+    printf("Could not find a baudrate to communicate to spotter with\n");
+  }
+
+  vSemaphoreDelete(ncp_baud_rate_discovery_lock);
+  ncp_baud_rate_discovery_lock = NULL;
```
towynlin:
```suggestion
  // Prevent race condition with semaphore usage in other tasks
  SemaphoreHandle_t tmp_sem = ncp_baud_rate_discovery_lock;
  ncp_baud_rate_discovery_lock = NULL;
  vSemaphoreDelete(tmp_sem);
```

My suggestion fixes a race condition here with the usage of `ncp_baud_rate_discovery_lock` in interrupts or other tasks. If we delete the semaphore, and then the scheduler stops this task before the static variable is set to NULL, we could call `xSemaphoreGive(ncp_baud_rate_discovery_lock)` on the deleted semphore (whose memory has been freed) in either `baud_rate_reply_cb` or `baud_rate_negotiate_timer_handle`.

towynlin (commented):
- We should measure how long negotiation takes when successful and set timeouts appropriately. I bet it's a lot faster than 250ms.
- I pointed out a race condition in ncp_uart.cpp.
- Do we handle bridge restarts well that are not at spotter boot time? A quick test would be to set a config so that bridge reboots, or just tap bridge reset button if accessible.

A comment for a possible future improvement, not in this PR: Since the most common use case here is that the same baud rate will be negotiated every single time, we could save the last negotiated baud rate to prevent having to negotiate on every boot. It might require a fair amount of code that isn't simple and might not save us much time at boot.

src/lib/bm_ncp/ncp_uart.cpp
matt001k: Ahhh very good call out! I will go ahead and make that update, thanks!

matt001k:
@towynlin to answer your question:

>- Do we handle bridge restarts well that are not at spotter boot time? A quick test would be to set a config so that bridge reboots, or just tap bridge reset button if accessible.

We do! But we start at the lowest acceptable baud rate when beginning discovery on the bridge (57600), to decrease latency on boot, a good change would be to start at the highest baud rate and work out way down during discovery, lets switch the order in which we iterate through `acceptable_bauds`.

matt001k:
@towynlin to answer this question:
> - We should measure how long negotiation takes when successful and set timeouts appropriately. I bet it's a lot faster than 250ms.

I added in some logic to evaluate the negotiation time, see the diff below:

```
@@ -327,7 +327,11 @@ static bool baud_rate_request_cb(uint32_t baud) {
   return ret;
 }
 
+#include "spotter.h"
 static bool baud_rate_reply_cb(void) {
+  uint32_t negotiation_time =
+      NCP_BAUD_RATE_NEGOTIATE_TIMEOUT_MS -
+      pdTICKS_TO_MS(xTimerGetExpiryTime(ncp_ctx.negotiate_timer) - xTaskGetTickCount());
   xTimerReset(ncp_ctx.negotiate_timer, 0);
   xTimerStop(ncp_ctx.negotiate_timer, 0);
   serialSetBaudRate(ncpSerialHandle, ncp_ctx.requested_baud);
@@ -336,6 +340,7 @@ static bool baud_rate_reply_cb(void) {
     xSemaphoreGive(ncp_baud_rate_discovery_lock);
   }
   xSemaphoreGive(ncp_ctx.negotiating_lock);
+  printf("Negotiation Time %" PRIu32 "ms\n", negotiation_time);
 
   return true;
}
```

After a few test runs  (tested 5 times at each baudrate), the evaluated times are as follows:

- 2ms at 921600
- 3-4ms at 115200
- 4-5ms at 57600

 I will update the timeout to 10ms to provide enough overhead for slower baudrates and any potential hiccups.

victorsowa12 (approved):
👏👏👏

PR #264, opened 2025-02-25
Increase TSYS01 conversion time delay

victorsowa12:
## What changed?
Increased the TSYS01 ADC conversion wait time.

It will decrease the number of missed readings on sensors that may have slightly slower conversion times.

## Checklist

### DIFF
--- a/src/apps/bm_soft_module/user_code/user_code.cpp
+++ b/src/apps/bm_soft_module/user_code/user_code.cpp
@@ -169,8 +169,8 @@ void loop(void) {
   }
   }
 
-  // Delay between readings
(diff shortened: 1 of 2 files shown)


matt001k (approved):

victorsowa12:
@matt001k Your comment in the other PR about blocking made me think of subtracting the conversion time in this loop so we stay at the proper configured delay. Do you think its a good idea?

matt001k:
@victorsowa12 that looks good to me and ensures that we will be delaying the proper amount of time!

victorsowa12:
Tested on hardware and is working!

PR #266, opened 2025-02-27
Adds Power Information Timing Service To Bridge App

matt001k:
## What changed?
Adds the power info timing service to the bridge power controller

Reports timing information for:

- Initial time of bridge (two minute initialization period)
- Bridge power controller with no subsampling
- Bridge power controller with subsampling
- Bridge power controller disabled

This helps informs nodes downstream from bridge how long they will have to complete critical operations!
Saving/writing to files, performing atomic operations, etc.

Looking for any edge cases where we could report incorrect information (`remaining_on_s` being the most important item here, found a race condition if `_timebaseSet` is `false` after initialization but was able to fix that).

When the power controller is disabled, the bridge will report `UINT32_MAX` for both `total_on_s` and `remaining_on_s` and `0` for `upcoming_off_s`. Does this create confusion?

## Testing

In order to test this PR a mote was requesting power timing information from this service every second.
There were 3 tests performed:

1. With subsampling enabled with the following config parameters:
    ```
    sampleIntervalMs u 600000
    sampleDurationMs u 300000
    subsampleIntervalMs u 90000
    subsampleDurationMs u 60000
    ```
2. With subsampling disabled and power controller enabled with the following config parameters:
    ```
    sampleIntervalMs u 300000
    sampleDurationMs u 150000
    ```
3. With the power controller disabled

Below is a screen capture from a serial console of the mote requesting this information every second showing the time for `remaining_on_s` decrementing:

image

The following output shows what the service reply from bridge is during the 2 minute initialization period:

```
Subscribing to Topic: bus_power_controller/timing/rep
Service: bus_power_controller/timing
Reply: 
 * total_on_s: 120
 * remaining_on_s: 113
 * upcoming_off_s: 480
```

1. Power controller enabled with subsampling enabled
    Output from a subsample interval/duration:
    
    ```
    Subscribing to Topic: bus_power_controller/timing/rep
    Service: bus_power_controller/timing
    Reply: 
     * total_on_s: 60
     * remaining_on_s: 6
     * upcoming_off_s: 30
    ```
    
    Output from end of subsample duration (end of sample duration):
    
    ```
    Subscribing to Topic: bus_power_controller/timing/rep
    Service: bus_power_controller/timing
    Reply: 
     * total_on_s: 60
     * remaining_on_s: 12
     * upcoming_off_s: 270
    ```

2. Power controller enabled without subsampling enabled
    Output from a sample duration:
    ```
    Subscribing to Topic: bus_power_controller/timing/rep
    Service: bus_power_controller/timing
    Reply: 
     * total_on_s: 150
     * remaining_on_s: 50
     * upcoming_off_s: 150
    ```

3. Power controller disabled
    Output indefinitely:
    ```
    Reply: 
     * total_on_s: 4294967295
     * remaining_on_s: 4294967295
     * upcoming_off_s: 0
    ```

## Checklist


src/apps/bridge/bridgePowerController.cpp
```diff
@@ -304,6 +307,57 @@ void BridgePowerController::powerControllerRun(void *arg) {
...
+
+  if ((!power_controller->_initDone || !power_controller->_timebaseSet) &&
+      power_controller->isPowerControlEnabled()) {
+    static const uint64_t init_uptime_s = 0;
+    uint64_t current_uptime_s = uptimeGetMs() / 1000;
+
+    // Calculate how long the bus will be on before initialization is done
+    d.total_on_s = INIT_POWER_ON_TIMEOUT_MS / 1000;
+    d.remaining_on_s = timeRemainingGeneric(init_uptime_s, current_uptime_s, d.total_on_s);
+    d.upcoming_off_s = power_controller->_sampleIntervalS - d.total_on_s;
```
victorsowa12: I supppose this will not always be true based on the alignment timing. Although we won't know the actual upciming off time until the init period is done and we calculate the alignment. Is that going to be okay for most use cases?

victorsowa12 (approved):
Looks good!

src/apps/bridge/bridgePowerController.cpp
matt001k: That's a good point, do you think it would be a good move to call `_alignNextInterval(power_controller->getCurrentTimeS(), 0, power_controller->_sampleIntervalS)` on line 322?

matt001k: Internal discussion with the Sofar team. We will utilize `UINT32_MAX` as a undefined value for situations such as this in regards to the power info timing service.

PR #270, opened 2025-03-10
Allow debug_adin_raw module to initialize adin properly

matt001k:
## What changed?
Configure power callback for `adin init` command line call in the bringup app.
Also start the task for the network spi traffic when `adin init` is invoked.

This allows the `adin init` call to work properly without failing in the bringup app.

Implementation in the `adin init` CLI call.
Should we just make a call to `bcl_init`  (that may defeat the purpose)?

## Checklist

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -111,7 +111,7 @@ endif()
 if(APP_NAME MATCHES "bringup")
     list(APPEND LIB_FILES
         ${LIB_DIR}/debug/debug_uart.c
-        ${LIB_DIR}/debug/debug_adin_raw.c
+        ${LIB_DIR}/debug/debug_adin_raw.cpp
     )
 endif()
 
--- a/src/lib/bm_integration/bristlemouth_client.cpp
+++ b/src/lib/bm_integration/bristlemouth_client.cpp
@@ -40,7 +40,7 @@ static bool network_device_interrupt(const void *pinHandle, uint8_t value, void 
   return bm_l2_handle_device_interrupt() == BmOK;
 }
 
-static void adin_power_callback(bool on) {
+void bcl_power_callback(bool on) {
   IOWrite(&ADIN_PWR, on);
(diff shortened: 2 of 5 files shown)


victorsowa12 (commented):
The changes are looking good. Have you run it through the BoN test where Uma found the bug?

I believe changing to use `bcl_init` would required BoN script changes since they use the raw adin CLI API. Unless we can use `bcl_init` and the raw adin CLI in parallel? Altough, I'm not sure how we would be able to recieve raw packets in that case.

matt001k:
@victorsowa12 I have ran the tests at my desk but I will reach out to Uma and see if we can run through it together or if she can async!

towynlin (commented):
Looks good to me, but like Victor said, the proof is in whether the bed-of-nails script and fixture work as intended

uma-kat (commented):
I did a test on a Bristlefin board stack I have and `adin init` is successful 👍 

```
info
APP_NAME: bringup_mote
UID: 203037484d30500800400055
MAC: 00:00:3d:ba:16:ef
FW Version: bringup_mote@ENG-v0.13.0-rc.3-8-g0ec98ad0+f6b82753
GIT SHA: 0EC98AD0
Build ID: f6b82753ad39476ce96a6209c3886b2469d1944d
BSP: mote_v1_0
Reset Reason: Bootloader reset
Node ID: 019285b23dba16ef
Bootloader Information:
  Version: v0.12.2
  SHA: FC385392
  Signature support: 0
  Encrytion support: 0
  Swap Type: NONE
info end
adin init
Adin initialized successfully
```

victorsowa12:
Talking about in person, this PR looks good for initializing the adin again, but the `debug_l2_rx` function which is used by the raw adin to recieve is now no longer being used. So we are exploring how to add it back in. Being able to recieve is checked in the B.O.N.

matt001k:
Closing until further work is done on the RX side.

PR #273, opened 2025-03-24
Optimize Bridge To Spotter Transactions

matt001k:
## What changed?
Update the max baudrate from 921600 to 1000000.
NCP UART handshaking semaphores moved from RX ISR to the GPIO ISR.
Create separate macros for TX and RX queue sizes

Increased Baud Rate (921600 to 1000000):
- Spotter is able to set this baudrate more accurately than 921600, as there is a 2.1% error when setting the baud to 921600 on Spotter
Handshaking semaphore changes:
- Creates more readable code and opens the opportunity to utilize DMA on the RX line on the bridge
Separate TX/RX queue macros:
- Increases readability

The largest change is the baud rate increase.

## Checklist

### DIFF
--- a/src/lib/bm_ncp/ncp_uart.cpp
+++ b/src/lib/bm_ncp/ncp_uart.cpp
@@ -30,7 +30,8 @@ extern "C" {
 
 #define NCP_NOTIFY_BUFF_MASK (1 << 0)
 #define NCP_NOTIFY (1 << 1)
-#define NCP_PROCESSOR_QUEUE_DEPTH (16)
+#define NCP_PROCESSOR_RX_QUEUE_DEPTH (16)
+#define NCP_PROCESSOR_TX_QUEUE_DEPTH (16)
 #define NCP_BAUD_RATE_NEGOTIATE_TIMEOUT_MS (10)
 
 static uint32_t ncpRXBuffIdx = 0;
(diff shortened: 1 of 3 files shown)


src/lib/bm_ncp/ncp_uart.cpp
```diff
@@ -393,15 +401,15 @@ void ncpInit(SerialHandle_t *ncpUartHandle, NvmPartition *dfu_partition,
...
   configASSERT(ncp_processor_queue_handle);
 
-  ncp_tx_queue_handle = xQueueCreate(NCP_PROCESSOR_QUEUE_DEPTH, sizeof(ProcessorQueueItem_t));
+  ncp_tx_queue_handle =
+      xQueueCreate(NCP_PROCESSOR_TX_QUEUE_DEPTH, sizeof(ProcessorQueueItem_t));
   configASSERT(ncp_tx_queue_handle);
 
   ncp_serial_lock = xSemaphoreCreateBinary();
   configASSERT(ncp_serial_lock);
-  xSemaphoreGive(ncp_serial_lock);
```
victorsowa12, line 404: Does not giving the lock here mean that we cannot tx until we have recieved something, thus triggering the release of the lock in the GPIO ISR? If so is that okay?

victorsowa12 (approved):
Looks good! I had one question about not giving the lock on init.

src/lib/bm_ncp/ncp_uart.cpp
matt001k: This lock is only taken when the `BM_INT` line is active (low) and given when it is inactive (high). For further reference see the new variable added `ncp_rx`. Thus, this is not necessary to give here.

PR #278, opened 2025-04-14
Added running variance calculation and report of stdev to remote LC messages

NicholasIanMerritt:
## What changed?

Added running variance calculation and report of stdev to the remote …

better lc data

Any breaking changes I made? 

## Checklist

### DIFF
--- a/src/lib/sensor_sampler/loadCellSampler.cpp
+++ b/src/lib/sensor_sampler/loadCellSampler.cpp
@@ -50,6 +50,11 @@ static bool loadCellSample() {
   float calFactor = _loadCell->getCalibrationFactor();
   int32_t zeroOffset = _loadCell->getZeroOffset();
 
+  float running_mean = 0;
+  float running_var = 0;
+  float old_mean = 0;
+  float stdev = 0;
+
   RTCTimeAndDate_t timeAndDate;
   char rtcTimeBuffer[32];
   if (rtcGet(&timeAndDate) == pdPASS) {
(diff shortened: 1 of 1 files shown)


src/lib/sensor_sampler/loadCellSampler.cpp
```diff
@@ -97,17 +109,33 @@ static bool loadCellSample() {
...
+            max_force, min_force, stdev, cellular_send_read_counter);
+
+    char data_string[200];
+    memset(data_string, 0, sizeof(data_string));
+
 
-    sprintf(data_string, "mean force: %f |  max force: %f  | min force: %f", mean_force,
-            max_force, min_force);
+    sprintf(data_string, "Loadcell  |  mean force: %f |  max force: %f  | min force: %f  | stdev: %f  | reading count: %d", mean_force,
+            max_force, min_force, stdev, cellular_send_read_counter);
```
matt001k, line 139: I recommend using `snprintf` here to avoid a buffer overflow of `data_string`. There also is some optimization if you make `data_string` a static variable. This way, the memory is allocated at compile time rather than every time this function is ran, saving you processing cycles and task memory usage.

src/lib/sensor_sampler/loadCellSampler.cpp
```diff
@@ -50,6 +50,11 @@ static bool loadCellSample() {
   float calFactor = _loadCell->getCalibrationFactor();
   int32_t zeroOffset = _loadCell->getZeroOffset();
 
+  float running_mean = 0;
+  float running_var = 0;
+  float old_mean = 0;
+  float stdev = 0;
```
matt001k, line 56: These values will be reset to zero every time this function is ran. To avoid this create a variable that will be `static`. That way the value is retained every time this function is ran. Note how `max_force`, `min_force` and `mean_force` are all static variables.

matt001k (changes requested):

PR #279, opened 2025-04-15
Add Handshaking When Determining A Nodes System Info

matt001k:
## What changed?
Handshake added between subsequent `sys_info_service_requests` in topology_sampler's topology callback.

Ensures that a system info request is received from nodes in order, this way nodes cannot receive info replies out of order. 

Adds print statements every time a topology request is made from topology sampler to ensure topology is not being reported out of order.

## Bug Explanation

There are one of two situations I can see occurring here:

1. The topology is being calculated incorrectly
2. There is a timing condition when obtaining the `sys_info` on a node in the topology (see below for details)

Big AHA moment when running through the logic for how this is handled.
Within the `topology_sampler` in the callback for the topology report here is how the flow of things work currently:

1. Iterate through the network topology (in order) and send a `sys_info_service_request`
    a. I thought this was a little interesting as we do not wait for a reply from the request, we immediately send the request to each node back to back
    b. In the topology module, we wait until we hear back from the each node after a neighbor table request before further nodes to the topology
2. Iterate through the NUMBER of nodes (not in order), wait for a `sys_info_reply` message to fill a queue item
    a. Ok this is where things can get out of order, if a `sys_info_reply` is received on one node before another, bad things will happen.
3. Send a `config_cbor_map_service_request` to the node we just received the `sys_info_reply` from
    a. This is done sequentially, we send a `config_cbor_map_service_request` in accordance the the node that we first received the `sys_info_reply` from
4. CBOR encode all this data, run a crc32 over  it, compare it against the last calculated crc32, if it is correct move on, if it is incorrect store the new config map and report it up to spotter.

Ok...
a lot going on.
From the information above, it seems like there is a timing condition with the `sys_info` request being sent to the nodes in the topology, but can we prove this?
Yes!

Tracing the variable `_node_list.nodes`  will help lead us to the promised land.
This variable is assigned based on the incoming topology.
`_node_list.nodes[0]` will be assigned to the first node in the topology, `_node_list.nodes[1]` will be assigned to the next node and so on.
This variable is used in the function: `topology_sampler_get_node_position`.
What calls this function?
All of the `sensorSubCallbacks` in the `src/apps/bridge/sensor_drivers` directory!
These callbacks are used to pipe data to the `XXXX_SENS_IND.csv` file on Spotter.
What does this mean you ask?
Well, if the topology report is correct, then the `node_positions` in the `XXXX_SENS_IND.csv` file will be correct, and they are!
This means that there *has* to be some timing issue in the sequencing of the `sys_info_request` messages, this can be remedied by ensuring that there is proper handshaking occurring here, ex a semaphore waiting until the previous node's `sys_info_service_reply` was received before moving onto the next.

Evidence:
[0001_BRIDGE_CFG.log](https://github.com/user-attachments/files/19763419/0001_BRIDGE_CFG.log)
[0001_SENS_IND.csv](https://github.com/user-attachments/files/19763455/0001_SENS_IND.csv.zip)

The evidence is a good place to start! Other than that, I want feedback on the new bridge prints added, these are for debugging to see if we ever get the topology out of order. I am fairly confident we will not, but it still could be a nice to have.

## Checklist

### DIFF
--- a/src/apps/bridge/topology_sampler.cpp
+++ b/src/apps/bridge/topology_sampler.cpp
@@ -68,6 +68,7 @@ static bool _send_on_boot;
 static node_list_s _node_list;
 static QueueHandle_t _sys_info_queue;
 static QueueHandle_t _config_cbor_map_queue;
+static SemaphoreHandle_t _sys_info_handshake;
 
 static void topology_timer_handler(TimerHandle_t tmr);
 static void topology_sampler_task(void *parameters);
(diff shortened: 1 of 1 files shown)


victorsowa12 (approved):
Looks good!!

PR #280, opened 2025-04-17
Bridge Utilizing DMA for USART3 (both tx/rx)

victorsowa12:
## What changed?
This gets the Bridge to use DMA with USART3 for the communications with Spotter.

Here we can see both TX and RX corectly using DMA. We can see that there is just one `USART3_IRQHandler` call after an RX event on the bridge, and one `USART3_IRQHandler` call at the end of the TX. We can see that the bridge receives a `bm info 0` and responds!

Screenshot 2025-04-18 at 1 25 20 PM

Here is a successful DFU update of a bridge while it is using DMA.
Screenshot 2025-04-24 at 10 19 47 AM

TODO:
✅done - update the bsp stm32cubemx version (this will auto update a bunch of files so I will do this after reviews and test everything again)

This will decrease the latency and CPU usage by the bridge when talking to the Spotter.

Making sure the tx lock does not have any race conditions that would cause a bad locked state.

## Checklist


src/bsp/bridge_v1_0/bsp.cpp
```diff
@@ -69,9 +69,9 @@ void bspInit() {
 
 void mxInit(void) {
   MX_GPIO_Init();
+  MX_GPDMA1_Init();
```
victorsowa12, line 72: DMA needs to be enabled before USART3 for things to be initialized properly. Otherwise the interrupt after a "DMA till IDLE" interrupt will not be triggered.

src/lib/bm_ncp/ncp_uart.cpp
```diff
@@ -25,28 +25,25 @@ extern "C" {
...
+#include "stm32u5xx_hal_uart.h"
+
+static constexpr uint8_t NCP_PROCESSOR_RX_QUEUE_DEPTH = 16;
+static constexpr uint8_t NCP_PROCESSOR_TX_QUEUE_DEPTH = 16;
+static constexpr uint8_t NCP_BAUD_RATE_NEGOTIATE_TIMEOUT_MS = 10;
+static constexpr uint8_t NCP_RX_BUFF_COUNT = 2;
+static constexpr uint16_t NCP_BUFF_LEN = 2048;
+static volatile uint8_t ncpRXCurrBuff = 0;
+static uint8_t ncpRXBuff[NCP_RX_BUFF_COUNT][NCP_BUFF_LEN];
+static uint32_t ncpRXBuffLen[NCP_RX_BUFF_COUNT];
```
matt001k: These two variables can be made volatile as they have the chance to be changed outside of the normal flow of execution (ie an interrupt)!

src/lib/bm_ncp/ncp_uart.cpp
```diff
@@ -108,12 +104,12 @@ static bool queue_tx(const uint8_t *buff, size_t len, bm_serial_message_t messag
     memcpy(tmp_buf, buff, len);
     if (xQueueSend(ncp_tx_queue_handle, &q_msg, pdMS_TO_TICKS(10)) != pdTRUE) {
       vPortFree(tmp_buf);
-      printf("Could not queue buffer for NCP message transmission");
+      printf("Could not queue buffer for NCP message transmission\n");
```
matt001k, line 107: Nice catch!

src/lib/bm_ncp/ncp_uart.cpp
```diff
@@ -422,7 +411,7 @@ void ncpInit(SerialHandle_t *ncpUartHandle, NvmPartition *dfu_partition,
 
   // Create the task
   BaseType_t rval = xTaskCreate(ncpTxTask, "NCP_TX", configMINIMAL_STACK_SIZE * 3, NULL,
-                                NCP_TASK_PRIORITY, &ncpRXTaskHandle);
+                                NCP_TASK_PRIORITY, NULL);
```
matt001k, line 415: Also nice catch 😁

src/lib/bm_ncp/ncp_uart.cpp
```diff
@@ -609,63 +607,64 @@ static void ncpBaudRateDiscovery(void) {
...
 }
 #endif
 
-// cppcheck-suppress constParameter
-static BaseType_t ncpRXBytesFromISR(SerialHandle_t *handle, uint8_t *buffer, size_t len) {
-  (void)handle;
+// HAL_UARTEx_RxEventCallback is a weakly defined function in the STM32 UART HAL.
+// Here we are overriding the weak function to handle the RX event.
+// This function called from within HAL_UART_IRQHandler function,
+// which is called above in the USART3_IRQHandler function.
```
matt001k, line 623: Might be worth noting here too that it can get called from the DMA IRQ as well.

matt001k (approved):
Code looks great! A couple of comments. Nice work :)

src/lib/bm_ncp/ncp_uart.cpp
victorsowa12: Done!

src/lib/bm_ncp/ncp_uart.cpp
victorsowa12: Added!

victorsowa12:
I have updated the package versions of the cubemx project for the bridge. 

TX/RX is still working and I was able to do a DFU update of the bridge successfully!

PR #281, opened 2025-04-23
Low Data Rate BOREALIS Message

matt001k:
## What changed?
Added Level Statistics message from BOREALIS to sensor report builder on the bridge app.
The Level Statistics message is also logged to the `SENS_AGG` file on Spotter.

Stepping stone in getting data from a smart mooring with a BOREALIS sensor to the sofar backend.
Informs user if their BOREALIS unit is not configured properly based on Bridge and BOREALIS configurations:
- Bridge
  - If subsampling is enabled warn user that only the last collected levels statistic message is used
  - If samples per report is greater than 2 warn user
  - If sampling duration is too short to be able to collect a levels statistic message warn user
- BOREALIS
  - If number of level statistic messages (PSPLS) being reported to the bridge are greater than 1 per sample duration warn user
  - If number of level statistic messages being reported to bridge are zero per sample duration warn user
Reduce some copy-pasta code in `reportBuilder.cpp`.

Are there any situations where the bridge will force-ably update configurations on itself or BOREALIS?

Reviewing to make sure that all potential use cases are covered here. Please see testing below.

## Testing
There are multiple conditions that can occur when a BOREALIS sensor is connected to a smart mooring in regards to this PR:

1. The power controller is disabled and sensor aggregation reports are not enabled
2. Aggregated sensor reports are sent to Spotter at an interval of 2 per report (or less)
  a. This is the most ideal situation, as an extended Levels Statistics report will fit into a single Spotter Iridium message (340 bytes)
3. Aggregated sensor reports are sent to Spotter with more than 2 intervals per report
  a. Due to the amount of data that is reported on an extended Levels Statistic message, this is not desired as it will not fit in the 340 byte message limit
  b. Should we force this to be 2 if a BOREALIS system is on the network?
4. Aggregated sensor reports are sent to Spotter and the sample duration on the Bridge's power controller is less than the the amount of time for a level statistics message to be calculated from BOREALIS
  a. This is defaulted to 5 minutes on BOREALIS but is configurable, ref: `report_interval`
5. Aggregated sensor reports are sent to Spotter and the sample duration for the bridge's power controller is greater than twice the amount of time for a levels statistics message to be calculated from BOREALIS
  a. This is defaulted to 5 minutes on BOREALIS but is configurable, ref: `report_interval`

Bridge will handle the above situations respectively:

1. Nothing is sent to spotter (default behavior as it is right now)
2. A single levels statistic message is packed into the sensor report
  a. The following is an example:
      ```
      Aggregation period done!
      [BM_COMMON_AGG] e1bd3457f9bf9d8a,1,borealis,1745520970.386,305,0.983,299.827,15.047,16,lZmfoKysra23t7e3ra6urpOVoqWrq66ulZqpqpSUl5mTlJmbkZSboI6TmpmOlZ2hjZigoYeNlJWHjZSUjpObnIiNk5KGiIuKi4yNjJaWl5eYmZybjY6QkI6Oj4+SkpKSmJiYmJCQkJA=
      Adding sample for e1bd3457f9bf9d8a to list
      Incrementing sample counter to 1
      Found data for node e1bd3457f9bf9d8a adding it to the the report
      Clearing the list
      Clearing the sample counter
      ```
3. The user is warned that the number of intervals per report might breach the 340 byte limit
  a. The following is an example from Spotter's console of the output from bridge when a BOREALIS unit is in the network and the `samplesPerReport` is higher than 1:
      ```
      2025-04-24T20:14:02.722Z [BRIDGE_SYS] [WARNING] Number of configured samplesPerReport higher than recommended for BOREALIS system there may be issues upon reporting data remotely: Configured - 3, Recommended - 2
      ```
4. NaN data is packed into the report (5 bytes):

      a. The following is an example from the spotter console when a device is configured to reach this condition:
      ```
      2025-04-24T22:07:38.062Z [BRIDGE_SYS] [WARNING] Number of BOREALIS level statistics messages is zero per reporting period on e1bd3457f9bf9d8a, level statistics data will not be reported remotely, please change report_interval: Configured - 300.000, Recommended - 57.000
      ```

      b. The following is an example from the bridge console when the NaN data is added to the report:
      ```
      Aggregation period done!
      Incrementing sample counter to 1
      No data for node e1bd3457f9bf9d8a in report period, adding it to the list
      Clearing the list
      Clearing the sample counter
      ```
5. The last levels statistic message will be reported
      a. The following is an example from the spotter console when a device is configured to reach this condition:
      ```
      2025-04-24T22:18:33.687Z [BRIDGE_SYS] [WARNING] Number of BOREALIS level statistics messages higher than recommended on e1bd3457f9bf9d8a, multiple level statistics messages cannot be reported remotely, only the las      t level statistics message will be reported, please change report_interval: Configured - 150.000, Recommended - 350.000
      ```

## Checklist


src/apps/bridge/reportBuilder.cpp
```diff
@@ -327,6 +250,26 @@ void ReportBuilderLinkedList::addSampleToElement(report_builder_element_t *eleme
...
+    }
+    // Copy the sensor data into the elements array in the correct location within the buffer
+    // If it is NULL then just fill it with NAN again
+    if (sensor_data != NULL) {
+      memcpy(dst, sensor_data, size);
+    } else {
+      memcpy(dst, nan, size);
+    }
+    element->sample_counter++;
+  }
```
matt001k: Consolidated code here to reduce copy-pasta.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -868,6 +828,12 @@ static void report_builder_task(void *parameters) {
                         break;
                       }
                       }
+                      if (size) {
+                        _ctx._reportBuilderLinkedList.findElementAndAddSampleToElement(
+                            _ctx._report_period_node_list[i],
+                            _ctx._report_period_sensor_type_list[i], NULL, size,
+                            _ctx._samplesPerReport, (_ctx._sample_counter - 1));
+                      }
```
matt001k, line 543: Consolidated code here to reduce copy-pasta.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -995,7 +961,7 @@ static void report_builder_task(void *parameters) {
           if (temp_network_crc32 != _ctx._report_period_max_network_crc32) {
             bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
                            "Getting topology in report builder!\n");
-            uint64_t temp_node_list[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
+            uint64_t temp_node_list[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE] = {};
```
matt001k, line 671: This was never zeroed out and could have big consequences 😁

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -58,6 +131,113 @@ bool BorealisSensor::subscribe() {
...
+    if (data) {
+      memset(data, 0, data_size);
+
+      // TODO: logic needs to be implemented around this
+      data->is_extended = 1;
+
+      // TODO: these need calculations or updates
+      data->spl = 0;
+      data->maq_iqr_band = 0;
+      data->entropy = 0;
```
matt001k: @towynlin TODO's here for when we determine if an extended message will be sent as well as calculations for entropy/spl.

src/apps/bridge/topology_sampler.cpp
```diff
@@ -125,18 +125,17 @@ static void topology_sample_cb(NetworkTopology *networkTopology) {
...
       _node_list.nodes[counter] = cursor->neighbor_table_reply->node_id;
+      bridgeLogPrint(BRIDGE_CFG, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER, "%016" PRIx64 "\n",
+                     _node_list.nodes[counter]);
+      if (!sys_info_service_request(_node_list.nodes[counter], sys_info_reply_cb,
+                                    NODE_SYS_INFO_REQUEST_TIMEOUT_S)) {
+        printf("Failed to send sys info request to node: %" PRIu64 "\n",
+               _node_list.nodes[counter]);
+        exit = true;
+        break;
+      }
```
matt001k, line 136: Noticed a bug here when only single nodes are on a Bristlemouth network 😬 , has been fixed.

towynlin: What is the `_sys_info_handshake` semaphore supposed to protect? Before this PR, it looked like the intent was that you had to have the semaphore before you could call `sys_info_service_request`. Now that request has been pulled out before `xSemaphoreTake`.

src/apps/bridge/reportBuilder.cpp
towynlin: Beautiful cleanup—thanks—but risky to mix in here. I'm ok with leaving it in, but it raises the bar for what we have to test before products ship. We need to ensure current meters, RBR, SOFT, and turbidity sensors all still aggregate correctly with backfilled NaNs.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -404,14 +347,39 @@ void ReportBuilderLinkedList::clear() {
...
+  err = cbor_encode_uint(&sample_array, local_sample_member);
+  return err;
+}
+
+CborError encode_uint32_sample_member(CborEncoder &sample_array, void *sample_member,
+                                      uint32_t size) {
+  (void)size;
   uint32_t local_sample_member = *(uint32_t *)sample_member;
   CborError err = CborNoError;
   err = cbor_encode_uint(&sample_array, local_sample_member);
```
towynlin: I think we should remove these 2 separate uint functions as confusing—just having one `uint` function is better. There's actually no difference between these 2 functions. They both take in a `void*` and call `cbor_encode_uint` which takes a `uint64_t`. In either case, whatever the `void*` is pointing at is getting interpreted as a uint64.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -721,6 +689,13 @@ static bool addSamplesToReport(sensor_report_encoder_context_t &context, uint8_t
     rval = true;
     break;
   }
+  case SENSOR_TYPE_BOREALIS_LEVEL_STATISTICS: {
+    if (BorealisSensorLevelStatistics::encode_sample(sensor_data, sample_index, context) ==
```
towynlin: Oh my goodness thank you for encapsulating that in a function. This giant addSamplesToReport function with long switch cases is something we should improve.

src/apps/bridge/reportBuilder.cpp
towynlin: Sooooo much better thank you thank you

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -58,6 +131,112 @@ bool BorealisSensor::subscribe() {
...
+        memcpy(data->spl_bands_stats, stats.levels, stats.levels_length);
+      }
+
+      reportBuilderAddToQueue(node_id, SENSOR_TYPE_BOREALIS_LEVEL_STATISTICS, data,
+                              sizeof(LevelStatisticsData_t) + stats.levels_length,
+                              REPORT_BUILDER_SAMPLE_MESSAGE);
+      bm_free(data);
+    } else {
+      bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_ERROR, USE_HEADER,
+                     "Failed to allocate memory for Borealis level statistics memory in %s\n",
```
towynlin: nit: "memory" twice

towynlin (approved):
Added some comments, but basically looks good. When it's in a reasonably working state in your testing, let's ship it and keep moving. The proof will be in the end-to-end tests once the whole pipeline is coded up.

matt001k:
@towynlin 

>Beautiful cleanup—thanks—but risky to mix in here. I'm ok with leaving it in, but it raises the bar for what we have to test before products ship. We need to ensure current meters, RBR, SOFT, and turbidity sensors all still aggregate correctly with backfilled NaNs.

I completely agree! I will go through and test before merging this PR. I will add a unit test as well giving us that much more assurance of this working as intended.

matt001k:
@towynlin 

>What is the _sys_info_handshake semaphore supposed to protect? Before this PR, it looked like the intent was that you had to have the semaphore before you could call sys_info_service_request. Now that request has been pulled out before xSemaphoreTake.

This is used to ensure that we receive a sys_info_reply from a node before sending the request to another node (the PR that fixed the timing condition when sending a `sys_info_request` to all nodes on the network #279 to fix the node swapping bug).
This worked fine for multi-node systems but had a bug in a single node system (single node besides the bridge that is). When the semaphore was given at the end of the `sys_info_reply` callback (line 314), the next time the topology sampler is ran, the `topology_sample_cb` does not wait long enough when sending the request to get a reply (the semaphore is already given and would bypass the waiting period)! Testing with a single Borealis helped me catch this whereas all my testing for the linked PR was with a multi-node system, good learning experience.

towynlin:
Got it. Please document in code comments, maybe on the declaration of the semaphore, the expectation of what it's guarding.

towynlin:
Consider clarifying the intent in the variable name, too. For example, sys_info_sequence_lock or something like that.

campbellaos:
I'm fairly strongly opposed to any use of IEEE NaNs as fillvalues in any C code. Any code handling floats should be defensively written as if it might be compiled with -ffinite-math-only, because there are situations where not doing so has too severe of a performance / power consumption penalty. FLT_MAX is a safe alternative fillvalue (although would still need to be tested in context)

src/apps/bridge/sensor_drivers/borealisSensor.h
```diff
@@ -16,6 +30,35 @@ typedef struct BorealisSensor : public AbstractSensor {
...
+    uint8_t is_extended;
+    uint8_t spl;
+    uint8_t max_iqr;
+    uint8_t max_iqr_band;
+    uint8_t entropy;
+    uint8_t spl_band_stats_size;
+    uint8_t spl_bands_stats[0];
+  } LevelStatisticsData_t;
+
+  static constexpr LevelStatisticsData_t AOS_BOREALIS_NAN_AGG = {};
```
victorsowa12: Should this have NAN's or is all zeros the correct value for borealis?

src/apps/bridge/reportBuilder.cpp
```diff
@@ -203,121 +204,43 @@ void ReportBuilderLinkedList::removeElement(report_builder_element_t *element) {
  */
 void ReportBuilderLinkedList::addSampleToElement(report_builder_element_t *element,
                                                  uint8_t sensor_type, void *sensor_data,
-                                                 uint32_t sample_counter) {
+                                                 uint32_t sample_counter, uint32_t size) {
+  void *dst = NULL;
+  const void *nan = NULL;
```
victorsowa12: nit: could we call this `nan_sample`? having `nan` (especially since it shows up blue here in github) makes me think it has some other purpose/meaning other than being a pointer to a sensors nan'd struct.

src/apps/bridge/sensor_drivers/borealisSensor.h
victorsowa12: Talked IRL and zeros is expected since these are all uint8_t's!

victorsowa12 (approved):
Awesome work! 🤘

matt001k:
@towynlin I added a unit test for the `ReportBuilderLinkedList` class. I copy and pasted all of the logic for this file into a separate file name `reportBuilderList.cpp`. This:

1. Cleaned up the original file (`reportBuilder.cpp`) by reducing the lines by 250+
2. Provided a bit of encapsulation for this class into its own file
3. Made testing just this class, which had the risky changes, much easier (it did not rely on having a ton of mocks and stubs that `reportBuilder.cpp` would have required)

matt001k:
I am going to make sure this works as intended on Spotter before merge... thanks for the reviews y'all

src/apps/bridge/sensor_drivers/borealisSensor.h
matt001k: After some further discussion with @towynlin, these values are changed as so:

```
  static constexpr LevelStatisticsData_t AOS_BOREALIS_NAN_AGG = {
      .is_extended = 0,
      .spl = 0xFF,
      .max_iqr = 0xFF,
      .max_iqr_band = 0xFF,
      .entropy = 0xFF,
      .spl_band_stats_size = 0,
      {},
  };
```

The entropy being at 0xFF will indicate that this payload is not valid, as 0 is a very valid entropy level.

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -124,12 +320,34 @@ void BorealisSensor::borealisSubCallback(uint64_t node_id, const char *topic,
...
         struct borealis_level_statistics d;
         if (borealis_levels_statistics_decode(&d, (uint8_t *)data, data_len) == CborNoError) {
-          err = borealis->send_spotter_log_individual(
-              "borealis", d.header,
-              MAX_BOREALIS_READING_PERIOD_MS(borealis->m_reading_period_ms),
-              "%.3f,%.3f,%.3f,%u,%.*s\n", d.dt, d.dt_report, d.max_iqr, d.first_band_index,
-              d.levels_length, d.levels);
-          bm_free(d.levels);
+          BorealisLevelsStatistics_t *level_statistics =
+              reinterpret_cast<BorealisLevelsStatistics_t *>(borealis);
```
towynlin: @matt001k It took me a few minutes of digging today to figure out why this reinterpret cast could even be valid. It's because _**at this particular moment in code history**_ there are no non-static data members of `BorealisLevelsStatistics_t` other than `struct borealis_level_statistics stats;`.

I believe that if anyone in the future ever adds another data member almost anywhere in borealisSensor.h, then this cast becomes silently incorrect and we would have a very hard time debugging.

Please design an API here that doesn't use a reinterpret cast. We should more clearly separate:
- type definitions
- static constants
- static variables
- static methods
- instance variables
- instance methods

Right now these are mixed throughout borealisSensor.h and even nested within each other.

If it's working fine now for the field deployment, such changes can be a separate PR. It's not urgent to do this immediately, but we should improve the maintainability before the final release.

matt001k: Agreed, I think with how this module has developed over the course of Borealis integrations, it grown a bit in a way to structure everything to conform to how other sensors aggregate and collect data. I have an idea of how we can clean this up and will open a PR in the near future.

matt001k: @towynlin updated with a refactor to `borealisSensor.cpp/h` fully tested on a system at my desk a few times to make sure it works and updated unit tests accordingly. Let me know your thoughts. It reduces a lot of extraneous global (some static) variables and typedefs inception 🌀 . Also fit everything in one borealis class and reduced some memory usage as well by only utilizing one Borealis sensor sub callback instead of 4. This will be more maintainable and easier to follow for future dev work.

matt001k: Examining some of the cleanup done in this PR for both the `borealisSensor` module as well as the `reportBuilder` module, I have some good ideas of how we can eliminate the dependency of custom logic for each sensor (Borealis, DO, RBR, etc.) in `reportBuilder`. The report builder should be able to live alone without knowledge of these sensor modules existing. Currently the report builder linked list takes in the sensor type, that is then used in a series of switch case statements to perform certain functions. What could actually be done is utilize a reference to the class `AbstractSensor` that would utilize all Children classes to perform these actions, encapsulating all of this logic within the respective sensor modules, eliminating the growing list of switch case statements.

towynlin: Looks great!

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -79,28 +234,33 @@ void BorealisSensor::borealisSubCallback(uint64_t node_id, const char *topic,
   (void)version;
   BmErr err = BmEINVAL;
   Borealis_t *borealis = NULL;
-  abstractSensorType_e sensor_type = SENSOR_TYPE_UNKNOWN;
+  BorealisSubscriptions_t sub_type = BOREALIS_SUB_COUNT;
+  static constexpr char subtag_spectrum[] = "/aos/borealis/spectrum";
+  static constexpr char subtag_levels[] = "/aos/borealis/levels";
+  static constexpr char subtag_level_statistics[] = "/aos/borealis/level_statistics";
+  static constexpr char subtag_recstatus[] = "/aos/borealis/recstatus";
```
towynlin: @matt001k These are duplicated here and around line 61. Let's dry it up by moving it to the top of the file.

PR #283, opened 2025-04-28
adding config flag to DO app to be able to turn on/off spotter tx data

victorsowa12:
## What changed?
Adding the `sensorDebugTxEnable` config to the DO app to allow for the enabling/disabling of spotter_tx_data from the DO node. By default it will be 0 (OFF).

Immediately after updating the node we can see the first DO sample being logged, but not being sent since by default it will not spotter_tx_data it:
Screenshot 2025-04-28 at 12 02 11 PM

After setting the config to 1 we can see that the data is sent:
Setting the config to 1:
Screenshot 2025-04-28 at 12 13 24 PM
The first DO sample sending out via spotter_tx_data:
Screenshot 2025-04-28 at 12 13 35 PM

After setting the config to 0 we can see that the data is no longer sent via spotter_tx_data again:
Screenshot 2025-04-28 at 12 10 08 PM

This allows users to decrease/increase output or debugging capabilities of their DO sensor.

## Checklist

### DIFF
--- a/src/apps/bristleback_apps/pme_do_sensor/user_code/pme_do_sensor.cpp
+++ b/src/apps/bristleback_apps/pme_do_sensor/user_code/pme_do_sensor.cpp
@@ -21,6 +21,7 @@
 //#define WIPE_INTERVAL_KEY "WipeInterval"
 
 extern uint32_t pmeLogEnable;
+extern uint32_t sensorDebugTxEnable;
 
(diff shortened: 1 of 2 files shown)


victorsowa12:
FYI @b-haist

matt001k (approved):
Code changes look good!

PR #285, opened 2025-05-03
feat: calculate broadband spl

towynlin:
## What changed?

Added calculation of broadband SPL on the bridge when aggregating a sample period given the most recently received Borealis level statistics message. Also report which frequency band had the max IQR.

These summary statistics help provide a brief glance into whether there may have been an interesting signal or only background noise in the acoustic data for the sample period. They are far from perfect, but they're better than nothing.

Spotter users will be able to set a threshold for broadband SPL to reduce satellite data costs when there was likely only background noise in the sample period. We all want to save that research budget for the most interesting and high impact ocean data!

@campbellaos please check my math in the conversions from db to linear units, summing, converting back to db, and quantizing back to 8 bits, as well as how I'm searching for which band had the max IQR.

@matt001k @victorsowa12 please check that I haven't broken the data flow or error states in the report builder pipeline.

## Checklist

_Note: This PR is intended to merge to `develop`, not to `feature/ldr_borealis_telemetry`, but it's currently based on that other PR branch so you can see only what I changed until that gets merged._


src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -98,51 +98,92 @@ bool BorealisSensor::subscribe() {
...
+      max_iqr_band = band_index;
+    }
+
+    // Sum mean SPL for all bands in linear units
+    float mean_db = band_stats[offset + 3] * db_step + db_min;
+    spl_linear_sum += powf(10.0f, mean_db / 10.0f);
+  }
+
+  float broadband_spl_db = 10.0f * log10f(spl_linear_sum);
+  spl = fmaxf(0.0f, fminf(255.0f, broadband_spl_db - db_min) / db_step) + 0.5f;
```
campbellaos: This looks like a subtly wrong order of operations due to misplaced parens, with the consequence that it won't properly clamp at the high end of its range and will instead wrap around. The correct version is: fmaxf(0.0f, fminf(255.0f, (broadband_spl_db - db_min) / db_step + 0.5f))

I can see where this is cribbed from in my code and it's wrong there too, i'll fix it in both places.

towynlin: Thanks, updated. ✅

matt001k (approved):
Code looks good, I enjoyed the changes that you made to the `borealisSensor.cpp` source file, such as adding log statements if the base64 decode fails and renaming the variable `data` to `report`, this feels robust and informative, nice work!

PR #286, opened 2025-05-13
fix: gcc format warnings

towynlin:
## What changed?

I fixed compiler warnings regarding printf format specifiers. They became visible by removing `-Wno-format` from src/CMakeLists.txt. Thanks @campbellaos for the tip!

I added presets for the SDI-12 example app and the PME DO sensor app.

This helps ensure that our printed strings contain the types and widths that we expect and helps us avoid memory corruption.

Going forward, these warnings that we were ignoring will be errors that fail the build.

If any of my format specifier changes look wrong or should be handled differently, let me know.

## Checklist

### DIFF
--- a/src/apps/bridge/network_config_logger.cpp
+++ b/src/apps/bridge/network_config_logger.cpp
@@ -13,7 +13,7 @@ static CborError print_stream_handler(void *out, const char *fmt, ...) {
 
   va_list args;
   va_start(args, fmt);
-  int n = vsnprintf(nullptr, log_buffer_max_size - log_buffer_size, fmt, args);
+  int n = vsnprintf(nullptr, 0, fmt, args);
   if (n > 0 && log_buffer_size < log_buffer_max_size) {
     vsnprintf(&log_buffer[log_buffer_size], log_buffer_max_size - log_buffer_size, fmt, args);
     log_buffer_size += n;
--- a/src/apps/bristleback_apps/pme_do_sensor/user_code/user_code.cpp
+++ b/src/apps/bristleback_apps/pme_do_sensor/user_code/user_code.cpp
@@ -90,16 +90,20 @@ void debugTx(void) {}
 
 void setup(void) {
   // Load system configuration & initialize if not configured by user
-  if (get_config_uint(BM_CFG_PARTITION_SYSTEM, SENSOR_BM_LOG_ENABLE, strlen(SENSOR_BM_LOG_ENABLE), &pmeLogEnable)) {
+  if (get_config_uint(BM_CFG_PARTITION_SYSTEM, SENSOR_BM_LOG_ENABLE,
+                      strlen(SENSOR_BM_LOG_ENABLE), &pmeLogEnable)) {
     pmeLogEnableCfg = true;
   }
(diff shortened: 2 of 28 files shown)


src/lib/CMakeLists.txt
```diff
@@ -126,7 +126,7 @@ if(CMAKE_APP_TYPE STREQUAL "BMDK" OR
    CMAKE_APP_TYPE STREQUAL "bristleback" OR
    CMAKE_APP_TYPE STREQUAL "rs232_expander" OR
    APP_NAME STREQUAL "bringup_mote" OR
-   APP_NAME STREQUAL "mote_bristlefin" OR
+   APP_NAME MATCHES "mote_bristlefin" OR
```
towynlin, line 129: This enabled the debug build, where `-dbg` gets appended to the app name, to also add bristlefin.cpp to `LIB_FILES`.

src/apps/bristleback_apps/pme_do_sensor/user_code/user_code.cpp
```diff
@@ -204,8 +208,9 @@ void loop(void) {
       size_t encodedLen = 0;
       if (PmeDissolvedOxygenMsg::encode(d, cborBuf, sizeof(cborBuf), &encodedLen) ==
           CborNoError) {
-        printf("#  DOT Encoding success! | Topic: %s, cborBuf: %d, \n", pmeDoTopic,
-               cborBuf);
+        printf("#  DOT Encoding success! | Topic: %s, cborBuf: %" PRIx8 "%" PRIx8 "%" PRIx8
+               "...\n",
+               pmeDoTopic, cborBuf[0], cborBuf[1], cborBuf[2]);
```
towynlin, line 213: This printing of the pointer as an integer was not at all valid. I decided to keep it simple and just print the first few bytes. The cbor buffer will always be long enough.

campbellaos:
I recommend replacing "-Wno-format" with "-Wno-error=format -Wno-error=format-security" rather than simply removing it, as the separate -Werror is causing warnings to be errors (and will do so across multiple repositories once merged).

src/apps/bridge/network_config_logger.cpp
```diff
@@ -13,7 +13,7 @@ static CborError print_stream_handler(void *out, const char *fmt, ...) {
 
   va_list args;
   va_start(args, fmt);
-  int n = vsnprintf(nullptr, log_buffer_max_size - log_buffer_size, fmt, args);
+  int n = vsnprintf(nullptr, 0, fmt, args);
```
towynlin, line 16: This was undefined behavior. It's only valid to pass null if the length is zero.

victorsowa12 (approved):

towynlin:
@campbellaos Thanks for the recommendation to use warnings instead of errors.

Since I see that the borealis code builds fine with these format warnings as errors, and since this whole repo does now too, I'm going to keep them as errors that fail the build.

If future folks have trouble compiling because of these errors, please don't hesitate to reach out here on github or on https://bristlemouth.discourse.group/ to discuss. My goal here is to set a high bar for coding standards and prevent us from accidentally slipping into bad habits.

PR #288, opened 2025-05-15
Add salinity config to PME DO

towynlin:
## What changed?

Added a salinity config to the PME DO application.

Code contributions look like they come from @towynlin but this was all in a lot of live coding collaboration with @b-haist and Peter at PME.

This will allow customers of the PME DO sensor to correct their dissolved oxygen values with a reasonable default salinity for the deployment site. In the future, when there's a Bristlemouth salinity sensor, this app will be able to subscribe to the live feed of salinity values!

Not much review needed. PME will test the salinity math against other known good software. Perhaps reviewers could share thoughts on how we're managing the config value and cbor encoder.

## Checklist

### DIFF
--- a/src/apps/bristleback_apps/pme_do_sensor/user_code/pme_do_sensor.cpp
+++ b/src/apps/bristleback_apps/pme_do_sensor/user_code/pme_do_sensor.cpp
@@ -3,6 +3,7 @@
 #include "bm_os.h"
 #include "bm_printf.h"
 #include "configuration.h"
+#include "do_calc.h"
 #include "payload_uart.h"
 #include "serial.h"
 #include "spotter.h"
@@ -145,7 +146,7 @@ void PmeSensor::init() {
  * @param d Reference to a PmeDissolvedOxygenMsg::Data structure where the parsed data will be stored.
  * @return Returns true if data was successfully retrieved and parsed, false otherwise.
  */
-bool PmeSensor::getDoData(PmeDissolvedOxygenMsg::Data &d) {
+bool PmeSensor::getDoData(PmeDissolvedOxygenMsg::Data &d, float salinityPpt) {
   bool success = false;
 
   RTCTimeAndDate_t time_and_date = {};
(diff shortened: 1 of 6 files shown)


src/apps/bristleback_apps/pme_do_sensor/user_code/pme_do_sensor.cpp
```diff
@@ -185,11 +186,16 @@ bool PmeSensor::getDoData(PmeDissolvedOxygenMsg::Data &d) {
...
+        d.header.version = PmeDissolvedOxygenMsg::VERSION;
         d.header.reading_time_utc_ms = rtcGetMicroSeconds(&time_and_date) / 1000;
         d.header.reading_uptime_millis = uptimeGetMs();
+        d.header.sensor_reading_time_ms = 0;
         d.temperature_deg_c = temp_signal.data.double_val;
-        d.do_mg_per_l = do_signal.data.double_val;
+        d.do_mg_per_l =
+            do_signal.data.double_val * salinityFactor(d.temperature_deg_c, salinityPpt);
         d.quality = q_signal.data.double_val;
+        d.salinity_ppt = salinityPpt;
```
victorsowa12, line 203: why are we adding the salinity to the published(output) message? This value seems to be an input into the DO calculations, where as the rest of the values are outputs of the system. Is it valuable to a potential subscriber so we can have a reference for the DO calculations? Should we log it on Spotter? (If so we would need to  parse it on Bridge)

src/apps/bristleback_apps/pme_do_sensor/user_code/user_code.cpp
```diff
@@ -203,7 +232,11 @@ void loop(void) {
       firstDo = false; //Turn off initial DO flag
     }
 
-    if (pmeSensor.getDoData(d)) {
+    // TODO: update PmeDissolvedOxygenMsg::Data to include salinity
```
victorsowa12: Nit: this todo is done?

victorsowa12 (commented):
Looks good! Although I was wondering why the salinity was added to the output message, mainly just curious there since it doesn't impact much.

towynlin:
It's so that one can back out the original uncorrected DO. Additionally once we start subscribing to salinity sensors, we'll need to know what salinity was used to calculate each reading.

victorsowa12 (approved):

src/apps/bristleback_apps/pme_do_sensor/user_code/user_code.cpp
towynlin: Thank you. Yep, we had to stop yesterday with WIP changes on my machine. Already removed there.

src/apps/bristleback_apps/pme_do_sensor/user_code/pme_do_sensor.cpp
towynlin: Yep, need to parse this on the bridge and pass to backend to allow downstream analysis to back out the original uncorrected DO concentration.

src/apps/bristleback_apps/pme_do_sensor/user_code/pme_do_sensor.cpp
```diff
@@ -185,13 +185,22 @@ bool PmeSensor::getDoData(PmeDissolvedOxygenMsg::Data &d) {
...
         d.quality = q_signal.data.double_val;
-        // DO Sat% not available at this time; setting to NULL causes error (converting NULL to double)
-        //d.do_saturation_pct = 0;
+        if (_is_baro_valid) {
+          d.do_saturation_pct = do_signal.data.double_val /
+                                doConcMg(d.temperature_deg_c, _barometric_pressure_mbar, 0.0) *
+                                100.0;
+        } else {
+          d.do_saturation_pct = 0.0;
+        }
```
towynlin: @b-haist please verify this saturation calculation is correct.

towynlin: Peter said the calculation looks good, but we decided to change the invalid value from 0.0 to NAN.

PR #290, opened 2025-06-02
Feature/hydrotwin ldr messages

matt001k:
## What changed?
Hydrotwin LDR message integration.
Subscribes to `/sensor/{node_id}/hydrotwin/ai/low-data-rate` topic and adds data to report builder.

Allows Hydrotwin low-data-rate data to be sent remotely.

## Testing

Testing was done with and without a Hydrotwin unit. Some of the latest released code had to be updated 

- No Hydrotwin Node
  - When there is not a Hydrotwin low-data-rate message to send, Spotter will fill in zero for the length of the hydrotwin data in accordance to the specification
Screenshot 2025-05-30 at 2 28 29 PM

- Hydrotwin Node Reporting LDR
  - When there is a Hydrotwin low-data-rate message to send, Spotter will fill in the length of the hydrotwin message in accordance to the specification
Screenshot 2025-05-30 at 5 04 58 PM
  - The length portion of the message (red output) 0x22, is left shifted by one bit when being bitpacked into the message, meaning the Hydrotwin LDR data is 17 bytes.
  - The following is an output from an online CBOR decoding tool of the 17 bytes in the above payload:
Screenshot 2025-05-30 at 5 05 27 PM

I renamed many variables to improve clarity/reduce obscurity. Please look at the new names and provide your thoughts.
The hydrotwin data was added as another subscription from BorealisSensor handling, I believe this is the cleanest way, but am open to other's opinions as well.

## Checklist

### DIFF
--- a/src/apps/bridge/reportBuilder.cpp
+++ b/src/apps/bridge/reportBuilder.cpp
@@ -526,7 +526,7 @@ static void report_builder_task(void *parameters) {
                         break;
                       }
                       case SENSOR_TYPE_BOREALIS: {
-                        size = sizeof(BorealisLevelStatisticsData_t);
+                        size = sizeof(BorealisAggregationData_t);
                         break;
                       }
                       default: {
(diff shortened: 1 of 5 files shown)


victorsowa12 (approved):

matt001k:
[SC-224850]

PR #291, opened 2025-06-03
Adds Logging To SD Card From Hydrotwin LDR and HDR Messages

matt001k:
## What changed?
Base64 encodes messages to be logged in human readable text
Logs HDR messages to individual log file
Logs LDR messages to aggregate log file

Logs messages directly to Spotter SD card.
Will also:
- Improve troubleshooting units that may have missing data on the back-end
- Provide HDR readings for a unit that may not have cellular connection

I named the app in the log statements "hydrotwin" instead of "borealis" to hopefully provided clarity on what is being output to the SD card as the payloads both include a base64 encoded string. Should it be named "borealis"?

The following is the output of these messages on Spotter:
image

## Checklist

### DIFF
--- a/src/apps/bridge/sensor_drivers/borealisSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/borealisSensor.cpp
@@ -323,12 +323,11 @@ BmErr BorealisSensor::hydrotwinSendSpotterLog(const uint8_t *data, uint16_t data
   // Base64 encoded data
   if (encoded_data_len) {
     uint8_t *encoded_data = static_cast<uint8_t *>(bm_malloc(encoded_data_len));
-
-    err = BmENODEV;
-    mbedtls_base64_encode(encoded_data, encoded_data_len, &encoded_data_len,
-                          (const unsigned char *)data, data_len);
+    err = BmENOMEM;
 
     if (encoded_data) {
+      mbedtls_base64_encode(encoded_data, encoded_data_len, &encoded_data_len,
+                            (const unsigned char *)data, data_len);
       if (type == HYDROTWIN_LDR) {
         err = send_spotter_log_aggregate("hydrotwin", m_hydrotwin_ldr_minute, "%.*s\n",
                                          encoded_data_len, encoded_data);
--- a/src/apps/bridge/sensor_drivers/borealisSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/borealisSensor.cpp
@@ -2,6 +2,7 @@ extern "C" {
 #include "util.h"
 }
 #include "app_config.h"
+#include "app_util.h"
 #include "bm_config.h"
 #include "bm_os.h"
 #include "borealisSensor.h"
@@ -13,6 +14,7 @@ extern "C" {
 #include "pubsub.h"
 #include "sensorController.h"
 #include "spectral_entropy.h"
+#include "uptime.h"
 #include <inttypes.h>
 #include <math.h>
 #include <new>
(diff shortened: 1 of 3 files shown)


src/lib/common/app_util.h
```diff
@@ -41,6 +41,8 @@ double degToRad(double deg);
   pdTICKS_TO_MS(timeRemainingGeneric(pdMS_TO_TICKS(startTimeMs), xTaskGetTickCountFromISR(),   \
                                      pdMS_TO_TICKS(timeoutMs)))
 
+#define MIN_TO_MS(minute) (minute * 1000)
```
victorsowa12: I believe this needs a `* 60`

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -287,6 +292,58 @@ BmErr BorealisSensor::encode_sample(void *data, uint32_t sample_index,
...
+
+  // Base64 encoded data
+  if (encoded_data_len) {
+    uint8_t *encoded_data = static_cast<uint8_t *>(bm_malloc(encoded_data_len));
+
+    err = BmENODEV;
+    mbedtls_base64_encode(encoded_data, encoded_data_len, &encoded_data_len,
+                          (const unsigned char *)data, data_len);
+
+    if (encoded_data) {
```
victorsowa12: would it make sense to put the encoding within the `if (encoded_data)` that way we don't encode if we failed to malloc?

victorsowa12 (approved):
Couple small comment but looks good overall!

src/apps/bridge/sensor_drivers/borealisSensor.cpp
matt001k: It would 100% make sense, oops!

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -287,6 +292,57 @@ BmErr BorealisSensor::encode_sample(void *data, uint32_t sample_index,
...
+  mbedtls_base64_encode(NULL, 0, &encoded_data_len, (const unsigned char *)data, data_len);
+
+  // Base64 encoded data
+  if (encoded_data_len) {
+    uint8_t *encoded_data = static_cast<uint8_t *>(bm_malloc(encoded_data_len));
+    err = BmENOMEM;
+
+    if (encoded_data) {
+      mbedtls_base64_encode(encoded_data, encoded_data_len, &encoded_data_len,
+                            (const unsigned char *)data, data_len);
```
towynlin: It would be safer to have a separate output variable, rather than using encoded_data_len twice. This should be fine in most cases, but I wouldn't want hard-to-understand, under-the-hood, compiler optimization shenanigans to screw that up.

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -608,6 +714,10 @@ Borealis_t *createBorealisSensorSub(uint64_t node_id) {
   static constexpr uint32_t default_borealis_reading_period_ms = 1000;
   static constexpr char reading_period_key[] = "bandsSampleTimeMs";
   static constexpr char level_statistics_period_key[] = "report_interval";
+  static constexpr char hydrotwin_ldr_key[] = "hydrotwin_ldr";
+  static constexpr char hydrotwin_hdr_key[] = "hydrotwin_hdr";
+  static constexpr uint32_t hydrotwin_ldr_minute_default = 5;
```
towynlin: The typical default we plan to ship single-BOREALIS systems with is `12` for the LDR minutes. I'm not 100% sure we should change this particular value. The goal would be to ensure that the hydrotwin LDR config is at least 90 seconds shorter than the bridge sample duration.

towynlin (approved):
I like hydrotwin for the logging app name. 👍 

Made a couple other minor comments.

PR #296, opened 2025-06-09
Brians latest DO changes

victorsowa12:
## What changed?
A few changes @b-haist has made to the microDOT mote code.

Gets the microDOT's OS version.

Not too many changes so all of them :)

## Checklist


b-haist (approved):
Much cleaner on the log, log_console, and spotter_tx_data functions with logEnable checks. SN included in wipe files too, so everything looks good!

PR #297, opened 2025-06-10
feat: spectral entropy

towynlin:
## What changed?

- Added spectral entropy module in a new lib/dsp directory
- On each received levels message (once per second) bridge inserts the levels into a growing list for later spectral entropy calculations
- When aggregating after each sample duration:
  - bridge passes the means from the borealis level stats to the spectral entropy module, which
  - computes prewhitened spectral entropy across the list of all received levels
  - and returns the _*minimum*_ spectral entropy from the entire sample duration.

The minimum spectral entropy from a sample duration will be one measure of signal that will allow Spotter Sound customers to control their satellite data expenditures.

The idea of using spectral entropy as a measure of likelihood of signal in acoustic data comes from [Automatic detection of marine mammals using information entropy](https://doi.org/10.1121/1.2982368) by Christine Erbe & Andrew R. King, published in Volume 124, Issue 5 of the Journal of the Acoustical Society of America, November 2008. The paper is very vague about how to perform the calculations, so this is our own implementation.

Customers can configure an entropy threshold on the Bridge. If the minimum spectral entropy from a particular sample duration is below the configured threshold, then there is more likely to be a signal of interest (more tonal, less like white noise), and Spotter will send a larger message with more detailed statistics per third-octave band over the sample duration.

If, on the other hand, the minimum spectral entropy from a sample period is above the configured threshold, then the sample duration was consistently noisy enough that it's less likely to contain a signal of interest. In this case, Spotter will send a smaller message to save on telemetry costs.

- Memory handling in the BOREALIS_LEVELS case of borealisSubCallback starting around line 359 of borealisSensor.cpp
- Memory handling in the list usage in the spectral entropy module
- Please don't hesitate to ask questions about the calculations.

## Checklist

### DIFF
--- a/src/apps/bridge/sensor_drivers/borealisSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/borealisSensor.cpp
@@ -190,9 +190,6 @@ void BorealisSensor::aggregate(void) {
       }
     }
 
-    // TODO: calculate entropy
-    report.entropy = REPORT_NAN_ERROR_VALUE;
-
     // Add hydrotwin LDR
     if (tracking_data.ldr.size) {
       report.hydrotwin_ldr.buf = static_cast<uint8_t *>(bm_malloc(tracking_data.ldr.size));
--- a/src/apps/bridge/sensor_drivers/borealisSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/borealisSensor.cpp
@@ -114,6 +114,7 @@ BmErr BorealisSensor::calculateQuantizedSplAndEntropy(uint8_t const *const band_
     spl = REPORT_NAN_ERROR_VALUE;
     max_iqr_band = REPORT_NAN_ERROR_VALUE;
     entropy = REPORT_NAN_ERROR_VALUE;
+    clear_spectral_entropy_list();
     return BmEINVAL;
   }
 
--- a/src/apps/bridge/sensor_drivers/borealisSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/borealisSensor.cpp
@@ -702,31 +702,3 @@ void BorealisSensor::parse_levels(float *spl_db, const char *levels_as_base64,
     spl_db[i] = unpack_12bit(raw_bytes, i * 3) * cstep + clow;
   }
 }
-
-void BorealisSensor::parse_level_statistics(float *stats_db, const char *stats_as_base64,
-                                            size_t b64_length) {
-  size_t out_len = 0;
-  mbedtls_base64_decode(NULL, 0, &out_len, reinterpret_cast<const uint8_t *>(stats_as_base64),
-                        b64_length);
-  uint8_t raw_bytes[out_len];
-  int err =
-      mbedtls_base64_decode(raw_bytes, sizeof(raw_bytes), &out_len,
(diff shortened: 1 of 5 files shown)


src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -106,22 +107,25 @@ bool BorealisSensor::subscribe() {
...
-  const float db_min = -256.0f * db_step;
-  const unsigned int num_stats = 4; // 25%, 50%, 75%, mean
+  constexpr float db_step = 0.75f;
+  constexpr float db_min = -256.0f * db_step;
+  constexpr unsigned int num_stats = 4; // 25%, 50%, 75%, mean
   const unsigned int num_bands = band_stats_len / num_stats;
 
   uint8_t max_iqr_so_far = 0;
   float spl_linear_sum = 0.0f;
+  float means[num_bands] = {0.0f};
```
matt001k: I do not have a problem with VLAs, but have run into some strange behavior in a past experience where, in comparison to this this case, `num_bands` got corrupted and `means` tried to allocate a huge amount of memory onto the stack crashing the processor. It might be a good idea to do some bound checking here, since we know the max number of bands will be 48 or utilize `bm_malloc` and `NULL` check the output.

src/lib/dsp/spectral_entropy.c
```diff
@@ -0,0 +1,99 @@
...
+  LLItem *item = NULL;
+  static uint32_t id = 0;
+  item = ll_create_item(item, spl_db, data_size, id++);
+  ll_item_add(&spectra, item);
+}
+
+static BmErr compute_entropies_and_find_minimum(void *data, void *arg) {
+  float *const band_levels = (float *const)data;
+  SpectralEntropyArgs *args = (SpectralEntropyArgs *)arg;
+  const float entropy = compute_spectral_entropy(args->num_bands, band_levels, args->means);
```
matt001k, line 79: Curious if it is worth checking if entropy is NAN here, I am not sure how line 80 will respond if NAN is assigned to `entropy`.
If we do fail `compute_spectral_entropy` by returning NAN, should the traversal through the linked list still be continued or should the API return an error?

matt001k (approved):
Code looks good, concise and readable! I liked how you were able to extend the `calculateQuantizedSpl` function without adding a lot of complexity. Had a couple of questions/concerns I would like some feedback on.

src/apps/bridge/sensor_drivers/borealisSensor.cpp
towynlin: Thanks for calling that out. Moved the means array to the heap in commit d7e3965.

src/lib/dsp/spectral_entropy.c
towynlin: As a general rule (that I just verified with some quick test code) all comparisons with NAN evaluate to false. That means in the rare case where `compute_spectral_entropy` returns NAN then we won't adjust `min_entropy`, and the code will do a good thing, keeping the previous value of `min_entropy`. I think the code as-is does the right thing. Thanks for highlighting it though, I hadn't worked through that. 🙏

PR #300, opened 2025-06-11
Feature/bm bridge lock removal

matt001k:
## What changed?
Allows receiving/transmitting messages over UART to occur at the same time.

Optimize throughput and reduce complexity by removing semaphore.

Review the testing data:
Data Gaps:
    - From a 16 hour test with a system that has two SOFT modules reporting:
    Screenshot 2025-06-11 at 11 12 07 AM
    - No decrease in performance here, but no increase either (will have to implement some ack/nack scheme for improved integrity), but reduces code complexity

## Checklist

### DIFF
--- a/src/lib/bm_ncp/ncp_uart.cpp
+++ b/src/lib/bm_ncp/ncp_uart.cpp
@@ -355,9 +355,6 @@ static bool bm_int_gpio_callback_fromISR(const void *pinHandle, uint8_t value, v
 
   if (!value) {
     lpmPeripheralActiveFromISR(LPM_USART3_RX); // Active low
-    HAL_UART_AbortReceive(&huart3);
-    HAL_UARTEx_ReceiveToIdle_DMA(&huart3, const_cast<uint8_t *>(ncpRXBuff[ncpRXCurrBuff]),
-                                 NCP_BUFF_LEN);
   }
 
   return xHigherPriorityTaskWoken;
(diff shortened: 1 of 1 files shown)


towynlin:
Looks fine at a glance, but I don't immediately understand the details. I'd love a verbal explanation, either here or in a quick huddle, of how the lock was being used prior to this change, why we don't need it, and what precisely some of the added conditionals are handling now that there's no lock.

matt001k:
@towynlin for sure! I will go ahead and make it public on here just in case anyone else has the same question!
When the UART line was interrupt driven at a low baud rate Spotter would occasionally drop some data, when the baud rate increased to 1M,  Spotter began to drop even more data. 
Why is this?

Well @victorsowa12 was able to prove we were not able to handle these interrupts quick enough by toggling a gpio line every time the UART interrupt fired and analyzing the output on a logic analyzer (there were times the GPIO would skip characters).

This became more apparent when bridge or spotter was attempting to transmit at the same time that it was receiving data.

As a band-aid I utilized the shared GPIO line between Spotter and Bridge as a way to control communication.
If bridge was asserting the line, Spotter would not send data until bridge was done transmitting and visa-versa.
This was controlled with a semaphore that would lock the transmit line until the other side de-asserted the shared GPIO line.
Allowing the processors to receive data without dropping bytes.

Now that DMA has come into play, the CPU does not have to spend any time in the USART interrupt transmitting and receiving data, meaning that the band-aid aforementioned, no longer has to be there!

towynlin (approved):
Thanks for the explanation!

victorsowa12 (approved):

PR #304, opened 2025-06-12
Fix USB detection when hwVersion is incorrect, develop-v0.12

victorsowa12:
## What changed?
check both pins so that way if hwVersion is 0 but we are actually a hwVersion 7 we can still talk via USB. This makes sure that when a v7 bridge has hwVersion set to 0 it will still be able to access USB.

Makes sure bridge can access USB.

## Checklist

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -204,7 +204,7 @@ set(MICROPY_COMPILER_FLAGS
 
 	-Wno-unused-parameter
 	-Wno-dangling-pointer
-	-Wno-error=dangling-pointer
+	-Wno-sign-compare
 
 	# GRRRR
 	# -Werror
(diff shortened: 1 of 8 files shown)


victorsowa12:
Getting some build errors on the v0.12 on my machine too so I'll figure those out. PR #305 is ready though. Build errors don't look related to these changes.

towynlin (approved):

victorsowa12:
I don't know why but for some reason the micropython builds still fail on the v0.12 branch in CI (I can't reproduce locally) even after adding the `-Wno-dangling-pointers` to the micropython compile flags. So I commented out the micropython CI builds for v0.12 here.

towynlin:
I have not heard of anyone actually using micropython, and we very much need this USB fix. :shipit:

PR #305, opened 2025-06-12
Fix USB detection when hwVersion is incorrect, develop

victorsowa12:
## What changed?
check both pins so that way if hwVersion is 0 but we are actually a hwVersion 7 we can still talk via USB. This makes sure that when a v7 bridge has hwVersion set to 0 it will still be able to access USB.

But this time on normal develop branch.

Makes sure bridge can access USB.

## Checklist


matt001k (approved):

matt001k:
💰

towynlin (approved):

PR #307, opened 2025-06-18
remove wiper aggregations

victorsowa12:
## What changed?
Removing aggregations of PME wiper data since the default interval is 4 hours, we are not going to be wiping more than once per sample duration.

### Testing:
I increased the wiper frequency to make sure we got some and here we can see several wipe logs in the SENS_IND, but none appearing in the SENS_AGG:
Screenshot 2025-06-18 at 3 29 25 PM

This will prevent a bunch of nananananananananan logs in the aggregations file.

Small change so everything.

## Checklist


towynlin (approved):
_Love_ all that deleted code. ✨

PR #309, opened 2025-06-24
Make VUSB and LED_G use iternal pull downs on Bridge

victorsowa12:
In order to make sure old/new bridge HW still enters stop mode the VUSB_DETECT and LED_G(formerly VUSB_DETECT) need to be set as pull down. This makes sure that when USB is not connected niether pin is floating high enough to make the USB detect return true accidentally.

### Testing:
Old Bridge HW in LPM, then USB is plugged in and the bridge is reset, thus staying out of LPM(the purple line high = not in stop mode):
Screenshot 2025-06-23 at 5 12 38 PM

New bridge HW w/ hwVersion == 0. We can see the USB was plugged in after a reset, keeping the bridge out of LPM, then after unplugging the USB and resetting bridge we enter LPM:
Screenshot 2025-06-23 at 5 14 28 PM

New bridge HW w/ hwVersion == 7. After an initial period of the USB being plugging in right after the reset, we can see the bridge enter/exit stop mode very quickly as USB is plugged/unplugged:
Screenshot 2025-06-23 at 5 15 59 PM

## What changed?
Makes VUSB_DETECT and LED_G use pull downs.

Makes sure that the bridge enters low power mode!

Everything. Should we just rename the LED_G to (VUSB_DETECT_OLD?).

## Checklist


towynlin (approved):
We could consider the naming change for the future, but no need right now. :shipit:

PR #311, opened 2025-06-25
fix: allow PME DO quality value greater than 1

towynlin:
## What changed?

Changed bad data max threshold for quality value from PME DO sensor from 1.0 to 2.0, based on conversations with PME.

Allows data to flow through the pipeline when Q > 1.0. Such data points previously would have been dropped and replaced with error signals.

## Checklist

### DIFF
--- a/src/apps/bridge/sensor_drivers/pmeDissolvedOxygenSensor.h
+++ b/src/apps/bridge/sensor_drivers/pmeDissolvedOxygenSensor.h
@@ -37,7 +37,7 @@ typedef struct PmeDissolvedOxygenSensor : public AbstractSensor {
   static constexpr double DO_SAMPLE_MEMBER_MIN = -1.0;
   static constexpr double DO_SAMPLE_MEMBER_MAX = 100.0;
   static constexpr double QUALITY_SAMPLE_MEMBER_MIN = 0.0;
-  static constexpr double QUALITY_SAMPLE_MEMBER_MAX = 1.0;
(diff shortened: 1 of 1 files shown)


matt001k (approved):

src/apps/bridge/sensor_drivers/pmeDissolvedOxygenSensor.h
```diff
@@ -37,7 +37,7 @@ typedef struct PmeDissolvedOxygenSensor : public AbstractSensor {
   static constexpr double DO_SAMPLE_MEMBER_MIN = -1.0;
   static constexpr double DO_SAMPLE_MEMBER_MAX = 100.0;
   static constexpr double QUALITY_SAMPLE_MEMBER_MIN = 0.0;
-  static constexpr double QUALITY_SAMPLE_MEMBER_MAX = 1.0;
+  static constexpr double QUALITY_SAMPLE_MEMBER_MAX = 2.0;
```
victorsowa12, line 40: this is going to require changes to the bitpacking since we use 10 bits with a range of 0-1023 to represent quality. 1,000 would equal 1.000. If we tried to send anything greater than 1.000 we would run out of bits.

towynlin: Got it. Thank you for calling that out. Let's make this change here and discuss separately how to update the bitpacking.

PR #312, opened 2025-06-26
Bridge Turns On Port 2 When Network Powers Back On

matt001k:
## What changed?
Ensure port 2 is disabled when bridge power controller turns the network back on

Saves power by disabling port 2.

The sequence of port power events that occurs here is as follows (from initialization):

1. Enable adin driver and turn on both ports bcl_init-> bristlemouth_init -> adin2111_init->adin_netdevice_enable
2. Disable port 2 bm_l2_netif_enable_disable_port
3. Wait 2 minutes for bridge initialization to finish and turn off the network power in the power controller module (bm_l2_netif_set_power->adin_netdevice_disable )
4.  Power up Bristlemouth network from the power controller module (bm_l2_netif_set_power-> adin_netdevice_enable )
  a. This is where the bug occurs, we turn on both ports here

## Checklist

### DIFF
--- a/src/apps/bridge/bridgePowerController.cpp
+++ b/src/apps/bridge/bridgePowerController.cpp
@@ -123,8 +123,11 @@ void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
     xEventGroupSetBits(_busPowerEventGroup, signal_to_set);
     if (notifyL2) {
       bm_l2_netif_set_power(on);
+      // Must turn port 2 off if powering up the network,
+      // port 2 is not used on the bridge and disabling the port will provide
+      // power savings
       if (on) {
-        bm_l2_netif_enable_disable_port(2, !on);
+        bm_l2_netif_enable_disable_port(2, false);
       }
     }
     bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER, "Bridge bus power: %d\n",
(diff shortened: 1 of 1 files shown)


src/apps/bridge/bridgePowerController.cpp
```diff
@@ -123,6 +123,9 @@ void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
     xEventGroupSetBits(_busPowerEventGroup, signal_to_set);
     if (notifyL2) {
       bm_l2_netif_set_power(on);
+      if (!on) {
+        bm_l2_netif_enable_disable_port(2, on);
+      }
```
victorsowa12: should we always turn off port 2? Bridge should never be using it right?

victorsowa12: Oh I see the `bm_l2_netif_set_power` call should handle turning it off! ignore my comment.

victorsowa12: wait - if the bus is turning off - then we are turning off the port. Shouldn't we already be doing that with `bm_l2_netif_set_power`? And then in this case we would want to also turn it off when the bus is turned on right?

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -123,6 +123,9 @@ void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
     xEventGroupSetBits(_busPowerEventGroup, signal_to_set);
     if (notifyL2) {
       bm_l2_netif_set_power(on);
+      if (on) {
+        bm_l2_netif_enable_disable_port(2, !on);
```
towynlin: I think I'd prefer an explicit `false` instead of `!on`.

I think it would also be helpful to document some implicit tribal knowledge with a comment on the preceding line(s) explaining to our future colleagues and to the open source community that bridge port 2 is not connected to anything, so we should always save the power by disabling it.

matt001k: Agreed.

towynlin (approved):

victorsowa12 (approved):

PR #315, opened 2025-07-01
fix: clean up unhelpful comments

towynlin:
## What changed?

Removed an unhelpful comment that was repeated in many apps.
Also simplified the periodic LED blink in hello world and serial payload example

Improving the friendliness of the code so it's more welcoming and approachable.

The LED blink code feels a little overcomplicated. I just quickly simplified the 2 apps that people look at the most. I chose not to continue sweeping up around the codebase, since it could be a neverending journey. Let me know if you disagree.

I also only fixed up the whitespace in a limited way without running clang-format on these entire files. Doing so would modify some of these files on almost every line. Let me know if you think they should be formatted anyway.

## Checklist

### DIFF
--- a/src/apps/bm_devkit/bm_loadcell/user_code/user_code.cpp
+++ b/src/apps/bm_devkit/bm_loadcell/user_code/user_code.cpp
@@ -83,10 +83,6 @@ void loop(void) {
     led2State = false;
   }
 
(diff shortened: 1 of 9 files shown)


matt001k (approved):
Reduce all the things! 
responsibility12(alternate)

PR #316, opened 2025-07-02
Prevent Watchdog From Tripping In Bridge App

matt001k:
## What changed?
Prevents task watchdog from tripping by providing better separation of responsibilities in topology_sampler.
The topology (not topology sampler) task has a heavy load placed on it 

Adds servicing watchdog to FreeRTOS idle hook rather than its own task to prevent 2 second latency in the Watchdog task.

Ensures topology sampling timers are reset rather than "started".

The watchdog was tripping while many tasks were in the running state:
image
This offloads a lot of responsibility for the topology task.

- Prevents Bridge Watchdog resets due to latency in the Topology task
- Reduces latency that could occur in the watchdog task

## Testing
With this branch, I have had the bridge running for 16 hours without a watchdog reset on a system that was experiencing a watchdog reset once every 8 hours.

`topology_sampler.cpp` has the main fix. Resetting of timers rather than starting and separating the logic from the topology callback into the `topology_sampler` task.

## Checklist


src/lib/common/watchdog.c
```diff
@@ -47,14 +25,9 @@ void watchdogFeed() {
 //        periodically and the watchdog is only refreshed when several
 //        conditions are met.
 //
-static void iWDGTask( void *parameters ) {
-  // Don't warn about unused parameters
-  (void) parameters;
 
-  for(;;) {
+void vApplicationIdleHook(void) {
```
matt001k, line 29: `bm_protocol`'s watchdog timeout is around 4 seconds, when this was in a task with a delay of 2 seconds, we cut our watchdog down, in worst case scenario to 2 seconds.
This provides some breathing room and a true 4 second watchdog as this hook gets ran continuously in the idle task (which is of lowest task priority).

victorsowa12:
Looks like there are two main changes here:
1. changes to the topolgy tasks
3. changes to the watchdog

Do we know if either one of these changes in isolation fixes the problem?

Might be worth testing if we still experience the watchdog asserts if we change the watchdog to use the idle hook and don't have the topology changes. That should at least tell us that the new watchdog is working properly. Or vice versa and test only the topology changes with the old watchdog and verify we don't hit the old watchdog.

src/apps/bridge/topology_sampler.cpp
```diff
@@ -616,28 +608,29 @@ void topology_sampler_task(void *parameters) {
...
       // Check if we are already sampling, if not, wait (using blocking waitForSignal) for power to turn on
       // if we are sampling, wait for power to turn off
-      if (!_sampling_enabled && _bridge_power_controller->waitForSignal(true, portMAX_DELAY)) {
+      if (!_sampling_enabled && _bridge_power_controller->waitForSignal(true, 0)) {
         // lets wait 5 seconds for the devices on the bus to power up
         vTaskDelay(pdMS_TO_TICKS(BUS_POWER_ON_DELAY));
         topology_sample();
         // start the timer! here while the bus is powered we will sample topology every minute
-        configASSERT(xTimerStart(topology_timer, 10));
+        configASSERT(xTimerReset(topology_timer, 10));
```
victorsowa12: When I look at the freertos API it states the xTimerStart does the same thing as xTimerReset when the timer is active and that xTimerReset does the same thing as xTimerStart when the timer is dormant. I.e. if the timer is dormant here it will always be started regardless of using start/reset and if the timer is active it will always be reset.

What is the behavior change that is expected here?

matt001k:
@victorsowa12 that is a good point in order to have a true answer to what it might be. It also might be the combination of both of these.
What is inconvenient is there is only one system to test on that is really exhibiting this behavior and there is a bit of a rush to get it shipped!

A long soak with 3 separate systems could maybe give us an indication, but seeing as it is inconsistent between different Bridge modules I am not sure how beneficial it will be.

If we can hold off on shipping the poor performing one, then long soaks can be ran with:

1. Changes to topology task
2. Updated watchdog handling
3. Combination of 1 & 2

src/apps/bridge/topology_sampler.cpp
matt001k: Doing a bit of research into FreeRTOS timers (and trying to understand the ambiguity between `xTimerStart` and `xTimerReset`), there does not seem to be a benefit of using one over another. I originally thought that the timer held state when stopped (i.e. the timer had the same uptick value), but it does not seem like that is the case when the timer is placed into the dormant state.
This API can be reverted with no harm.

matt001k:
@victorsowa12, ran a test over the long weekend. On the system showing this behavior the worst, in 5 days there were no resets due to watchdog crashes (there was a crash due to another reason but not longer a watchdog crash).

I ran another system with just the watchdog file changes and did not see any crashes on that system and it is hard to correlate whether or not just that was the fix, due to the fact this system was not experiencing watchdog resets ever.
But my gut is telling me a combination of both of these will help prevent watchdogs by:

1. Having a true 4 second watchdog
2. Reducing latency in the TOPO task

towynlin:
@matt001k I think this seems like a positive change and the lack of crashes over the weekend is a strong signal. However, this PR is pretty tough to review. Please schedule a meeting to walk us through the changes and so we can have some live discussion.

towynlin (approved):

PR #317, opened 2025-07-02
Mote Example App & BQ25820 Driver for UPitt

mbella-sofar:
## What changed?
1. I added the bq25820 driver source files I wrote in [this branch](https://github.com/bristlemouth/bm_protocol/tree/mote_bringup_arpa-e).
2. I made a copy of the bm_devkit example app called arpa-e_example that enables VBUS to the mezzanine connector, initializes the bq25820 and disables PFM with a register write to the bq25820 in the setup function.

This branch adds basic support for the ARPA-E power expander board so that the University of Pittsburgh has a good starting point for their firmware development.

Please focus on the details that will impact the customer (UPitt) experience working with this example as a starting point for their firmware. I would especially like to make sure the customer facing code in the setup function is representative of how to things should be done.

## Checklist


src/lib/drivers/bq25820.cpp
```diff
@@ -0,0 +1,343 @@
...
+#include "app_util.h"
+
+using namespace BQ;
+
+#define PN_REV 0x1A
+#define BQ_RST (1 << 7)
+#define POWERPATH_REG_DEF 0x20
+
+static const uint16_t inaAvgCounts[] = {1, 4, 16, 64, 128, 256, 512, 1024};
+static const uint16_t inaConvTimes[] = {140, 203, 332, 588, 1100, 2116, 4156, 8244};
```
victorsowa12: nit: ina is the power monitor, so these could be named bqAvgCounts and bqConvTimes.

src/lib/drivers/bq25820.h
```diff
@@ -0,0 +1,48 @@
...
+  bool disablePfm();
+  bool readSensors(char *buffer, size_t len);
+  bool readFaults(char *buffer, size_t len);
+
+private:
+  bool setCfgBits(uint16_t bits, uint8_t mask, uint8_t shift);
+  bool readReg(Reg_t reg, int16_t *value);
+  bool read8(Reg_t reg, uint8_t *value);
+  bool writeReg(Reg_t reg, uint16_t value);
+  bool write8(Reg_t reg, uint8_t value);
```
victorsowa12: nit: could make this consistent and have write16, read16, just to help clarify between the different writes and reads

victorsowa12 (approved):
Couple of minor nitpicks, but overall looks really good! Nice work! 😎🤘

src/lib/drivers/bq25820.cpp
mbella-sofar: Good catch. I forgot to remove those when I copied the ina232 source files.

src/lib/drivers/bq25820.h
mbella-sofar: Good call. Fixed!

src/lib/drivers/bq25820.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+/*!
+  Set bits in configuration register
+
+  \param[in] bits Value to set in config
+  \param[in] mask Mask for value size (since it will have to clear the bits first)
+  \param[in] position Number of bits to left shift before clearing/setting value
+  \return true if successfull false otherwise
+*/
+bool BQ25820::disablePfm() {
```
matt001k: Is this comment accurate for this API? Looks like it might be associated with `setCfgBits`, should that function be defined somewhere/is it necessary for this application?

src/lib/drivers/bq25820.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+  uint16_t i;
+
+  write8(ADC_CH_CTRL_REG, 0x00);
+
+  do {
+    if(!write8(ADC_CTRL_REG, 0xE0)) { // 0xA8 = old val
+      break;
+    }
+    for (i = 0; i < 1000; i++) {
```
matt001k, line 226: Could you explain this function a bit in the above as a comment block similar to some of the other API you wrote?

It looks like you are looping through and reading the `CHARGER_FLAG_1` register until bit  `ADC_DONE_FLAG` is set, then reading the ADCs on the device for different functionalities. Should this function also store those values in a struct and return them or populate them as a pass by reference argument into the function rather than assign a string buffer? I feel like external modules might want to know some of these values and it would be harder to access them in the string format.
The struct could be something like:
```
typedef struct {
  int16_t iac;
  int16_t ibat;
  float vac;
  float vbat;
  float vbat_fb;
  float vsys;
  float ts;
} BQ25820ADC_t;
```

Then passed by reference into the function and assigned like so:

```
bool BQ25820::readSensors(BQ25820ADC_t &adc_readings) {
...
```

src/lib/drivers/bq25820.cpp
```diff
@@ -0,0 +1,328 @@
...
+  } while(0);
+  
+  read8(CHARGER_CTRL_REG, &tmpreg);
+  printf("0x%X\n", tmpreg);
+
+  return rval;
+}
+
+
+bool BQ25820::readFaults(char *buffer, size_t len) {
```
matt001k: With this function I believe the faults may be more useful externally to this module as an `enum` or `integer` type value rather than a string, it can be handled much cleaner if there is an associated action to take with the fault.  There could be another API in the driver (or rename this API) that may have the name `readFaultsStrings` which would perform this action.
Let me know your thoughts.

src/lib/drivers/bq25820.cpp
mbella-sofar: The comment has been updated

src/lib/drivers/bq25820.cpp
mbella-sofar: Very true. I started this branch from the quick/dirty bringup code I wrote to enable soak testing. I don't think UPitt cares about the battery charger too much, but I am still going to rewrite this function and the read sensors function before I share the branch with them so we can get info about the health and operation of the charger for debugging if we need it later.

matt001k (approved):
I believe some of my comments can be addressed in future work and are not necessary right now, it would definitely add a lot of value to those using this diver! Awesome work!

mbella-sofar:
I cleaned up a bit more of the things that Matt highlighted. I also started looking into adding the needed code to the sensorSampler to log the data from my bq25820 driver in order to provide UPitt with correctly written sensor logging code. I stopped working on this once I realized the scope of the work needed. They can navigate the process to add this logging to their code later on if we need to debug battery charger issues.

PR #322, opened 2025-07-10
Fix/do memory leak

matt001k:
## What changed?
Removed unused log buffer leftover from previous spotter logging functionality and clean up variable that was being malloc'd and never freed.

Prevent crashing on the bridge with DO systems due to running out of RAM.

Rubber stamp.

## Checklist


src/apps/bridge/sensor_drivers/pmeDissolvedOxygenSensor.cpp
```diff
@@ -49,8 +49,6 @@ void PmeDissolvedOxygenSensor::pmeDissolvedOxygenSubCallback(
     if (bm_semaphore_take(dissolved_oxygen_sensor->_mutex, BM_MAX_DELAY_UINT32) == BmOK) {
       static PmeDissolvedOxygenMsg::Data dissolved_oxygen_data;
       if (PmeDissolvedOxygenMsg::decode(dissolved_oxygen_data, data, data_len) == CborNoError) {
-        char *log_buff = static_cast<char *>(bm_malloc(SENSOR_LOG_BUF_SIZE));
-        configASSERT(log_buff);
```
matt001k, line 53: This memory was never being freed when receiving a published message from the DO.

victorsowa12 (approved):

PR #323, opened 2025-07-10
When Bridge Aggregation Is Disabled, Prevent Borealis Memory Leak

matt001k:
When transmit aggregations are disabled on the bridge, memory is leaked into the spectral entropy list whenever a Borealis levels message is received.
The aggregation method in the BorealisSensor class will never get called and the entropy list elements will never be freed.

## What changed?
Only allow levels SPL data to be added to spectral entropy list when aggregations are enabled
Always clear spectral entropy list whenever an aggregation report is created

When bridge report aggregation parameter is disabled, prevent memory from leaking through the spectral entropy list
Always clears the spectral entropy list whenever an aggregation report is created preventing memory from never being freed when aggregation reporting is enabled

Any other edge cases that could be leaking memory here?
From my initial testing everything looks good, but would like more eyes on this file.

## Checklist

### DIFF
--- a/src/apps/bridge/sensor_drivers/borealisSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/borealisSensor.cpp
@@ -128,7 +128,6 @@ BmErr BorealisSensor::calculateQuantizedSplAndEntropy(uint8_t const *const band_
   max_iqr_band = REPORT_NAN_ERROR_VALUE;
   entropy = REPORT_NAN_ERROR_VALUE;
   if (band_stats == NULL || band_stats_len == 0) {
-    clear_spectral_entropy_list();
     return BmEINVAL;
   }
 
@@ -141,7 +140,6 @@ BmErr BorealisSensor::calculateQuantizedSplAndEntropy(uint8_t const *const band_
   if (means == NULL) {
     bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_ERROR, USE_HEADER,
                    "Failed to allocate for Borealis level stats means\n");
-    clear_spectral_entropy_list();
     return BmENOMEM;
   }
   for (size_t band_index = 0; band_index < num_bands; band_index++) {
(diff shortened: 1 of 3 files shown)


src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -229,6 +227,8 @@ void BorealisSensor::aggregate(void) {
             exceedsConfiguredThresholds(report.spl, report.max_iqr, report.entropy);
       }
     }
+    // Clear the spectral entropy list after entropy has been calculated
+    clear_spectral_entropy_list();
```
matt001k, line 231: If line 212 is false because we cannot allocate memory, then the spectral entropy list is never freed.
This is a very plausible thing to occur, if the `report_interval` on Borealis is set higher than the `sampleDurationMs` of the bridge, then the `bm_malloc` will fail due to an attempt to allocate 0 bytes (evaluate the function definition of `pvPortMalloc` and see that if 0 bytes are attempted to be allocated, it will return NULL).
It is best to free the memory here!

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -476,6 +479,7 @@ void BorealisSensor::borealisSubCallback(uint64_t node_id, const char *topic,
             // to be less than half of the bridge power controller's sampleDurationMs, we cannot
             // send multiple of level stat messages yet and need to free the unused string
             bm_free(borealis->tracking_data.stats.levels);
+            borealis->tracking_data.stats.levels = NULL;
```
matt001k, line 482: NULL this to prevent a potential double free if the `report_interval` ever gets changed.

src/lib/dsp/spectral_entropy.c
```diff
@@ -90,11 +90,9 @@ void clear_spectral_entropy_list(void) {
   }
 }
 
-float calc_min_spectral_entropy_and_clear_list(const unsigned int num_bands,
-                                               float const *const means) {
+float calc_min_spectral_entropy(const unsigned int num_bands, float const *const means) {
   min_entropy = 1.0f;
   SpectralEntropyArgs arg = {.num_bands = num_bands, .means = means};
   ll_traverse(&spectra, compute_entropies_and_find_minimum, &arg);
-  clear_spectral_entropy_list();
```
matt001k, line 98: This is now cleared in the data aggregation method.

towynlin (approved):
Great work!

victorsowa12 (approved):

PR #325, opened 2025-07-11
If Power Controller Is Disabled, Avoid Leaking Memory; Format Hydrotwin HDR Data Nicely

matt001k:
## What changed?
Do not track data if it will never be reported.
Data will never be reported if aggregation reports are disabled or the power controller is disabled.
Formats hydrotwin HDR `SENS_IND` log statements to be more cohesive.

Prevents memory leak.
Prevents confusion in users when viewing SENS_IND file.

Mainly the memory leak conditions.

## Checklist

### DIFF
--- a/src/apps/bridge/sensor_drivers/borealisSensor.cpp
+++ b/src/apps/bridge/sensor_drivers/borealisSensor.cpp
@@ -1,7 +1,6 @@
 extern "C" {
 #include "util.h"
 }
-#include "app_config.h"
 #include "app_util.h"
 #include "bm_config.h"
 #include "bm_os.h"
@@ -14,6 +13,7 @@ extern "C" {
 #include "pubsub.h"
 #include "sensorController.h"
 #include "spectral_entropy.h"
+#include "stm32_rtc.h"
 #include "uptime.h"
 #include <inttypes.h>
 #include <math.h>
@@ -47,7 +47,7 @@ static constexpr char config_borealis_tx_entropy_threshold[] = "borealisTxEntrop
 void BorealisSensor::init(void) {
   static constexpr uint32_t MAX_SAMPLES_PER_REPORT = 2;
   uint32_t samples_per_report = 0;
-  power_config_s pwr_cfg = {};
+  m_pwr_cfg = getPowerConfigs();
 
   m_aggregation_reports = report_builder_get_transmit_aggregations();
   samples_per_report = report_builder_get_samples_per_report();
@@ -60,7 +60,7 @@ void BorealisSensor::init(void) {
                    samples_per_report, MAX_SAMPLES_PER_REPORT);
   }
 
-  if (pwr_cfg.subsampleEnabled) {
+  if (m_pwr_cfg.subsampleEnabled) {
     bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_WARNING, USE_HEADER,
                    "Subsampling enabled, be aware that BOREALIS data will only transmit last "
                    "collected level statistics message in sampleDurationsMs period remotely\n");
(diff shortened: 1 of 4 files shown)


src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -365,8 +365,8 @@ BmErr BorealisSensor::hydrotwinSendSpotterLog(const uint8_t *data, uint16_t data
         SensorHeaderMsg::Data header = {
             hydrotwin_message_version,
             uptimeGetMs(),
-            bm_ticks_to_ms(bm_get_tick_count()),
-            0,
+            rtcGetMicrosecondsSimple(),
+            rtcGetMicrosecondsSimple(),
```
matt001k, line 369: This formatting looked improper when reviewing some data:

The following is an example:

```
bm_node_id,node_position,node_app_name,reading_uptime_millis,reading_time_utc_s,sensor_reading_time_s
...
b3ae73485cf40ff4,1,hydrotwin,1662214,1662.214,0.000,hAEaaG9J1YIYHAmiBRgcEQk=
```

Even though Hydrotwin is not formatting the header like other nodes on the system, it can still be done on Bridge to clean things up.

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -497,7 +498,8 @@ void BorealisSensor::borealisSubCallback(uint64_t node_id, const char *topic,
             bm_debug("Failed to send borealis aggregated log to spotter, reason: %d\n", err);
           }
 
-          if (!borealis->m_aggregation_reports) {
+          if (!borealis->m_aggregation_reports ||
+              !borealis->m_pwr_cfg.bridgePowerControllerEnabled) {
             // Free levels information here as it will not be aggregated
             bm_free(d.levels);
```
matt001k, line 504: This will actually get freed on line 481, but let's prevent any lingering allocation from occurring anyway if `bridgePowerControllerEnabled` is false.

src/apps/bridge/sensor_drivers/borealisSensor.cpp
```diff
@@ -517,7 +519,8 @@ void BorealisSensor::borealisSubCallback(uint64_t node_id, const char *topic,
         }
 
         err = borealis->hydrotwinSendSpotterLog(data, data_len, sub_type);
-        if (borealis->m_aggregation_reports) {
+        if (borealis->m_aggregation_reports &&
+            borealis->m_pwr_cfg.bridgePowerControllerEnabled) {
           borealis->tracking_data.ldr.buf = static_cast<uint8_t *>(bm_malloc(data_len));
```
matt001k, line 524: This will actually get freed on line 516, but let's prevent any lingering allocation from occurring anyway if `bridgePowerControllerEnabled` is false.

src/lib/drivers/stm32_rtc.c
```diff
@@ -163,6 +163,13 @@ BaseType_t rtcSet(const RTCTimeAndDate_t *timeAndDate) {
   return rval;
 }
 
+uint64_t rtcGetMicrosecondsSimple(void) {
+  RTCTimeAndDate_t time_and_date = {};
+  rtcGet(&time_and_date);
+
+  return rtcGetMicroSeconds(&time_and_date);
+}
```
towynlin, line 171: 🎉 🤩

src/apps/bridge/sensor_drivers/borealisSensor.cpp
towynlin: Nice. Thank you!

towynlin (approved):

PR #332, opened 2025-08-18
Feature/serial bridge subscribe

matt001k:
## What changed?
Added subscription capability to serial bridge application.
Adds serial bridge application to Bristleback.

Expands the serial bridge application to allow companion devices to subscribe to topics.

Mainly the diff in the serial bridge apps!

## Checklist

### DIFF
--- a/src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
+++ b/src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
@@ -1,52 +1,115 @@
+#include "user_code.h"
 #include "bm_os.h"
 #include "bm_serial.h"
-#include "bristlefin.h"
 #include "cobs.h"
 #include "payload_uart.h"
 #include "pubsub.h"
-#include "sensors.h"
 #include "task_priorities.h"
-#include <inttypes.h>
 #include <string.h>
 
-typedef enum {
-  BM_NONE,
-  BM_SUB,
-} __attribute__((__packed__)) bm_serial_tx_message_t;
-typedef struct {
-  uint8_t type;
-  uint32_t length;
-  uint8_t data[0];
-} bm_serial_tx_t;
-static bm_serial_callbacks_t callbacks;
-static constexpr size_t rx_buf_size = 2048;
-static uint8_t rx_buffer[rx_buf_size];
-static size_t rx_idx = 0;
-static constexpr size_t packet_buf_size = rx_buf_size;
-static uint8_t packet_buffer[rx_buf_size];
+static constexpr size_t RX_BUF_SIZE = 2048;
 
+struct BmSerialBridgeCtx {
+  bm_serial_callbacks_t callbacks;
+  struct {
+    size_t idx;
+    uint8_t buffer[RX_BUF_SIZE];
+  } rx;
+  struct {
+    uint8_t buffer[RX_BUF_SIZE];
+  } packet;
+};
+
+static struct BmSerialBridgeCtx_t ctx; //Serial Bridge Context
+
+/*!
+ @brief Serial Transmit Callback For bm_serial Library
+
+ @details Raw bytes (buff) are passed from the bm_serial library to this
+          function, which will output data to the PLUART interface.
+
+ @param buff Buffer to transmit serially
+ @param len Length of buffer to transmit
+ @param message Unused
+
+ @return true always
+ */
 static bool tx_cb(const uint8_t *buff, size_t len, bm_serial_message_t message) {
   (void)message;
   PLUART::write((uint8_t *)buff, len);
   return true;
 }
 
+/*!
+ @brief Subscription Callback When Topic Of Interest Is Published To Network
+
+ @details Passes topic information and data serially to serially connected
+          device.
+
+ @param node_id Node ID which published the message
+ @param topic Topic message was published on
+ @param topic_len topic length
+ @param data Data associated with topic, can be NULL
+ @param data_len Length of data
+ @param type Topic type
+ @param version Topic version
+ */
 static void sub_message_cb(uint64_t node_id, const char *topic, uint16_t topic_len,
                            const uint8_t *data, uint16_t data_len, uint8_t type,
                            uint8_t version) {
-  printf("Publishing To Topic: %s\n", topic);
-  bm_serial_pub(node_id, topic, topic_len, data, data_len, type, version);
+  if (bm_serial_pub(node_id, topic, topic_len, data, data_len, type, version) != BM_SERIAL_OK) {
+    printf("Failed to publish message to serial device\n");
+
+  } else {
+    printf("Publishing To Topic: %s\n", topic);
+  }
 }
 
+/*!
+ @brief Subscription Callback For bm_serial Library
+
+ @details When serial message is received for this type of message, subscribe
+          to the topic of interest on the Bristlemouth network and pass it
+          the received data through to the serially connected device.
+
+ @param topic Topic to subscribe to
+ @param topic_len Length of topic
+
+ @return True on success, false on failure
+ */
 static bool sub_cb(const char *topic, uint16_t topic_len) {
   return bm_sub_wl(topic, topic_len, sub_message_cb);
 }
 
+/*!
+ @brief Publish Callback For bm_serial Library
+
+ @details Handles publishing data to bristlemouth network from serially
+          connected device.
+
+ @param topic Topic to publish to
+ @param topic_len Length of topic
+ @param node_id Unused
+ @param payload Data to publish can be NULL
+ @param len Length of data
+ @param type Topic type
+ @param version Topic version
+
+ @return 
+ */
 static bool pub_cb(const char *topic, uint16_t topic_len, uint64_t node_id,
                    const uint8_t *payload, size_t len, uint8_t type, uint8_t version) {
   (void)node_id;
   printf("publishing %u bytes to %.*s\n", len, topic_len, topic);
   return bm_pub_wl(topic, topic_len, payload, len, type, version) == BmOK;
+}
+
+/*!
+ @brief Reset RX Buffer
+ */
+static void serial_rx_reset() {
+  ctx.rx.idx = 0;
+  memset(ctx.rx.buffer, 0, RX_BUF_SIZE);
 }
 
 void setup() {
@@ -56,11 +119,11 @@ void setup() {
   PLUART::setUseLineBuffer(false);
   PLUART::enable();
 
-  memset(&callbacks, 0, sizeof(bm_serial_callbacks_t));
-  callbacks.pub_fn = pub_cb;
-  callbacks.sub_fn = sub_cb;
-  callbacks.tx_fn = tx_cb;
-  bm_serial_set_callbacks(&callbacks);
+  memset(&ctx.callbacks, 0, sizeof(bm_serial_callbacks_t));
+  ctx.callbacks.pub_fn = pub_cb;
+  ctx.callbacks.sub_fn = sub_cb;
+  ctx.callbacks.tx_fn = tx_cb;
+  bm_serial_set_callbacks(&ctx.callbacks);
   IOWrite(&BB_VBUS_EN, 0);
   // ensure Vbus stable before enable Vout with a 5ms delay.
   vTaskDelay(pdMS_TO_TICKS(5));
@@ -68,24 +131,20 @@ void setup() {
   IOWrite(&BB_PL_BUCK_EN, 0);
 }
 
-static void reset() {
-  rx_idx = 0;
-  memset(rx_buffer, 0, rx_buf_size);
-}
-
 void loop() {
   while (PLUART::byteAvailable()) {
     const uint8_t b = PLUART::readByte();
-    rx_buffer[rx_idx++] = b;
-    if (rx_idx >= rx_buf_size) {
+    ctx.rx.buffer[ctx.rx.idx++] = b;
+    if (ctx.rx.idx >= RX_BUF_SIZE) {
       printf("buffer full, resetting\n");
-      reset();
+      serial_rx_reset();
     } else {
       if (b == 0) {
         cobs_decode_result result =
-            cobs_decode(packet_buffer, packet_buf_size, rx_buffer, rx_idx - 1);
+            cobs_decode(ctx.packet.buffer, RX_BUF_SIZE, ctx.rx.buffer, ctx.rx.idx - 1);
         if (result.status == COBS_DECODE_OK) {
-          bm_serial_packet_t *packet = reinterpret_cast<bm_serial_packet_t *>(packet_buffer);
+          bm_serial_packet_t *packet =
+              reinterpret_cast<bm_serial_packet_t *>(ctx.packet.buffer);
           bm_serial_error_e err = bm_serial_process_packet(packet, result.out_len);
           if (err != BM_SERIAL_OK) {
             printf("packet processing error: %d\n", err);
@@ -93,7 +152,7 @@ void loop() {
         } else {
           printf("cobs decode error: %d\n", result.status);
         }
-        reset();
+        serial_rx_reset();
       }
     }
   }
--- a/src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
+++ b/src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
@@ -78,7 +78,7 @@ static void sub_message_cb(uint64_t node_id, const char *topic, uint16_t topic_l
  @return True on success, false on failure
  */
 static bool sub_cb(const char *topic, uint16_t topic_len) {
-  return bm_sub_wl(topic, topic_len, sub_message_cb);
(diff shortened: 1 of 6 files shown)


victorsowa12 (approved):

towynlin (approved):

src/apps/bristleback_apps/serial_bridge/user_code/user_code.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
+
+void setup();
```
LarkspurRaven, line 3: Add comments for functions. Though, this interface is used in many other places. Thinking this common interface could go somewhere like `src/lib/user_app`

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
+#include "bm_os.h"
+#include "bm_serial.h"
+#include "bristlefin.h"
+#include "cobs.h"
+#include "payload_uart.h"
+#include "pubsub.h"
+#include "sensors.h"
+#include "task_priorities.h"
+#include <inttypes.h>
+#include <string.h>
```
LarkspurRaven, line 5: 1. This doesn't include `user_code.h`. 
2. `<inttypes.h> appears to not be used`. I think standard int types are coming from `<stdin.h>` via `bm_os.h`

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+typedef enum {
+  BM_NONE,
+  BM_SUB,
+} __attribute__((__packed__)) bm_serial_tx_message_t;
```
LarkspurRaven: This could change to
```
__PACKED_STRUCT bm_serial_tx_message_t {
    // fields here
};
```

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+static bool pub_cb(const char *topic, uint16_t topic_len, uint64_t node_id,
+                   const uint8_t *payload, size_t len, uint8_t type, uint8_t version) {
+  (void)node_id;
+  printf("publishing %u bytes to %.*s\n", len, topic_len, topic);
+  return bm_pub_wl(topic, topic_len, payload, len, type, version) == BmOK;
+}
+
+void setup() {
+  PLUART::init(USER_TASK_PRIORITY);
```
LarkspurRaven: What happens if PLUART is used by another task?

LarkspurRaven: Looking at `payload_uart::init()`, I have some concern there. 
1. I suppose PLUART is attempting to do a Singleton pattern, but there is nothing I see that prevents it from being called by multiple users, each potentially changing the setup
2. The init function just allocates a new task, but there is no way I see to destroy the task. 
3. For all cases, `payload_uart::init()` configures `MX_LPUART1_UART_Init`, but then we are allowed to set the other parameters such as `baud`, `useByteStreamBuffer`, etc and also independantly `enable/disable`
4. There is nothing that calls `payload_uart::disable()`. But even if we did, it still does not clean up the FreeRTOS task. Perhaps it shouldnt' for disable, but the name should be called `disable_serial()`. Perhaps there could be a seperate function `disable_serial_task()` -- though, that wouldn't just work as pluart is designed because no handle to the task is returned to the caller. 
5. Outside the scope of this PR, but I don't understand why payload_uart was hardcoded for a specific uart

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
LarkspurRaven: wait -- `bm_serial_tx_message_t` is an enum. I don't think it makes sense to make an enum a packed struct. Did you mean to put that on `bm_serial_tx_t`?

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "bristlefin.h"
+#include "cobs.h"
+#include "payload_uart.h"
+#include "pubsub.h"
+#include "sensors.h"
+#include "task_priorities.h"
+#include <inttypes.h>
+#include <string.h>
+
+typedef enum {
```
LarkspurRaven: Add comments to `bm_serial_tx_message_t`, `bm_serial_tx_message_t`, and the variables declarations below. Add new lines after the enum & struct declarations.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+  uint8_t data[0];
+} bm_serial_tx_t;
+static bm_serial_callbacks_t callbacks;
+static constexpr size_t rx_buf_size = 2048;
+static uint8_t rx_buffer[rx_buf_size];
+static size_t rx_idx = 0;
+static constexpr size_t packet_buf_size = rx_buf_size;
+static uint8_t packet_buffer[rx_buf_size];
+
+static bool tx_cb(const uint8_t *buff, size_t len, bm_serial_message_t message) {
```
LarkspurRaven: Add comments for each internal method

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+  callbacks.tx_fn = tx_cb;
+  bm_serial_set_callbacks(&callbacks);
+  IOWrite(&BB_VBUS_EN, 0);
+  // ensure Vbus stable before enable Vout with a 5ms delay.
+  vTaskDelay(pdMS_TO_TICKS(5));
+  // enable Vout, 12V by default.
+  IOWrite(&BB_PL_BUCK_EN, 0);
+}
+
+static void reset() {
```
LarkspurRaven: `reset_rx` may be more clear b/c that is all it is resetting

should we reset packet_buffer here too?

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+typedef enum {
+  BM_NONE,
+  BM_SUB,
+} __attribute__((__packed__)) bm_serial_tx_message_t;
+typedef struct {
+  uint8_t type;
+  uint32_t length;
+  uint8_t data[0];
+} bm_serial_tx_t;
```
LarkspurRaven: `bm_serial_tx_t` doesn't seem to be used. If correct, can remove it.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+  PLUART::init(USER_TASK_PRIORITY);
+  PLUART::setBaud(115200);
+  PLUART::setUseByteStreamBuffer(true);
+  PLUART::setUseLineBuffer(false);
+  PLUART::enable();
+
+  memset(&callbacks, 0, sizeof(bm_serial_callbacks_t));
+  callbacks.pub_fn = pub_cb;
+  callbacks.sub_fn = sub_cb;
+  callbacks.tx_fn = tx_cb;
```
LarkspurRaven: The callbacks could be defined compile time
```
static bm_serial_callbacks_t callbacks = {
    callbacks.pub_fn = pub_cb,
    callbacks.sub_fn = sub_cb,
    callbacks.tx_fn = tx_cb;,
}
```
you'd just have to forward declare the functions

I do have a slight concern for namespace collection with the generic function names `pub_sub`, `sub_cb`, `tx_cb`. We should be safe for compile b/c they are delcared static. Something like `serial_bridge_pub_cb` could make it easier to read and trace.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
+#include "bm_os.h"
```
LarkspurRaven: I think it'd be more descriptive to name this file something like `serial_bridge` rather than the generic `user_code.cpp`. Though I realize this is a common pattern.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
LarkspurRaven: actually, it's worse `user_byte_stream_buffer` is static and is reset on each call of init() (same for `_user_line`, `_postTxSemaphore`, `uart_handle`). It may still work ok, but this would be a leak for multiple users. Granted, out of scope for this PR, but unless I'm crazy, if pluart  is intended to be a singleton with one client, init should be something like this
```
// Check if already initialized
        if (rxTaskHandle != NULL) {
            return pdFAIL; // Already initialized
        }
```

though it _may_ be okay to allow mutliple callers with one instance. In the current design, allowing multiple tasks doesn't make sense because we are hard-binded to one specific uart.

LarkspurRaven: man...pluart also exposes these in the header
```
// external declarations for the variables
extern SerialLineBuffer_t lpUART1LineBuffer;
extern SerialHandle_t uart_handle;
```
these need moved to the cpp file

@matt001k sorry for these pluart comments! I just don't want to lose the thoughts!

LarkspurRaven (dismissed):

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+typedef enum {
+  BM_NONE,
+  BM_SUB,
+} __attribute__((__packed__)) bm_serial_tx_message_t;
+typedef struct {
+  uint8_t type;
+  uint32_t length;
+  uint8_t data[0];
+} bm_serial_tx_t;
+static bm_serial_callbacks_t callbacks;
```
LarkspurRaven: Add comments for variable declarations

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+static constexpr size_t rx_buf_size = 2048;
```
LarkspurRaven: 1. `rx_buf_size` s/b capitalized --> `RX_BUF_SIZE`. Well, not 100% certain on our naming standards...need to check
2. As we are using this for both `rx_buf` and `pkt_buff`, could be better to name as `BUFFER_SIZE`
3. This is just my preference, but I prefer to have configurable values like this declared at the top of the file in a marked section. Perhaps in a header. Is this buffer size common for other user_code apps?

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+static uint8_t rx_buffer[rx_buf_size];
+static size_t rx_idx = 0;
+static constexpr size_t packet_buf_size = rx_buf_size;
+static uint8_t packet_buffer[rx_buf_size];
+
+static bool tx_cb(const uint8_t *buff, size_t len, bm_serial_message_t message) {
+  (void)message;
```
LarkspurRaven: should probably add a check on `buff` for null pointer, since it is not coming from us. Perhaps add a check for length as well. Though maybe PLUART::write() checks both. 

I don't really understand who calls this method `tx_cb()`, or when it'd be called. Comment could clarify that.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+  PLUART::write((uint8_t *)buff, len);
+  return true;
+}
+
+static void sub_message_cb(uint64_t node_id, const char *topic, uint16_t topic_len,
```
LarkspurRaven: in `sub_message_cb()` are `topic` and `data` guaranteed to not be null? If not, we should probably add a null check. is there any validation we could consider on the topic and data? 

Perhaps `bm_serial_pub()` handles my concerns? If so, we should call it first

LarkspurRaven: `bm_serial_pub` does check the params, so we should call it first
```
if(bm_serial_pub(...){
      printf("Failed to send serial_bridge msg\n");
      some_method_to_safely_print_this_message(...);
    } else {
      printf("Publishing To Topic: %s\n", topic);
    }
```

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+                           const uint8_t *data, uint16_t data_len, uint8_t type,
+                           uint8_t version) {
+  printf("Publishing To Topic: %s\n", topic);
+  bm_serial_pub(node_id, topic, topic_len, data, data_len, type, version);
+}
+
+static bool sub_cb(const char *topic, uint16_t topic_len) {
+  return bm_sub_wl(topic, topic_len, sub_message_cb);
```
LarkspurRaven: `bm_sub_wl` returns `BmErr` not bool. Should cast or handle

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+static bool pub_cb(const char *topic, uint16_t topic_len, uint64_t node_id,
+                   const uint8_t *payload, size_t len, uint8_t type, uint8_t version) {
+  (void)node_id;
+  printf("publishing %u bytes to %.*s\n", len, topic_len, topic);
+  return bm_pub_wl(topic, topic_len, payload, len, type, version) == BmOK;
```
LarkspurRaven: bm_pub_wl returns BmErr not bool. Should cast or handle

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
LarkspurRaven: hmm...seeing the `vTaskDelay(pdMS_TO_TICKS(5));` code in the `setup` function made me think that we may actually have the context for the task here, so we may be able to teardown the task from here. Though the memory leaks in pluart remain

src/apps/bristleback_apps/serial_bridge/user_code/user_code.h
matt001k: I think this is supposed to mimic an Arduino like interface for folks looking to quickly dev on the Bristlemouth development kits. Hence why all of the `user_code` files follow this trend. I would take a look at some of the [Bristlemouth DevKit](https://bristlemouth.notion.site/Development-Kit-User-Guides-e9ca1b3c5a1c41c890d0105f2eb7c4b8) instructions for users looking to develop in the Bristlemouth ecosystem if you haven't taken  a look yet!

towynlin:
Good comments, thanks @LarkspurRaven — this PR is low priority and does not need to make it into v0.13.4.

LarkspurRaven (commented):
not quite done with review, but progress so far

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
matt001k: This typedef is unused and has been removed.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
matt001k: This is unused and has been removed.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
matt001k, line 1: @LarkspurRaven this file was copied from another application. I will make suggested changes but it will not be reflected in the other `user_code.cpp` file located in this diff.
There should probably be some common location for application that can exist in multiple application spaces.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
matt001k: Correct, the statements reads:

` bm_pub_wl(topic, topic_len, payload, len, type, version) == BmOK`

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
matt001k: Definitely worth capturing, right now there is no contention in any of our apps but could for sure be in the future.

src/apps/bristleback_apps/serial_bridge/user_code/user_code.cpp
LarkspurRaven: my mistake 👍

PR #335, opened 2025-08-21
turn off the boost converter when the bus is OFF

victorsowa12:
## What changed?
Turns off the Boost converter when the BUS if OFF. This saves ~3mW of power!

Here is a screenshot showing a bridge before this change. I measured the BUS off spotter + bridge system power at around 27mW. I then connected to the bridge's CLI, toggled the BOOST_EN pin low to turn it off, unplugged the USB and let the joulescope collect an average again and saw it went down to ~24mW!
Screenshot 2025-08-20 at 2 15 41 PM

This saves ~3mW of power!

Smal change so all of it!

## Checklist


victorsowa12:
Marking as draft because I actually haven't tested yet with nodes attached! I will notify when that testing is complete and it is ready to review!

matt001k:
👀 save all the power!

victorsowa12:
After a test with @mbella-sofar, we determined that a 15ms delay would give ample time + some buffer for the ~8mS charge up time on the capacitors.
SDS2354X Plus_PNG_159

victorsowa12:
Now I will test with nodes to make sure there is no funky business during a soak test.

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -18,11 +18,13 @@
 #endif // RAW_PRESSURE_ENABLE
 
 BridgePowerController::BridgePowerController(
-    IOPinHandle_t &BusPowerPin, uint32_t sampleIntervalMs, uint32_t sampleDurationMs,
-    uint32_t subsampleIntervalMs, uint32_t subsampleDurationMs, bool subsamplingEnabled,
-    bool powerControllerEnabled, uint32_t alignmentS, bool ticksSamplingEnabled)
-    : _BusPowerPin(BusPowerPin), _powerControlEnabled(powerControllerEnabled),
-      _sampleIntervalS(sampleIntervalMs / 1000), _sampleDurationS(sampleDurationMs / 1000),
+    IOPinHandle_t &BusPowerPin, IOPinHandle_t &BoostPowerPin, uint32_t sampleIntervalMs,
+    uint32_t sampleDurationMs, uint32_t subsampleIntervalMs, uint32_t subsampleDurationMs,
```
LarkspurRaven: ## Constructor Parameter Count Code Smell

_**Note:** Though not strictly related to your power changes, we should at least discuss these suggestions and either make changes now, or create an SC ticket for a quick followup._ 

Adding `BoostPowerPin` brings our constructor to 9 parameters, which violates established best practices (I had to pause and carefully parse each parameter to line them up, and it would be easy to mistake order):

**Clean Code (Robert C. Martin):** *"More than three arguments (triadic) should be avoided where possible. More than three (polyadic) requires very special justification—and then shouldn't be used anyway."*

**Refactoring (Martin Fowler):** Lists "Long Parameter List" as a classic code smell, recommending the Parameter Object pattern as the solution.

**Parameter Object Pattern:** Groups related parameters into a single object to improve readability, reduce errors, and encapsulate validation logic. Transforms complex constructors into self-documenting, maintainable code.

### Proposed Solution: Parameter Object Pattern [F1]

```cpp
struct BridgePowerControllerCfg {
  IOPinHandle_t &BusPowerPin;
  IOPinHandle_t &BoostPowerPin;  // New parameter fits naturally
  uint32_t sampleIntervalMs;
  uint32_t sampleDurationMs;
  // ... other config values
};

// Much clearer usage:
BridgePowerControllerCfg cfg = {
  .BusPowerPin = VBUS_SW_EN,
  .BoostPowerPin = BOOST_EN,     // Self-documenting
  .sampleIntervalMs = pwrcfg.sampleIntervalMs,
  // ...
};
BridgePowerController controller(cfg);
```

ctor changes to 
```
BridgePowerController::BridgePowerController(BridgePowerControllerCfg cfg) 
    : _cfg(cfg), _timebaseSet(false), _initDone(false), _configError(false) {
```


## Move checks into validate function [F2]

Can also move the ctor validation checks into a `validate()` function in `BridgePowerControllerCfg`
```
// Validation belongs with the data it validates
  bool validate() {
    bool hasErrors = false;
    if (_sampleIntervalS > MAX_SAMPLE_INTERVAL_S || _sampleIntervalS < MIN_SAMPLE_INTERVAL_S) {
       printf("INVALID SAMPLE INTERVAL, using default.\n");
       configError = true;
       sampleIntervalS = DEFAULT_SAMPLE_INTERVAL_S;
       hasErrors = true;
    }
   ...
    return hasErrors;
}
```
use it
```
BridgePowerController::BridgePowerController(BridgePowerControllerCfg cfg)
    : _cfg(cfg) {
  
  // Single line validation!
  _configError = !_cfg.validate();
```

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -119,7 +121,14 @@ void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
     EventBits_t signal_to_set = (on) ? BridgePowerController::ON : BridgePowerController::OFF;
     EventBits_t signal_to_clear = (on) ? BridgePowerController::OFF : BridgePowerController::ON;
     xEventGroupClearBits(_busPowerEventGroup, signal_to_clear);
-    IOWrite(&_BusPowerPin, on);
+    if (on) {
```
LarkspurRaven: Need clarification. The parameters are not defined in the fxn signature here or in the header. What does `on` mean in this context? [A1]

I think `on` is not descriptive enough -- what is turning on? The function name is overloaded with powerBus and SetSignal. [A2]

Also, now that we are adding `_BoostPowerPin` to this method, should the function name be changed to reflect that the `boost` is affected, or we add another param to set the Boost separate from the `busPower` pin? [A3]

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -119,7 +121,14 @@ void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
...
+      IOWrite(&_BoostPowerPin, on);
+      vTaskDelay(100); // Allow time for capacitors to charge
+      IOWrite(&_BusPowerPin, on);
+    } else {
+      IOWrite(&_BusPowerPin, on);
+      IOWrite(&_BoostPowerPin, on);
```
LarkspurRaven: ~~What does the change in ordering accomplish here? The end result seems the same.~~

^--- _at first glance, I missed that the else clause is actually powering the bus off. Though the function name `powerBus` part of `powerBusAndSetSignal` primed me to think this function was just for power up. AND just setting `true/false` on the parameter isn't real clear either. 

Could be improved to provide wrapper functions to manager setting true/false -- [F3]
```
void powerBusAndSetSignalOn(...) { powerBusAndSetSignal(true, ...) }
void powerBusAndSetSignalOff(...) { powerBusAndSetSignal(false, ...) }
```
Or using an enum [F3]
```
enum ePowerBusCommand {
   BUS_OFF,
   BUS_ON,
};
powerBusAndSetSignal(BUS_OFF, ...);
powerBusAndSetSignal(BUS_ON, ...);
```

Again -- just recording my initial reactions ... :)_ 

---

In the `on` case, we do `_BoostPowerPin` followed by `_BusPowerPin` and the opposite in the else clause. Also, why do we not want a delay for the caps to charge in the else clause?

Can you add some clarifying comments?

Or, perhaps better, we could add helper functions to wrap the logic
```
void BridgePowerController::enableBusPower() {
  // Power-up sequence: Boost first, then bus (but why?) 
  IOWrite(&_BoostPowerPin, true);      // Enable boost power supply
  vTaskDelay(100);                     // Allow capacitors to charge and voltage to stabilize
  IOWrite(&_BusPowerPin, true);        // Enable bus power
}

void BridgePowerController::disableBusPower() {
  // Power-down sequence: Bus first, then boost  (but why?) 
  IOWrite(&_BusPowerPin, false);       // Disable bus power (remove load first)
  IOWrite(&_BoostPowerPin, false);     // Disable boost power supply
  // No delay needed - capacitors will discharge naturally
}
```
then

```
if (on) { 
    enableBusPower() 
} else { 
   disableBusPower() 
}
```
or more compact
```
on ? enableBusPower() : disableBusPower();
```

another option is to extract into one larger function
```
// Handles power-up and power-down sequences for X platform
void BridgePowerController::setBusPower(bool enable) {
  if (enable) {
    // Power-up sequence: Boost first, then bus (but why?) 
    IOWrite(&_BoostPowerPin, true);
    vTaskDelay(100); // Allow capacitors to charge
    IOWrite(&_BusPowerPin, true);
  } else {
    // Power-down sequence: Bus first, then boost (but why?) 
    IOWrite(&_BusPowerPin, false);
    IOWrite(&_BoostPowerPin, false);
  }
}
```
then
```
void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
...
    setBusPower(on);
```

test/src/common/bridgePowerController_ut.cpp
```diff
@@ -154,23 +156,23 @@ TEST_F(BridgePowerControllerTest, subsampling1) {
   const uint32_t kNineteenMinutes = 1140000;
   const uint32_t kFiveMinutes = 300000;
   const uint32_t kOneMinute = 60000;
-  BridgePowerController powerController(FAKE_VBUS_EN, kTwentyMinutes, kNineteenMinutes,
-                                        kFiveMinutes, kOneMinute, true, true);
+  BridgePowerController powerController(FAKE_VBUS_EN, FAKE_BOOST_EN, kTwentyMinutes,
```
LarkspurRaven: Adding the `BridgePowerControllerCfg` mentioned above and passing allows us to more easily and clearly define many other test configs for testing. [F1]

test/src/common/bridgePowerController_ut.cpp
```diff
@@ -154,23 +156,23 @@ TEST_F(BridgePowerControllerTest, subsampling1) {
...
+                                        kNineteenMinutes, kFiveMinutes, kOneMinute, true, true);
   powerController._update();
   // Init sequence powers the bus on for two minutes
   EXPECT_EQ(fake_io_read_func_fake.call_count, 1);
-  EXPECT_EQ(fake_io_write_func_fake.call_count, 1);
+  EXPECT_EQ(fake_io_write_func_fake.call_count, 2);
   EXPECT_EQ(fake_io_write_func_fake.arg1_history[0], 1);
   // 2 min + MIN_TASK_SLEEP_MS
-  uint32_t curtime = 2 * kOneMinute + 1000;
+  uint32_t curtime = 2 * kOneMinute + 1100;
```
LarkspurRaven: I don't want to do math here. Let's add a magic number. Actually, w/o further checkign, I have no context for the units of 1100. I'll assume msec. Also recommend to change `curtime` to include the time units `curtimeMs` [A6]

```
const uint32_t k_1_min_ms = 60000;

const uint32_t k_1p0_sec_ms = 1000;
const uint32_t k_1p1_sec_ms = 1100;

const uint32_t k_2min_1p0_sec_ms = 2 * k_1_min_ms + k_1p0_sec_ms;
const uint32_t k_2min_1p1_sec_ms = 2 * k_1_min_ms + k_1p1_sec_ms;

uint32_t curtimeMs = k_2min_1p1_sec_ms;
```

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -119,7 +121,14 @@ void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
     EventBits_t signal_to_set = (on) ? BridgePowerController::ON : BridgePowerController::OFF;
     EventBits_t signal_to_clear = (on) ? BridgePowerController::OFF : BridgePowerController::ON;
     xEventGroupClearBits(_busPowerEventGroup, signal_to_clear);
-    IOWrite(&_BusPowerPin, on);
+    if (on) {
+      IOWrite(&_BoostPowerPin, on);
+      vTaskDelay(100); // Allow time for capacitors to charge
```
LarkspurRaven: We should use a magic number for the time delay. Later I suggest this for more clearly defining time constants
```
const uint32_t k_100_msec_ms = 100;
```
But we should then have a more specific time value defined for use here and in the tests [A5]
```
// Delay to allow capacitors to charge
const uint32_t DELAY_POWER_BUS_UP_DELAY = k_100_msec_ms;
```

test/src/common/bridgePowerController_ut.cpp
```diff
@@ -195,18 +197,18 @@ TEST_F(BridgePowerControllerTest, subsampling1) {
   rtcGetMicroSeconds_fake.return_val = 1713232800000000;
   powerController._update();
   EXPECT_EQ(fake_io_read_func_fake.call_count, 5);
-  EXPECT_EQ(fake_io_write_func_fake.call_count, 3);
+  EXPECT_EQ(fake_io_write_func_fake.call_count, 6);
   // The bus stays on for one minute until the next Subample Off time
-  curtime += kOneMinute;
+  curtime += (kOneMinute + 100);
```
LarkspurRaven: use constant -- maybe add comment for why we are adding the extra 100 msec? Here and elsewhere [A7]
```
const uint32_t k_100_msec_ms = 100;

curtimeMs += (k_1_min_ms + k_100_msec_ms);
``

Comment (LarkspurRaven) [reply test/src/common/bridgePowerController_ut.cpp]: Oh I think this maps to the delay added in `powerBusAndSetSignal`. Up there I suggested adding this
`const uint32_t DELAY_POWER_BUS_UP_DELAY = k_100_msec_ms;
`
If this were defined in the header, it could be shared. [A7]

File: test/src/common/bridgePowerController_ut.cpp [hunk 8]
```diff
@@ -217,18 +219,18 @@ TEST_F(BridgePowerControllerTest, subsampling1) {
   rtcGetMicroSeconds_fake.return_val = 1713233100000000;
   powerController._update();
   EXPECT_EQ(fake_io_read_func_fake.call_count, 7);
-  EXPECT_EQ(fake_io_write_func_fake.call_count, 5);
-  EXPECT_EQ(fake_io_write_func_fake.arg1_history[4], 1);
-  curtime += kOneMinute;
+  EXPECT_EQ(fake_io_write_func_fake.call_count, 10);
```
LarkspurRaven: I'm debating whether we should add magic numbers for all the `call_count`, `arg1_history`. Not sure yet. Without inspecting the code path, it's hard to follow the expected counts. It might be helpful to add the magic numbers and explain where the numbers come from. Still pondering... [F4]

src/apps/bridge/bridgePowerController.cpp
LarkspurRaven: I'm assuming that the addition to powerBusAndSetSignal in the `if (on) ...` is the meat of the change that reduces power, but it's not clear why. Please add comments for that context. If it exists, would be good to link to a Notion doc or SC ticket with more info. [A4]

LarkspurRaven (commented):
@victorsowa12 I'm ok with the code changes you made, but I added a number of comments that aren't specifically targeted to your changes here, though related. You pretty much followed the existing patterns used, which is appropriate. My comments are largely motivated by my initial impressions from reading the code. 

Consequently, many of my comments are not blocking for this PR. However, we should at least discuss these suggestions and either make changes now, or create an SC ticket for a quick followup.

victorsowa12:
In a soak I ran overnight, there was no decrease in performance with a 15ms delay in terms on number of readings for a SOFT equipped system. On the left we have the system with BOOST toggling and we consistantly had 56 SOFT readings per sample + subsample duration. On the right we have the same system from before BOOST toggling was added and we had around 51 readings per same duration. With both in a very close ballpark I don't think there is any performance degredation using the 15ms delay.
Screenshot 2025-08-22 at 1 53 37 PM

Marking as ready for review. @LarkspurRaven I'll look over your comments today or monday!

test/src/common/bridgePowerController_ut.cpp
victorsowa12: I could create a tracking variable and a IO_WRITE_NUM_CALLS_INCREMENTER type of vairable that = 2, since we call write twice per ON/OFF toggle.

LarkspurRaven: that'd be interesting

src/apps/bridge/bridgePowerController.cpp
victorsowa12: I'll add a comment! Adding links to notion/sc is not a good idea since this is a public repo and others who may look at this will not have access to those links

src/apps/bridge/bridgePowerController.cpp
victorsowa12: I added a comment before the function to explain what is happening. I also renamed some variables and the function to hopefully make it clearer what they are used for. I didn't break out the IOWrites into enable/disable functions since they are only a few short lines and only used here. Lmk what you think! These changes should also cover A1/2 from the prior comment.

src/apps/bridge/bridgePowerController.cpp
victorsowa12: I tackled A1/2 in the changes related to the comment below. For A3, I don't think mentioning the boost in the function name is necessary since I believe setting the boost and load switch should always happen together. So the "setting of bus power" should always entail toggling both pins. I could rename the `_BusPowerPin` to `_BusLoadSwitchEnablePin` and the `_BoostPowerPin` to `_BoostEnablePin` to help differentiate between what the pins are actuall toggling and the outcomes of toggling them?

test/src/common/bridgePowerController_ut.cpp
victorsowa12: added variables that can be used to increment number of calls tracking variables after each event. LMK what you think!

matt001k (approved):
Nice changes and good work. I enjoyed the configuration for the PowerController as well as all of the added comments.

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -17,17 +17,27 @@
...
-      _subsampleDurationS(subsampleDurationMs / 1000), _sampleIntervalStartS(0),
-      _subsampleIntervalStartS(0), _alignmentS(alignmentS),
-      _ticksSamplingEnabled(ticksSamplingEnabled), _timebaseSet(false), _initDone(false),
-      _subsamplingEnabled(subsamplingEnabled), _configError(false) {
+BridgePowerController::BridgePowerController(const Config &config)
+    : _BusLoadSwitchEnablePin(config.BusLoadSwitchEnablePin), _BoostEnablePin(config.BoostEnablePin),
+      _powerControlEnabled(config.powerControllerEnabled),
+      _sampleIntervalS(config.sampleIntervalMs / 1000),
+      _sampleDurationS(config.sampleDurationMs / 1000),
+      _subsampleIntervalS(config.subsampleIntervalMs / 1000),
```
LarkspurRaven, line 25: why keep these params in the BridgePowerController and in BridgePowerControllerCfg? Violates DRY (don't repeat yourself).

src/apps/bridge/app_main.cpp
```diff
@@ -413,14 +413,21 @@ static void defaultTask(void *parameters) {
...
-      static_cast<bool>(pwrcfg.bridgePowerControllerEnabled),
-      (pwrcfg.alignmentInterval5Min * BridgePowerController::ALIGNMENT_INCREMENT_S),
-      static_cast<bool>(pwrcfg.ticksSamplingEnabled));
+  BridgePowerController::Config config = {
+      .BusLoadSwitchEnablePin = VBUS_SW_EN,
+      .BoostEnablePin = BOOST_EN,
+      .sampleIntervalMs = pwrcfg.sampleIntervalMs,
+      .sampleDurationMs = pwrcfg.sampleDurationMs,
+      .subsampleIntervalMs = pwrcfg.subsampleIntervalMs,
+      .subsampleDurationMs = pwrcfg.subsampleDurationMs,
```
LarkspurRaven, line 423: hmm...why store and copy these variables and not just use power_config_s directly? 

config.pwrcfg.sampleInterval

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -17,17 +17,27 @@
...
+  _busPowerEventGroup = xEventGroupCreate();
+  configASSERT(_busPowerEventGroup);
+  BaseType_t rval = xTaskCreate(BridgePowerController::powerControllerRun, "Power Controller",
+                                128 * 4, this, BRIDGE_POWER_TASK_PRIORITY, &_task_handle);
+  configASSERT(rval == pdTRUE);
+
+  configASSERT(power_info_service_init(powerInfoStatsCb, this) == BmOK);
+}
+
+void BridgePowerController::validateConfig(void) {
```
LarkspurRaven, line 40: validateConfig should be part of BridgePowerController::Config

src/apps/bridge/bridgePowerController.h
```diff
@@ -9,20 +9,28 @@
...
+    uint32_t sampleDurationMs = DEFAULT_SAMPLE_DURATION_S * 1000;
+    uint32_t subsampleIntervalMs = DEFAULT_SUBSAMPLE_INTERVAL_S * 1000;
+    uint32_t subsampleDurationMs = DEFAULT_SUBSAMPLE_DURATION_S * 1000;
+    bool subsamplingEnabled = static_cast<bool>(DEFAULT_SUBSAMPLE_ENABLED);
+    bool powerControllerEnabled = static_cast<bool>(DEFAULT_POWER_CONTROLLER_ENABLED);
+    uint32_t alignmentS = DEFAULT_ALIGNMENT_S;
+    bool ticksSamplingEnabled = static_cast<bool>(DEFAULT_TICKS_SAMPLING_ENABLED);
+  };
+
+  explicit BridgePowerController(const Config &config);
```
LarkspurRaven, line 25: love addition of explicit!

src/apps/bridge/bridgePowerController.h
```diff
@@ -9,20 +9,28 @@
...
 class BridgePowerController {
 public:
-  explicit BridgePowerController(
-      IOPinHandle_t &BusPowerPin, uint32_t sampleIntervalMs = DEFAULT_SAMPLE_INTERVAL_S * 1000,
-      uint32_t sampleDurationMs = DEFAULT_SAMPLE_DURATION_S * 1000,
-      uint32_t subsampleIntervalMs = DEFAULT_SUBSAMPLE_INTERVAL_S * 1000,
-      uint32_t subsampleDurationMs = DEFAULT_SUBSAMPLE_DURATION_S * 1000,
-      bool subsamplingEnabled = false, bool powerControllerEnabled = false,
-      uint32_t alignmentS = DEFAULT_ALIGNMENT_S, bool ticksSamplingEnabled = false);
+  struct Config {
```
LarkspurRaven, line 12: Add comment for Config. Like that you picked struct. I think the validate should live in Config.

src/apps/bridge/bridgePowerController.h
```diff
@@ -9,20 +9,28 @@
...
+    IOPinHandle_t &BusLoadSwitchEnablePin;
+    IOPinHandle_t &BoostEnablePin;
+    uint32_t sampleIntervalMs = DEFAULT_SAMPLE_INTERVAL_S * 1000;
+    uint32_t sampleDurationMs = DEFAULT_SAMPLE_DURATION_S * 1000;
+    uint32_t subsampleIntervalMs = DEFAULT_SUBSAMPLE_INTERVAL_S * 1000;
```
LarkspurRaven, line 17: sampleIntervalMs ... subsampleDurationMs could be changed to power_config_t

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -260,25 +302,25 @@ void BridgePowerController::_update(void) {
...
+          setBusPowerAndSetSignal(true);
         }
-      } else if (!_timebaseSet && _powerControlEnabled && IORead(&_BusPowerPin, &enabled)) {
+      } else if (!_timebaseSet && _powerControlEnabled && IORead(&_BusLoadSwitchEnablePin, &enabled)) {
         if (enabled) { // If our timebase is not set and we've enabled the power manager, we should disable the VBUS
           bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
                          "Bridge State Disabled - controller enabled, but timebase is not yet "
                          "set - bus off\n");
-          powerBusAndSetSignal(false);
+          setBusPowerAndSetSignal(false);
```
LarkspurRaven, line 323: would prefer more descriptive parameter than bool true/false. Intent/expectation is still not clear

Alternatively, can just have two wrapper functions
setBusPowerAndSetSignalOn();
setBusPowerAndSetSignalOff();

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -110,34 +113,73 @@ bool BridgePowerController::waitForSignal(bool on, TickType_t ticks_to_wait) {
...
+ *
+ * \param[in] busOn - true to turn on bus power, false to turn off
+ * \param[in] notifyL2 - true to notify L2 network layer of power changes (default: true)
+ *
+ * \note The boost converter must be turned on before the load switch to allow
+ *       capacitor charging time. See PR#335 for detailed analysis.
+ * \note Port 2 is automatically disabled during power-up as it's unused on bridge
+ *       hardware and disabling provides power savings.
+ */
+void BridgePowerController::setBusPowerAndSetSignal(bool busOn, bool notifyL2) {
```
LarkspurRaven, line 143: Should we return an error code here? 

I think several items can fail: IOWrite, IORead, xEventGroupClearBits

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -110,34 +113,73 @@ bool BridgePowerController::waitForSignal(bool on, TickType_t ticks_to_wait) {
   return rval;
 }
 
-void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
+/*!
+ * \brief Controls bridge bus power with proper sequencing and event signaling.
+ *
+ * Power-up sequence (when busOn=true):
+ * 1. Turn on boost converter first
```
LarkspurRaven, line 120: Can you add a little more to explain why the boost is powered on first?

LarkspurRaven: Oh -- see it in the note below. Maybe just add this

```
 * 1. Turn on boost converter first (see note 1 below)
 ```

File: src/apps/bridge/bridgePowerController.cpp [hunk 18]
```diff
@@ -110,34 +113,73 @@ bool BridgePowerController::waitForSignal(bool on, TickType_t ticks_to_wait) {
...
+ *
+ * Power-down sequence (when busOn=false):
+ * 1. Turn off load switch first to disconnect the load
+ * 2. Turn off boost converter (saves ~3mW)
+ *
+ * The function also:
+ * - Checks current pin state to avoid unnecessary operations
+ * - Sets/clears event group bits to signal power state changes to waiting tasks
+ * - Optionally notifies L2 network layer of power state changes
+ * - Disables unused port 2 during power-up for power savings
```
LarkspurRaven, line 132: add (see note 2 below)

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -110,34 +113,73 @@ bool BridgePowerController::waitForSignal(bool on, TickType_t ticks_to_wait) {
   return rval;
 }
 
-void BridgePowerController::powerBusAndSetSignal(bool on, bool notifyL2) {
+/*!
+ * \brief Controls bridge bus power with proper sequencing and event signaling.
```
LarkspurRaven, line 117: pretty good description, ty

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -110,34 +113,73 @@ bool BridgePowerController::waitForSignal(bool on, TickType_t ticks_to_wait) {
...
+ *
+ * Power-up sequence (when busOn=true):
+ * 1. Turn on boost converter first
+ * 2. Wait for capacitors to charge (CAPACITOR_CHARGE_DELAY_MS)
+ * 3. Turn on load switch to enable bus power
+ *
+ * Power-down sequence (when busOn=false):
+ * 1. Turn off load switch first to disconnect the load
```
LarkspurRaven, line 125: can you add a note for why the load is disconnected before the boost is turned off?

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -110,34 +113,73 @@ bool BridgePowerController::waitForSignal(bool on, TickType_t ticks_to_wait) {
...
+ * 2. Turn off boost converter (saves ~3mW)
```
LarkspurRaven, line 126: Can you add a note for the leading theory on why/how 3mW is saved?

src/apps/bridge/bridgePowerController.cpp
LarkspurRaven, line 132: ` * The function also:`

hmm, that may be too much. Consider extracting each of those steps into named functions

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -186,7 +228,7 @@ void BridgePowerController::_update(void) {
             MAX((_sampleIntervalStartS - currentCycleS) * 1000, MIN_TASK_SLEEP_MS);
         // The default state until first sample interval depends
         // on whether bus power control is enabled.
-        powerBusAndSetSignal(!_powerControlEnabled);
+        setBusPowerAndSetSignal(!_powerControlEnabled);
       }
       _initDone = true;
     } else if (_powerControlEnabled && _timebaseSet) { // Sampling Enabled
```
LarkspurRaven, line 234: Not something that you need to do, but planting seeds for how this huge if/else section in update() could be cleaned up. This is a staged approach. 

for each of these if/else clauses, the conditions could be wrapped into helper functions
```
if (!_initDone) --> if (isInitDone()  // yes it is simple now, but this could be more complicated later
} else if (_powerControlEnabled && _timebaseSet) ==> if (isSamplingEnabled())
```
then....
the code under each if/else block can be converted to a handler function
```
if (isInitDone() {
   handleInit();
else if (isSamplingEnabled()) {
   handleSamplingEnabled(); // not the best name, but whatever it is doing
} ...
```
From there, we map the if/else and handlers into something like this (using modern C++)
```
// shouldn't need params b/c we have Config object 
// if params are needed, can pass void* object
using Checker = std::function<bool()>; /
using Handler = std::function<Result()>;

struct UpdateHandler {
    Checker checker;
    Handler handler;
};

constexpr size_t NumHandlers = 2;

std::array<UpdateHandler, NumHandlers> updateCheckHandlers = {{
    { isInitDone, handleInit },
    { isSamplingEnabled, handleSamplingEnabled }
}};

Result res = DEFAULT_RESULT;
bool handled = false;

for (const auto& h : updateCheckHandlers) {
    if (h.checker()) {
        res = h.handler();
        handled = true;
        break;
    }
}
```

LarkspurRaven (commented):
nice updates, @victorsowa12 !!

a few more comments. I'm good with the PR as is, so long as we follow up on the comments (create a ticket). I was thrilled with all the changes you made -- much more than I had expected, as many of the comments were more for discussion. 
🚢 ⛵ 🦀 🦐

src/apps/bridge/app_main.cpp
```diff
@@ -413,14 +413,21 @@ static void defaultTask(void *parameters) {
...
   printf("Using bridge power controller.\n");
-  IOWrite(&BOOST_EN, 1);
   power_config_s pwrcfg = getPowerConfigs();
-  BridgePowerController bridge_power_controller(
-      VBUS_SW_EN, pwrcfg.sampleIntervalMs, pwrcfg.sampleDurationMs, pwrcfg.subsampleIntervalMs,
-      pwrcfg.subsampleDurationMs, static_cast<bool>(pwrcfg.subsampleEnabled),
-      static_cast<bool>(pwrcfg.bridgePowerControllerEnabled),
-      (pwrcfg.alignmentInterval5Min * BridgePowerController::ALIGNMENT_INCREMENT_S),
-      static_cast<bool>(pwrcfg.ticksSamplingEnabled));
+  BridgePowerController::Config config = {
```
LarkspurRaven, line 417: Though I said Config, I think `Context ctx` makes more sense

src/apps/bridge/bridgePowerController.cpp
LarkspurRaven: I would
* remove all the members that are now in Context from this class
* make X a private member `Context _ctx`

Though we'd have to append `_ctx` to the former private members.

src/apps/bridge/bridgePowerController.cpp
```diff
@@ -17,17 +17,27 @@
...
+BridgePowerController::BridgePowerController(const Config &config)
+    : _BusLoadSwitchEnablePin(config.BusLoadSwitchEnablePin), _BoostEnablePin(config.BoostEnablePin),
+      _powerControlEnabled(config.powerControllerEnabled),
+      _sampleIntervalS(config.sampleIntervalMs / 1000),
+      _sampleDurationS(config.sampleDurationMs / 1000),
+      _subsampleIntervalS(config.subsampleIntervalMs / 1000),
+      _subsampleDurationS(config.subsampleDurationMs / 1000), _sampleIntervalStartS(0),
+      _subsampleIntervalStartS(0), _alignmentS(config.alignmentS),
+      _ticksSamplingEnabled(config.ticksSamplingEnabled), _timebaseSet(false), _initDone(false),
+      _subsamplingEnabled(config.subsamplingEnabled), _configError(false) {
```
LarkspurRaven, line 29: these can also be added to context:

_timebaseSet
_initDone
_configError

maybe _busPowerEventGroup, etc

src/apps/bridge/bridgePowerController.h
```diff
@@ -61,13 +69,13 @@ class BridgePowerController {
...
   static constexpr uint32_t DEFAULT_ALIGNMENT_5_MIN_INTERVAL = (1);
   static constexpr uint32_t DEFAULT_TICKS_SAMPLING_ENABLED = (0);
-
-private:
+  static constexpr uint32_t CAPACITOR_CHARGE_DELAY_MS = 15;
   static constexpr uint32_t MIN_TASK_SLEEP_MS = (1000);
   static constexpr uint32_t INIT_POWER_ON_TIMEOUT_MS = (2 * 60 * 1000);
 
 private:
-  IOPinHandle_t &_BusPowerPin;
```
LarkspurRaven, line 70: move all these private members to Context

src/apps/bridge/app_main.cpp
```diff
@@ -413,14 +413,21 @@ static void defaultTask(void *parameters) {
...
+      .BusLoadSwitchEnablePin = VBUS_SW_EN,
```
LarkspurRaven, line 418: {BusLoadSwitchEnablePin | BoostEnablePin}  --> {busLoadSwitchEnablePin | boostEnablePin}

src/apps/bridge/app_main.cpp
```diff
@@ -413,14 +413,21 @@ static void defaultTask(void *parameters) {
...
+      .BoostEnablePin = BOOST_EN,
+      .sampleIntervalMs = pwrcfg.sampleIntervalMs,
+      .sampleDurationMs = pwrcfg.sampleDurationMs,
+      .subsampleIntervalMs = pwrcfg.subsampleIntervalMs,
+      .subsampleDurationMs = pwrcfg.subsampleDurationMs,
+      .subsamplingEnabled = static_cast<bool>(pwrcfg.subsampleEnabled),
+      .powerControllerEnabled = static_cast<bool>(pwrcfg.bridgePowerControllerEnabled),
+      .alignmentS =
+          (pwrcfg.alignmentInterval5Min * BridgePowerController::ALIGNMENT_INCREMENT_S),
+      .ticksSamplingEnabled = static_cast<bool>(pwrcfg.ticksSamplingEnabled),
```
LarkspurRaven, line 428: What does the trailing `S` mean in `alignmentS` and `ALIGNMENT_INCREMENT_S` ? Thought there'd be a comment in the header for `ALIGNMENT_INCREMENT_S`, but there isn't.

PR #336, opened 2025-08-21
Fix RTC Back In Time Events

matt001k:
## What changed?
Sets milliseconds when assigning the RTC time from a remote clock.
Adds mutex protection to RTC peripheral
Bypasses shadow registers when reading RTC values.

When RTC to,e is being reported to the Spotter platform, there is a chance for the time to jump back a whole second.
Setting the TIME and DATE registers on the STM32U5 resets the SSR (Sub-Seconds Register), accounting for a sub-second (ms) error whenever the RTC time is set.
This PR ensures that the error is minimized by setting the sub-second shifts.

Mutex protection is also added as well to ensure that two separate threads do not attempt to access the RTC at the same time.

Bypassing the shadow register's ensure there is no potential race condition of getting the TIME and DATE register where there could be a potential rollover in months between getting the TIME and DATE which would create very strange behavior (thanks @campbellaos for this one!) by reading the register multiple times until the previous and current read registers match.

There are a lot of edge cases.
Please review [AN4759 Section 2.6](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://www.st.com/resource/en/application_note/an4759-introduction-to-using-the-hardware-realtime-clock-rtc-and-the-tamper-management-unit-tamp-with-stm32-mcus-stmicroelectronics.pdf&ved=2ahUKEwjNrqTAsJyPAxVWDTQIHWTNDcsQFnoECBYQAQ&usg=AOvVaw1EtvsXQ5KVdXJipO9GEAMe), this explains how the RTC clock's sub-seconds are shifted (SPOILER, it is not very trivial).
When shifting milliseconds forward (ex: remotely reported milliseconds are greater than the device's), by setting the `ADD1S` in `RTC_SHIFTR` register, there is a chance that time rolls over to the next:

- minute
- hour
- day
- month
- year

 please see the function `decrement_one_second`, which attempts to address all of these edge cases listed above (I have tested them manually with the `rtc set` CLI command).

## Checklist


src/apps/bridge/app_main.cpp
```diff
@@ -283,8 +283,15 @@ static void handle_subscriptions(uint64_t node_id, const char *topic, uint16_t t
...
-               rtc_time.day, rtc_time.hour, rtc_time.minute, rtc_time.second, rtc_time.ms);
+        bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, true,
+                       "Updating RTC to %u-%u-%u %02u:%02u:%02u.%-04u\n", rtc_time.year,
+                       rtc_time.month, rtc_time.day, rtc_time.hour, rtc_time.minute,
+                       rtc_time.second, rtc_time.ms);
+        rtcGet(&rtc_time);
+        bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_INFO, true,
+                       "RTC is %u-%u-%u %02u:%02u:%02u.%-04u\n", rtc_time.year, rtc_time.month,
+                       rtc_time.day, rtc_time.hour, rtc_time.minute, rtc_time.second,
+                       rtc_time.ms);
```
matt001k: This can be removed, was used for debugging.

src/lib/drivers/stm32_rtc.c
```diff
@@ -1,13 +1,32 @@
...
 #include <stdio.h>
 #include <string.h>
 
 // Magic number to write into backup register 0 to indicate that the rtc is set
 #define RTC_SET_MAGIC 0x836A20DD
 
-static const uint8_t monthDays[]={31,28,31,30,31,30,31,31,30,31,30,31}; // API starts months from 1, this array starts from 0
+static const uint8_t monthDays[] = {
+    31, 28, 31, 30, 31, 30,
+    31, 31, 30, 31, 30, 31}; // API starts months from 1, this array starts from 0
```
LarkspurRaven, line 13: move comment 

// API starts months from 1, this array starts from 0
static const uint8_t monthDays[] = {

victorsowa12 (approved):
🔥⏱️

towynlin (approved):
Nice work. 🙏 🕐

src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
```diff
@@ -36,11 +36,12 @@ void AanderaaSensor::aanderaSubCallback(uint64_t node_id, const char *topic, uin
   (void)version;
   printf("Aanderaa data received from node %016" PRIx64 " On topic: %.*s\n", node_id, topic_len,
          topic);
-  Aanderaa_t *aanderaa = static_cast<Aanderaa_t *>(sensorControllerFindSensorById(node_id, SENSOR_TYPE_AANDERAA));
+  Aanderaa_t *aanderaa =
```
LarkspurRaven, line 39: @matt001k I think you may need to rebase, since there is Aanderaa stuff here

src/lib/drivers/stm32_rtc.c
```diff
@@ -1,13 +1,32 @@
...
+static BmSemaphore rtc_mutex = NULL;
+
+static uint16_t calculate_rtc_ms(void) {
```
LarkspurRaven, line 16: add comment for function

src/lib/drivers/stm32_rtc.c
```diff
@@ -1,13 +1,32 @@
...
 #include <string.h>
 
 // Magic number to write into backup register 0 to indicate that the rtc is set
 #define RTC_SET_MAGIC 0x836A20DD
 
-static const uint8_t monthDays[]={31,28,31,30,31,30,31,31,30,31,30,31}; // API starts months from 1, this array starts from 0
+static const uint8_t monthDays[] = {
+    31, 28, 31, 30, 31, 30,
+    31, 31, 30, 31, 30, 31}; // API starts months from 1, this array starts from 0
+static BmSemaphore rtc_mutex = NULL;
```
LarkspurRaven, line 14: add comment to explain what rtc_mutex it is protecting. Seems to be the `RTC` object.

src/lib/drivers/stm32_rtc.c
```diff
@@ -1,13 +1,32 @@
...
+
+static uint16_t calculate_rtc_ms(void) {
+  uint32_t subSecond = LL_RTC_TIME_GetSubSecond(RTC);
+  uint32_t prediv = LL_RTC_GetSynchPrescaler(RTC);
+  uint32_t diff = prediv - subSecond;
+  uint32_t ret = 0;
+
+  if (prediv < subSecond) {
```
LarkspurRaven, line 22: please add comment for what it means when `prediv < subSecond` an what the calculation is doing. Magic number for 2U?

src/lib/drivers/stm32_rtc.c
```diff
@@ -1,13 +1,32 @@
...
+    diff = 2U * prediv - subSecond;
+  }
+
+  ret = (1000 * diff) / (prediv + 1);
```
LarkspurRaven, line 26: please add comment for calculation. Perhaps a reference to the stm manual? Add magic number for 1000.

src/lib/drivers/stm32_rtc.c
```diff
@@ -30,15 +49,30 @@ BaseType_t rtcInit() {
...
   LL_RTC_InitTypeDef RTC_InitStruct = {LL_RTC_HOURFORMAT_24HOUR, 127, 255};
-  if(LL_RTC_Init(RTC, &RTC_InitStruct) != SUCCESS) {
+  if (LL_RTC_Init(RTC, &RTC_InitStruct) != SUCCESS) {
     rval = pdFAIL;
   }
 
+  // Disable reading from shadow register as it is updated at 1Hz and may cause
+  // concurrency problems
+  LL_RTC_DisableWriteProtection(RTC);
+  LL_RTC_EnableShadowRegBypass(RTC);
```
LarkspurRaven, line 59: create a helper function disableShadowRegisterReading(RTC_TypeDef *RTCx). Perhaps enableShadowRegisterReading(RTC_TypeDef *RTCx)

src/lib/drivers/stm32_rtc.c
```diff
@@ -203,34 +306,57 @@ uint64_t rtcGetMicroSeconds(RTCTimeAndDate_t *timeAndDate){
   return microseconds;
 }
 
-uint32_t getCALM(){
-  return LL_RTC_CAL_GetMinus(RTC);
+uint32_t getCALM() {
```
LarkspurRaven, line 309: like these helpers getCALP, getRECALP, etc. Although they at least need comments for what CALM, CALP, etc are. Or name the functions for clearly. The LL methods are slightly better named LL_RTC_CAL_GetMinus, so I infer the M is minus, but I don't still don't know what minus or CAL is w/o looking it up

src/lib/drivers/stm32_rtc.c
```diff
@@ -246,24 +372,15 @@ bool logRtcGetTimeStr(char *timeStr, size_t len, bool epoch) {
...
   if (rtcGet(&timeAndDate) == pdPASS) {
-    if(epoch) {
-      uint32_t time = utcFromDateTime(timeAndDate.year,
-                                      timeAndDate.month,
-                                      timeAndDate.day,
-                                      timeAndDate.hour,
-                                      timeAndDate.minute,
-                                      timeAndDate.second);
+    if (epoch) {
+      uint32_t time = utcFromDateTime(timeAndDate.year, timeAndDate.month, timeAndDate.day,
```
LarkspurRaven, line 376: I'm curious why utcFromDateTime does not use RTCTimeAndDate_t?

LarkspurRaven: Oh...well utc can be sourced from other things besides RTC. Though still wonder why the util::utc doesn't have an analog utcTimeAndDate, then we could translate.

src/lib/drivers/stm32_rtc.c
```diff
@@ -246,24 +372,15 @@ bool logRtcGetTimeStr(char *timeStr, size_t len, bool epoch) {
...
     } else {
-      numChars = snprintf(timeStr, len, "%04d-%02d-%02dT%02d:%02d:%02d.%03uZ",
-                          timeAndDate.year,
-                          timeAndDate.month,
-                          timeAndDate.day,
-                          timeAndDate.hour,
-                          timeAndDate.minute,
-                          timeAndDate.second,
-                          timeAndDate.ms);
+      numChars = snprintf(timeStr, len, "%04d-%02d-%02dT%02d:%02d:%02d.%03uZ", timeAndDate.year,
```
LarkspurRaven, line 381: consider: extract sprintf into helper `formatRtcTimeAndDate(TimeAndDate_t *timeAndDate)`

src/lib/drivers/stm32_rtc.c
```diff
@@ -182,19 +285,19 @@ uint64_t rtcGetMicroSeconds(RTCTimeAndDate_t *timeAndDate){
   microseconds = (timeAndDate->year - 1970) * (SECS_PER_DAY * 365);
   for (i = 1970; i < timeAndDate->year; i++) {
     if (LEAP_YEAR(i - 1970)) {
-      microseconds +=  SECS_PER_DAY;   // add extra days for leap years
+      microseconds += SECS_PER_DAY; // add extra days for leap years
     }
   }
 
   // add days for this year, months start from 1
   for (i = 1; i < timeAndDate->month; i++) {
```
LarkspurRaven, line 293: consdier extracting loop into well-named helper fxn

src/lib/drivers/stm32_rtc.c
```diff
@@ -182,19 +285,19 @@ uint64_t rtcGetMicroSeconds(RTCTimeAndDate_t *timeAndDate){
   microseconds = (timeAndDate->year - 1970) * (SECS_PER_DAY * 365);
   for (i = 1970; i < timeAndDate->year; i++) {
```
LarkspurRaven, line 286: consdier extracting loop into well-named helper fxn

src/lib/drivers/stm32_rtc.c
```diff
@@ -182,19 +285,19 @@ uint64_t rtcGetMicroSeconds(RTCTimeAndDate_t *timeAndDate){
...
     } else {
-      microseconds += SECS_PER_DAY * monthDays[i-1];  //monthDay array starts from 0
+      microseconds += SECS_PER_DAY * monthDays[i - 1]; //monthDay array starts from 0
     }
   }
-  microseconds += (timeAndDate->day-1) * SECS_PER_DAY;
+  microseconds += (timeAndDate->day - 1) * SECS_PER_DAY;
   microseconds += timeAndDate->hour * SECS_PER_HOUR;
   microseconds += timeAndDate->minute * SECS_PER_MIN;
   microseconds += timeAndDate->second;
```
LarkspurRaven, line 303: comment what these last `microseconds +=` are doing or potential for well-named helper

src/lib/drivers/stm32_rtc.c
```diff
@@ -152,14 +239,30 @@ BaseType_t rtcSet(const RTCTimeAndDate_t *timeAndDate) {
...
+  // meaning the ms will always have to be shifted forward
+  uint32_t pre = LL_RTC_GetSynchPrescaler(RTC) + 1U;
+  uint32_t adjust = (1000 * pre - timeAndDate->ms * pre) / 1000U;
+
+  LL_RTC_DisableWriteProtection(RTC);
+  LL_RTC_TIME_Synchronize(RTC, LL_RTC_SHIFT_SECOND_ADVANCE, adjust);
+  LL_RTC_EnableWriteProtection(RTC);
+  // Wait for shift operation pending flag to clear
+  while (LL_RTC_IsActiveFlag_SHP(RTC)) {
+  };
```
LarkspurRaven, line 258: consider: extracting this block under `if (rval == pdPass)` into well named helper

src/lib/drivers/stm32_rtc.c
```diff
@@ -152,14 +239,30 @@ BaseType_t rtcSet(const RTCTimeAndDate_t *timeAndDate) {
...
+
+  if (rval == pdPASS) {
```
LarkspurRaven, line 260: what should the behavior be if `rval == pdFAIL` ?? Should we assert?

LarkspurRaven (approved):
Beautiful investigation @matt001k. This is good to be merged for BM v0.13.4. Please create a follow-up ticket to address my comments.

I think the most important thing is to add some unit tests. 

This rtc code would be perfect for some unit tests to cover the boundary conditions, leap years, etc.

PR #337, opened 2025-08-21
ignore spurious null bytes from half-duplex transaction switching

evanShap:
## What changed?

I ran into an annoying issue/bug when updating the RS-485 Rx-LIVE shark tag receiver demo app to latest BM release for the Stanford Pioneer demo unit.
- After the initial version of this app (v0.12.0), the PLUART line buffer and transaction management system had some minor updates for SDI-12 in v0.12.x revs.
- I couldn't figure out why (and can't sink more time into figuring out why) these changes cause a single `0` byte to be clocked in to RS-485 Half Duplex Rx when we switch from Tx to Rx in an RS-485 HD transaction.
- I don't know if this is an issue with RS-485 HD transactions in general, or idiomatic to this sensor.
- This byte borked a feature – the PLUART Line Processor – that abstracts handling of ASCII serial protocols (every sensor has their own custom flavor of this), because the implementation of that feature makes use of `%s` and `%.*s` string interpolation in eg `sprintf` calls for buffer management and logging.
- This patch PR simply ignores any invalid `00` bytes when adding new bytes to the buffer for this feature. Other PLUART usages (user bytes buffer, raw Rx/Tx, custom user callbacks) are unaffected.

- Does not change anything Bristlemouth core, just this PLUART helper feature in this serial text processing use case.
- The resultant change is good and what we want - if, for whatever reason, a `00` comes over the wire in a ASCII protocol, we should ignore it.
- The implementation feels a little gross. It could be better to do a combination of:
-- Figure out whether this is an artifact of our RS-485 HD Tx/Rx switching, and if so, fix that.
-- Modify the implementation of the PLUART Line Processor so that `00` doesn't mess things up, and can be ignored, warned, errored or even used per a specific implementation's concern. We haven't seen any line schemas that use `00` as the frame delimiter (it's almost always \r, \n or \t), but someone *could* do this silly thing.
- Neither of these are important enough to consider fixing any time soon. However, I did need to get the Rx-LIVE demo app working.

- Is this an acceptable bandaid to merge in? If not, how should we handle *other* than sinking effort into more robust solutions?
- Is the modified code used in any of our production apps, and is there any risk of these changes?

## Checklist


towynlin:
Hm 🤔 it seems to me we should handle this differently. You could imagine someone using this code and having null bytes be valid. Having them as a line terminator in particular seems perfectly sane to me, even if the typical target is ASCII text. I could be wrong.

evanShap:
> Hm 🤔 it seems to me we should handle this differently. You could imagine someone using this code and having null bytes be valid. Having them as a line terminator in particular seems perfectly sane to me, even if the typical target is ASCII text. I could be wrong.

@towynlin - yep, I agree this is not the cleanest way to handle this (I think the detail I captured above captures your concern in more depth). However, there are other places in the lineparser feature that already assume/require data is ascii, so the only case this implementation precludes is ascii data with specifically null byte separator - more generic binary data formats already don't work with line parser lib.

I can't put more time into this in the near term, and I don't think it's a high priority overall in general, but it is **critical** Rx-LIVE users (currently just one, but others are considering testing it out) get this fix. Choices are:
- Merge in as is, add a TODO/issue to fix when/if it becomes an issue for another use case.
- Keep on an isolated branch and update docs to point Rx-LIVE users to that branch for now.
- Shuffle some Sofar FW team priority to do a better implementation now for the sake of hygiene (_one challenge is our test Rx-LIVE was just deployed in the field, so we won't be able to fully test a different implementation_).

towynlin:
Understood. I'll take a closer look today.

towynlin (approved):
Ship it. 👍 Sorry for the pointless delay.

evanShap:
> Ship it. 👍 Sorry for the pointless delay.

Thanks! Not pointless - good to consider these things.

PR #340, opened 2025-09-15
Add Aanderaa Conductivity Sensor

LarkspurRaven:
## What changed?

Add support for the Aanderaa Conductivity Sensor!

Expanding the smart mooring ecosystem!

This PR includes some scaffolding for removing some boiler-plate code and reducing the number of locations for adding new sensors. 

The sensor largely follows the same pattern used when introducing the Seapoint Turbidity Sensor.

## Checklist


src/apps/bridge/sensorController.cpp
```diff
@@ -340,3 +362,22 @@ static bool node_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
   }
   return rval;
 }
+
+static bool initializeConductivitySensorConfig(bool &save) {
+  _ctx.aanderaa_conductivity_reading_period_ms = DEFAULT_AANDERAA_CONDUCTIVITY_READING_PERIOD_MS;
+  if (!get_config_uint(BM_CFG_PARTITION_SYSTEM, AppConfig::AANDERAA_CONDUCTIVITY_READING_PERIOD_MS,
```
LarkspurRaven: pull into Aanderaa?

src/apps/bridge/sensorController.cpp
```diff
@@ -304,6 +314,18 @@ static bool node_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
             abstractSensorAddSensorSub(seapoint_turbidity_sub);
           }
         }
+      } else if (strncmp(reply.app_name, "aanderaa_conductivity",
+                         MIN(reply.app_name_strlen, strlen("aanderaa_conductivity"))) == 0) {
+        if (!sensorControllerFindSensorById(reply.node_id, SENSOR_TYPE_AANDERAA_CONDUCTIVITY)) {
```
LarkspurRaven: pull into Aanderaa?

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,315 @@
...
+#include "device_info.h"
+#include "pubsub.h"
+#include "semphr.h"
+#include "spotter.h"
+#include "stm32_rtc.h"
+#include "topology_sampler.h"
+#include <new>
+
+/** @brief Default sensor reading period in milliseconds */
+#define DEFAULT_AANDERAA_CONDUCTIVITY_READING_PERIOD_MS 1000 // 1 second
```
towynlin: Need to DRY up the duplicate between here and sensorController.cpp:26.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -1,25 +1,28 @@
...
 #include "timers.h"
 #include "topology_sampler.h"
+#include "aanderaaSensor.h"
+// Sensor drivers
+#include "sensor_drivers/aanderaaConductivitySensor.h"
+#include "sensor_drivers/borealisSensor.h"
+#include "sensor_drivers/pmeDissolvedOxygenSensor.h"
+#include "sensor_drivers/rbrCodaSensor.h"
+#include "sensor_drivers/seapointTurbiditySensor.h"
+#include "sensor_drivers/softSensor.h"
```
matt001k, line 24: Why do we need the prepended `sensor_drivers` to the path now? Just to keep them in the same spot when formatting it ran?

src/apps/bridge/reportBuilderList.cpp
```diff
@@ -116,6 +117,10 @@ void ReportBuilderLinkedList::addSampleToElement(report_builder_element_t *eleme
         element->sensor_data))[element->sample_counter];
     break;
   }
+  case SENSOR_TYPE_AANDERAA_CONDUCTIVITY: {
+    AanderaaConductivitySensor::setupSensorPointers(element, &nan_sample, &dst);
```
towynlin: Ben is advocating for a new pattern here that could allow us to clean up this whole switch-case block. Let's review this PR keeping this new pattern, but we should queue up work in the coming weeks to choose one pattern or the other and ensure consistency here. Otherwise we make this code confusing to read and harder to maintain.

towynlin (approved):
Saw one small thing to clean up. Good job.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
towynlin: Also I changed both to 30000 in commit 210083e. Seems to have gotten lost. The default is 30 seconds.

towynlin: Yeah in commit 5aec865 you reintroduced the strncmp bugs I fixed as well.

LarkspurRaven: fixing

src/apps/bridge/sensorController.cpp
LarkspurRaven: I created a helper function `loadSensorConfig()` along with struct `sensor_config_def_t`. This will allow this whole chunk of code to be reduced to just an array of `sensor_config_def_t` for each structs. `sensor_config_def_t` may have overlap with other structs, but will reconsider when we get to other sensors.

src/apps/bridge/sensorController.cpp
LarkspurRaven: I created a helper function `createAndConfigureSensorSubscription()` along with struct `sensor_subscription_config_t`. This will allow this whole chunk of code to be reduced to just an array of `sensor_subscription_config_t` for each structs. `sensor_subscription_config_t` may have overlap with other structs, but will reconsider when we get to other sensors.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -1,25 +1,28 @@
...
 #include "semphr.h"
-#include "softSensor.h"
 #include "task.h"
 #include "task_priorities.h"
 #include "timer_callback_handler.h"
 #include "timers.h"
 #include "topology_sampler.h"
+#include "aanderaaSensor.h"
+// Sensor drivers
+#include "sensor_drivers/aanderaaConductivitySensor.h"
```
LarkspurRaven, line 19: Adding the `sensor_driver` prefix took care of some include warnings in cppcheck. Though cppcheck may not be smart enough or configured well enough to understand the include system of cmake. Worth exploring. Though I actually prefer the prefix for grouping. The other includes could benefit from similar grouping. Hard to tell which are freeRtos, which are bm_core, bm_proto, etc

src/apps/bridge/reportBuilder.cpp
```diff
@@ -66,7 +69,8 @@ static void report_builder_task(void *parameters);
 
 CborError encode_buffer_sample_member(CborEncoder &sample_array, void *sample_member,
                                       uint32_t size) {
-  const uint8_t *local_sample_member = (const uint8_t *)sample_member;
+  const uint8_t *local_sample_member = static_cast<uint8_t *>(sample_member);
```
LarkspurRaven, line 72: Not related to task, but it fixed some cppcheck complaints. Here and elsewhere

src/apps/bridge/reportBuilder.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
 #include "abstractSensor.h"
+#include "cbor_sensor_report_encoder.h"
+#include <vector>
```
LarkspurRaven: vector is only temporary. But planning to do in next PR, after i work on the Spotter code

src/apps/bridge/reportBuilder.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
 #include "abstractSensor.h"
+#include "cbor_sensor_report_encoder.h"
```
LarkspurRaven, line 4: Planning to forward declare items from cbor_sensor_report_encoder.h

src/apps/bridge/reportBuilder.h
```diff
@@ -16,16 +18,112 @@ typedef struct {
...
+ * Contains information needed to construct a sensor report including
+ * context, sensor data, and metadata for CBOR encoding. Used by the report
+ * builder system to process sensor data in a structured way.
+ */
+ // cppcheck-suppress unusedStructMember
+ typedef struct {
+  /// Context for CBOR encoding
+  sensor_report_encoder_context_t &context;
+  /// Pointer to the sensor data to be encoded
+  void *sensor_data;
```
LarkspurRaven, line 44: I'm not crazy about the pattern of `void *` here and elsewhere, but I'm sticking with the pattern. I'd prefer to use a typed base class, but reportBuilder is not consistent in it's use of C/C++. So I didn't feel comfortable to make the that change yet.

src/apps/bridge/sensorController.cpp
```diff
@@ -36,10 +38,41 @@ typedef struct sensorControllerCtx {
...
+typedef struct {
+  /// Pointer to the context field in _ctx that holds the sensor's reading period
+  uint32_t* ctx_field;
+  /// Default value to use if the configuration is not found
+  uint32_t default_value;
+  /// Key to use to store the configuration
+  const char* config_key;
+  /// Name of the sensor to use in log messages
+  const char* sensor_name;
+} sensor_config_def_t;
```
LarkspurRaven: May be able to consolidate `sensor_config_def_t` and `sensor_subscription_config_t`. Perhaps with `sensorsControllerCtx_t` as well. But need to consider further.

src/apps/bridge/sensorController.cpp
```diff
@@ -258,7 +355,22 @@ static bool node_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
...
                       strlen(AppConfig::SUBSAMPLE_ENABLED), &subsample_enabled);
 
-      if (strncmp(reply.app_name, "aanderaa", MIN(reply.app_name_strlen, strlen("aanderaa"))) ==
+      // Define Aanderaa conductivity subscription config
+      static const sensor_subscription_config_t aanderaa_conductivity_config = {
+        "aanderaa_conductivity",
+        SENSOR_TYPE_AANDERAA_CONDUCTIVITY,
+        &_ctx.aanderaa_conductivity_reading_period_ms,
+        AanderaaConductivity_t::N_SAMPLES_PAD,
+        [](uint64_t node_id, uint32_t duration, uint32_t samples) -> AbstractSensor* {
```
LarkspurRaven: Don't think I've seen use of C++ lambda functions in our code, but here we are. Please give your thoughts. I love it b/c I don't need to define another function as a wrapper.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,313 @@
...
+#include "pubsub.h"
+#include "semphr.h"
+#include "spotter.h"
+#include "stm32_rtc.h"
+#include "topology_sampler.h"
+#include <new>
+
+
+bool AanderaaConductivitySensor::subscribe() {
+  bool rval = false;
```
LarkspurRaven, line 53: TODO: As a follow-up, I'd like to extract this pattern to the base class. Much of this code is duplicated across all sensors. We can make a struct that the base class can handle. This would apply to nearly all functions: 
* subscribe
* aanderaaConductivitySubCallback
* aggregate
* createAanderaaConductivitySub --> createSensorSub

And, truly, almost none of that is needed if we go config-driven generated code!

Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.h
```diff
@@ -0,0 +1,213 @@
...
+ * - Integration with bridge reporting system
+ * - Configurable aggregation periods
+ */
+typedef struct AanderaaConductivitySensor : public AbstractSensor {
+  /// Aggregation period in milliseconds
+  uint32_t current_agg_period_ms;
+
+  // Statistical samplers for each sensor parameter
+  /// Conductivity sampler (mS/cm)
+  AveragingSampler conductivity_ms_cm;
```
LarkspurRaven: Not a fan of this pattern of defining flat AveragingSampler's. 
a. They should be in a container
b. AveragingSampler's is too specific. There should be a SamplerBase class so that we can support many other types of Samplers
c. The mapping from the Sampler to the element in the aggregation struct and the Sensor::Data is manual and doesn't need to be

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.h
```diff
@@ -0,0 +1,213 @@
...
+#define AANDERAA_CONDUCTIVITY_NUM_SAMPLE_MEMBERS 6
+
+/**
+ * @brief Aggregated conductivity sensor data structure
+ *
+ * Contains the averaged/aggregated values from multiple conductivity sensor readings.
+ * This structure is used for storing and transmitting aggregated sensor data
+ * in the bridge's report system.
+ */
+typedef struct aanderaa_conductivity_aggregations_s {
```
LarkspurRaven: Not a fan of this pattern
a. thinking we make an AggregationType that can stand-in for the types (double, float, etc) and include a name, default value, range, etc
b. we can then create a container of the AggregationTypes for easier iteration

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.h
```diff
@@ -0,0 +1,213 @@
...
+#include "avgSampler.h"
+#include "sensorController.h"
+#include "reportBuilder.h"
+#include "reportBuilderList.h"
+#include <cmath>
+#include <stdint.h>
+#include <stdlib.h>
+
+/** @brief Number of sensor parameters measured by the Aanderaa conductivity sensor */
+#define AANDERAA_CONDUCTIVITY_NUM_SAMPLE_MEMBERS 6
```
LarkspurRaven: hmm. Should probably make this a class-scoped constexpr. Missed it

src/apps/bridge/topology_sampler.cpp
```diff
@@ -817,7 +817,10 @@ static void _update_sensor_type_list(uint64_t node_id, char *app_name, uint32_t
   (void)app_name_len;
   for (uint8_t i = 0; i < TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE; i++) {
     if (_node_list.nodes[i] == node_id) {
-      if (strncmp(app_name, "aanderaa", strlen("aanderaa")) == 0) {
+      if (strncmp(app_name, "aanderaa_conductivity", strlen("aanderaa_conductivity")) == 0) {
```
LarkspurRaven, line 820: looks like I missed an extraction

src/apps/bridge/reportBuilderList.cpp
LarkspurRaven: I left similar comments elsewhere for similar things

src/apps/bridge/reportBuilder.h
```diff
@@ -16,16 +18,112 @@ typedef struct {
...
+  /// Index of the sample to be encoded
+  uint32_t sample_index;
+  /// Failure text for logging
+  const char *failText;
+  /// Sample type identifier
+  const char *sampleType;
+  /// Number of sample members (fields) in the sensor data
+  uint32_t numSampleMembers;
+  /// Vector of sample member parameters
+  std::vector<sample_member_params_t> sampleMembers;
```
LarkspurRaven: vector is temporary. It's ok to use, but has some inefficiencies with how storage is allocated. Plan to address when I change what is included in the Reports (we don't need all the members/fields)

src/apps/bridge/reportBuilder.h
```diff
@@ -16,16 +18,112 @@ typedef struct {
...
+ // cppcheck-suppress unusedStructMember
+ typedef struct {
+  /// Context for CBOR encoding
+  sensor_report_encoder_context_t &context;
+  /// Pointer to the sensor data to be encoded
+  void *sensor_data;
+  /// Index of the sample to be encoded
+  uint32_t sample_index;
+  /// Failure text for logging
+  const char *failText;
```
LarkspurRaven: would prefer to switch from `const char *` to `std::string` (RAII) or `std::string_view` for additional safety, but need to evaluate further as reportBuilder has a mix of C/C++. So keeping the pattern for now.

src/apps/bridge/reportBuilder.h
```diff
@@ -16,16 +18,112 @@ typedef struct {
...
+ * Convenience function that opens a sample, adds all sample members from
+ * the report parameters, and closes the sample. This is the main function
+ * sensors should use to add their data to reports.
+ *
+ * @param params Report parameters containing sample members and context
+ * @return true if successful, false on encoding error
+ */
+bool report_builder_add_samples(report_params_t &params);
+
 CborError encode_buffer_sample_member(CborEncoder &sample_array, void *sample_member,
```
LarkspurRaven, line 62: These `encode_x_sampler_member` functions may be better suited in another file

src/apps/bridge/reportBuilder.h
```diff
@@ -16,16 +18,112 @@ typedef struct {
...
+  /// Number of sample members (fields) in the sensor data
+  uint32_t numSampleMembers;
+  /// Vector of sample member parameters
+  std::vector<sample_member_params_t> sampleMembers;
+
+} report_params_t;
+
 void reportBuilderInit(void);
+
 void reportBuilderAddToQueue(uint64_t node_id, uint8_t sensor_type, void *sensor_data,
```
LarkspurRaven, line 54: just noticed these functions use camelCase while others use underscore:
* reportBuilderInit
* reportBuilderAddToQueue

src/apps/bridge/reportBuilder.cpp
```diff
@@ -133,6 +138,65 @@ void reportBuilderAddToQueue(uint64_t node_id, uint8_t sensor_type, void *sensor
...
+bool report_builder_add_samples(report_params_t &params) {
+  if (!report_builder_open_sample(params)) {
+    return false;
+  }
+
+  for (auto &s : params.sampleMembers) {
+    if (!report_builder_add_sample_member(params, s)) {
+      return false;
+    }
+  }
```
matt001k: I like the use of `auto` here. Make this API very clean.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -133,6 +138,65 @@ void reportBuilderAddToQueue(uint64_t node_id, uint8_t sensor_type, void *sensor
...
+  return true;
+}
+
+// Encode a single data field from the sensor reading into the CBOR report
+bool report_builder_add_sample_member(report_params_t &params, sample_member_params_t &s) {
+  if (sensor_report_encoder_add_sample_member(params.context, s.sampleMemberEncoderCb,
+                                              s.sampleMember) != CborNoError) {
+    bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_ERROR, USE_HEADER, params.failText);
+    return false;
+  }
```
matt001k: One thing to note, is that the `Borealis` sensor will only add some sample members based on certain conditional statements. This would work nicely for most every sensor. Would the conditionals aforementioned be handled in the `getReportParams` API? Seems logical, but wanted to point it out nonetheless.

src/apps/bridge/reportBuilder.h
matt001k: Good call out. Maybe add a `TODO` so it does not get lost and link it to a create an issue in this repository to track (and a ticket in Shortcut for Sofar tracking)?

src/apps/bridge/reportBuilder.h
```diff
@@ -16,16 +18,112 @@ typedef struct {
...
+
+/**
+ * @brief Open a new sample in the sensor report encoder
+ *
+ * Initializes a new sample entry in the CBOR report with the specified
+ * number of sample members and sample type identifier.
+ *
+ * @param params Report parameters containing context and metadata
+ * @return true if successful, false on encoding error
+ */
```
matt001k: Curious as to the preference for Doxygen like function comments here in the header file rather than source?

src/apps/bridge/reportBuilder.h
```diff
@@ -16,16 +18,112 @@ typedef struct {
...
+ * @brief Add all sample members from a report to the encoder
+ *
+ * Convenience function that opens a sample, adds all sample members from
+ * the report parameters, and closes the sample. This is the main function
+ * sensors should use to add their data to reports.
+ *
+ * @param params Report parameters containing sample members and context
+ * @return true if successful, false on encoding error
+ */
+bool report_builder_add_samples(report_params_t &params);
```
matt001k: Do all of these API need to be public? It looks like currently they are all used internally?

src/apps/bridge/reportBuilderList.cpp
matt001k: It would be good to meet on this again and finalize a way of how things SHOULD be done and DO the work SOON (possibly next Sofar Firmware Iteration?). This way the code can be consistent when new sensors are added and eliminate the risk of another hand in the pot of a semi-refactor.

src/apps/bridge/sensorController.cpp
matt001k: Could you make this `PascalCase`? This paradigm has been adopted from the `bm_core` work for most of `bm_protocol`'s current work. See: https://github.com/bristlemouth/bm_core?tab=readme-ov-file#code-style

src/apps/bridge/sensorController.cpp
matt001k: I don't mind it, what is the reason that `createAanderaaConductivitySub` cannot be passed here itself?
I do not see any casts and it returns a child class of `AbstractSensor`?

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
matt001k: I know in C++ there are no issues with having "open" function parameters in a function. But the C developer in me is screaming, could we use `void` for functions with no parameters.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * @brief Static callback function for handling incoming conductivity sensor data
+ *
+ * This function is called when CBOR-encoded conductivity sensor data is received
+ * via the pub/sub system. It decodes the data and adds samples to the appropriate
+ * sensor's statistical aggregators.
+ *
+ * @param node_id Source node ID
+ * @param topic MQTT topic string
```
matt001k, line 282: This is `Bristlemouth` topic string

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,313 @@
...
+                     reading_time_millis, sensor_reading_time_sec, sensor_reading_time_millis,
+                     composite_cbor_msg.conductivity_ms_cm, composite_cbor_msg.temperature_deg_c,
+                     composite_cbor_msg.salinity_psu, composite_cbor_msg.water_density_kg_m3,
+                     composite_cbor_msg.sound_speed_m_s, composite_cbor_msg.depth_m);
+        if (log_buflen > 0) {
+          BRIDGE_SENSOR_LOG_PRINTN(BM_COMMON_IND, log_buf, log_buflen);
+        } else {
+          printf("ERROR: Failed to print Aanderaa Conductivity data\n");
+        }
+        vPortFree(log_buf);
```
matt001k: There is a helper function that performs a lot of this already called `send_spotter_log_individual` in abstractSensor.h`. Might clean this up a bit.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,313 @@
...
+                 "%.3f\n",                // depth_mean_m
+                 node_id, node_position, time_str, conductivity_aggs.reading_count,
+                 conductivity_aggs.conductivity_mean_ms_cm, conductivity_aggs.temperature_mean_deg_c,
+                 conductivity_aggs.salinity_mean_psu, conductivity_aggs.water_density_mean_kg_m3,
+                 conductivity_aggs.sound_speed_mean_m_s, conductivity_aggs.depth_mean_m);
+    if (log_buflen > 0) {
+      BRIDGE_SENSOR_LOG_PRINTN(BM_COMMON_AGG, log_buf, log_buflen);
+    } else {
+      printf("ERROR: Failed to print Aanderaa Conductivity aggregation to log\n");
+    }
```
matt001k: There is a function called `send_spotter_log_aggregate` in `abstractSensor.cpp` that handles a lot of this logic.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,313 @@
...
+}
+
+report_params_t AanderaaConductivitySensor::getReportParams(sensor_report_encoder_context_t &context, void *sensor_data, uint32_t sample_index) {
+
+  report_params_t params = {
+    .context = context,
+    .sensor_data = sensor_data,
+    .sample_index = sample_index,
+    // TODO: fail text should probably be on sample member
+    .failText = "Failed to open aanderaa_conductivity sample in addSamplesToReport\n",
```
matt001k: I do not really understand this `TODO` could you clear it up for me?

src/lib/bm_ncp/ncp_uart.cpp
```diff
@@ -640,7 +640,8 @@ static BaseType_t ncpRXBytesFromISR(SerialHandle_t *handle, uint8_t *buffer, siz
   configASSERT(len == 1);
 
   // but the byte into the current buffer
-  ncpRXBuff[ncpRXCurrBuff][ncpRXBuffIdx++] = byte;
+  ncpRXBuff[ncpRXCurrBuff][ncpRXBuffIdx] = byte;
+  ncpRXBuffIdx += 1;
```
matt001k, line 644: Could this be removed? Was this accidentally snuck in?

matt001k (changes requested):
Have some comments that I'd like to get addressed. Other than that good work!
In the next coming weeks, the Firmware team here should lock in place how the Report builder should look and put in the work necessary for a proper refactor.

src/apps/bridge/reportBuilder.cpp
LarkspurRaven: It removes warnings from cppcheck

src/apps/bridge/reportBuilder.cpp
LarkspurRaven: comment from @towynlin - if add the pattern, make it the only way; remove path from cmake file

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
LarkspurRaven: fixed. Now read from `AanderaaConductivitySensor::getDefaultReadingPeriodMs()`

src/lib/bm_ncp/ncp_uart.cpp
towynlin: Good question. Thanks for calling it out.

I added this one to get CI passing. C++ 20 deprecated some uses of volatile.
https://eel.is/c++draft/depr.volatile.type

Line 39 of this file declares:
`static volatile uint32_t ncpRXBuffIdx = 0;`

If I understand correctly, the intention of the deprecation is to make code more explicit for volatile variables, where each statement contains only a read or a write, but not both. Here's [a blog post](https://blog.feabhas.com/wp-content/uploads/2021/05/volatile.html) I found on the deprecation, which ends:

> Finally, when developing new code, whether in C or C++, prefer to use the explicit “read-modify-write” to compound statements.

Read-modify-write is what the new code in this PR does. Line 643 reads the volatile `ncpRXBuffIdx`. Line 644 modifies and writes the volatile variable.

That said, my attempts to get CI passing went like this:

1. Why are designated initializers disallowed all of a sudden?
2. Ask AI, which says they're allowed in C++20 but not C++17.
3. Push commit a0f72ff to use C++ 20. Indeed the error goes away.
4. Get this volatile deprecation warning-as-error and fix it in commit 644d919.
5. Fix some other legit new warnings-as-errors.
6. Eventually get the designated initializer error again! Wat?
7. Bite the bullet and refactor report_builder_ut.cpp to assign every member individually as one part of commit 5508b90.

So, we could go back to C++ 17 and we could go back to using the volatile increment, since the impetus to update was trying to avoid refactoring the report builder unit test setup, but we had to do it anyway.

That said, I think we should keep the C++ 20 usage and the future-facing fixes.

Thoughts?

matt001k: I agree, keeping the C++ 20 usage is better. I do like the explicit read, (modify,) write on `volatile` variables. This is usually pretty standard when accessing registers. It also captures a bit more of what would be happening to a `volatile` variable during runtime, no optimizations on the variable causes the variable to be loaded every time that it is accessed.

src/apps/bridge/reportBuilder.h
LarkspurRaven: Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge for follow-up.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
LarkspurRaven: yeah, you are right. This is not portable even for C++. This can cause a silent failure/crash. Good catch!

matt001k (approved):

src/apps/bridge/reportBuilder.cpp
LarkspurRaven: I had a comment above regarding this -- I actually prefer to add it. It takes some work to figure out which headers are bm_protocol, bm_core, FreeRtos, etc. 

I added a note in https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge to reconsider.

src/apps/bridge/reportBuilder.cpp
LarkspurRaven: Yes, this would be delegated to getReportParams().

```
std::vector<sample_member_params_t> AanderaaConductivitySensor::getSampleMemberParams(void *sensor_data, uint32_t sample_index ) {
   aanderaa_conductivity_aggregations_t aanderaa_conductivity_sample = (static_cast<aanderaa_conductivity_aggregations_t *>(sensor_data))[sample_index];
   
   // Build the default 
   std::vector<sample_member_params_t> sampleMemberParams = {
       {.sampleMember = &borealis_data.is_extended,
          .sampleMemberEncoderCb = encode_uint_sample_member},
       {.sampleMember = &borealis_data.spl,
          .sampleMemberEncoderCb = encode_uint_sample_member},
       {.sampleMember = &borealis_data.max_iqr,
          .sampleMemberEncoderCb = encode_uint_sample_member},
       {.sampleMember = &borealis_data.max_iqr_band,
          .sampleMemberEncoderCb = encode_uint_sample_member},
       {.sampleMember = &borealis_data.entropy,
          .sampleMemberEncoderCb = encode_uint_sample_member},
    }

   // Add optional elements
   if (borealis_data.is_extended) {
       sampleMemberParams.push_back({.sampleMember = &borealis_data.spl_band_stats,
          .sampleMemberEncoderCb = encode_buffer_sample_member, .size = borealis_data.spl_band_stats_size});
   }
   if (borealis_data.hydrotwin_ldr.size) {
       sampleMemberParams.push_back({.sampleMember = &borealis_data.hydrotwin_ldr.buf,
          .sampleMemberEncoderCb = encode_buffer_sample_member, .size = borealis_data.hydrotwin_ldr.size});
   }
   
}
   
report_params_t BorealisSensor::getReportParams(sensor_report_encoder_context_t &context, void *sensor_data, uint32_t sample_index) {

  report_params_t params = {
    .context = context,
    .sensor_data = sensor_data,
    .sample_index = sample_index,
    // TODO: fail text should probably be on sample member
    .failText = "Failed to open borealis sample in addSamplesToReport\n",
    .sampleType = "bm_aos_borealis_v0",
    .numSampleMembers = BOREALIS_DEFAULT_NUM_SAMPLE_MEMBERS,
    .sampleMembers = {}  // Initialize empty, fill below
  };

  // Fill the sample members array using the existing function
  getSampleMemberParams(sensor_data, sample_index, params.sampleMembers);

  return params;
}
```

src/apps/bridge/reportBuilder.h
LarkspurRaven: keeping. Even with removing the `report_builder_x()` methods, we need to still need reference to elements in `cbor_sensor_report_encoder.h` for `sample_member_params_t` and `report_params_t`. 

Forward declarations would repeat a lot from `cbor_sensor_report_encoder.h`, which is a fairly lightweight header. 

```
// Forward declarations to avoid including cbor_sensor_report_encoder.h
typedef struct sensor_report_encoder_context sensor_report_encoder_context_t;
typedef enum CborError CborError;  
struct CborEncoder;
typedef CborError (*sample_encoder_cb)(CborEncoder &sample_array, void *sensor_data, uint32_t size);
```

src/apps/bridge/reportBuilder.h
LarkspurRaven: Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge for follow-up.

src/apps/bridge/reportBuilder.h
LarkspurRaven: Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge for follow-up.

src/apps/bridge/reportBuilder.h
LarkspurRaven: at one time I needed them, but they can be private

src/apps/bridge/reportBuilder.h
LarkspurRaven: Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge for follow-up.

src/apps/bridge/reportBuilder.h
LarkspurRaven: It's not so much my preference, but inherited from the Google C++ style guide, which I've long used since my time at Amazon
https://google.github.io/styleguide/cppguide.html#Function_Comments

> **Function Declarations**
> * Almost every function declaration should have comments immediately preceding it that describe what the function does and how to use it. 
> * These comments may be omitted only if the function is simple and obvious (e.g., simple accessors for obvious properties of the class). 
> * Private methods and functions declared in .cc files are not exempt.
> 
> **Function Definitions**
> * If there is anything tricky about how a function does its job, the function definition should have an explanatory comment. 
> * For example, in the definition comment you might describe any coding tricks you use, give an overview of the steps you go through, or explain why you chose to implement the function in the way you did rather than using a viable alternative. 
> * For instance, you might mention why it must acquire a lock for the first half of the function but why it is not needed for the second half.
> * Note you should not just repeat the comments given with the function declaration, in the .h file or wherever. It's okay to recapitulate briefly what the function does, but the focus of the comments should be on how it does it.
>

src/apps/bridge/reportBuilder.h
LarkspurRaven: From @matt001k's comment below:

> Could you make this PascalCase? This paradigm has been adopted from the bm_core work for most of bm
_protocol's current work. See: https://github.com/bristlemouth/bm_core?tab=readme-ov-file#code-style

snake_case is our preference for function names. Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge

LarkspurRaven: I updated any new additions from this PR to use that style. Didn't touch things not introudced in this PR. 

I'm not going to touch reportBuilderInit and reportBuilderAddToQueue in this PR b/c they were not in scope. I just happened to notice. However, I made a note in the follow-up ticket https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge.

src/apps/bridge/reportBuilderList.cpp
LarkspurRaven: Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge for follow-up.

src/apps/bridge/sensorController.cpp
LarkspurRaven: Updated. Though I want to revisit the code style in `bm_core`, `bm_protocol` and in `fleet-spotter-fw`. 

Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
LarkspurRaven: used it. Surprisingly, despite my many changes in reportBuilder, I pretty much duplicated the aanderaaConductivtySensor from the turbidity sensor. There's a lot more opportunity for cleaning these sensors up!

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
LarkspurRaven: Used it, thanks.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
LarkspurRaven: I meant that failText should probably applied to each `sample_member_params_t`, not just a general failure text that applies to all samples.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.h
LarkspurRaven: done

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
LarkspurRaven: fixed

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.h
LarkspurRaven: Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge

src/apps/bridge/reportBuilder.cpp
LarkspurRaven: Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge

src/apps/bridge/reportBuilder.cpp
LarkspurRaven: lost it with getting rid of vector :(

src/apps/bridge/sensorController.cpp
LarkspurRaven: updated

src/apps/bridge/topology_sampler.cpp
LarkspurRaven: Added note to https://app.shortcut.com/sofar/story/243503/conductivity-sensor-follow-up-bridge

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+    // Initialize aggregation structure with NaN values (invalid data indicator)
+    AanderaaConductivityAggregations conductivity_aggs = {};
+
+    // Only calculate means if we have sufficient data points
+    if (samplers[SamplerType::Conductivity_ms_cm].getNumSamples() > MIN_READINGS_FOR_AGGREGATION) {
+
+      // Update conductivity_aggs from AverageSamplers
+      for (const auto& f : kSamplerMap) {
+        (conductivity_aggs.*(f.aggr_ptr)) = samplers[f.type].getMean();
+      }
```
matt001k: Pretty, I like the `kSamplerMap`!

src/apps/bridge/reportBuilder.cpp
```diff
@@ -64,9 +67,24 @@ static QueueHandle_t _report_builder_queue = NULL;
...
+bool report_builder_add_sample_member(ReportParams &params, SampleMemberParams &s);
+
+/// Finalize the current sample entry in the CBOR report
+bool report_builder_close_sample(ReportParams &params);
+
+/// Helper function that opens a sample, adds all sample members from
+/// the report parameters, and closes the sample.
+/// This is the main function sensors should use to add their data to reports.
+bool report_builder_add_samples(ReportParams &params);
+
```
matt001k: Make static, only used locally.

src/apps/bridge/reportBuilder.cpp
```diff
@@ -133,6 +152,108 @@ void reportBuilderAddToQueue(uint64_t node_id, uint8_t sensor_type, void *sensor
   xQueueSend(_report_builder_queue, &item, portMAX_DELAY);
 }
 
+// =============================================================================================================================
+// ----------------------------------  report_builder helper functions  --------------------------------------------------------
+// =============================================================================================================================
+
```
matt001k: Remove for now, only used in this file for just the 3 newly added API, a bit of visual clutter.
But the idea is good.

src/apps/bridge/sensorController.cpp
```diff
@@ -32,14 +33,49 @@ typedef struct sensorControllerCtx {
   uint64_t _node_list[TOPOLOGY_SAMPLER_MAX_NODE_LIST_SIZE];
   bool _initialized;
   BridgePowerController *_bridge_power_controller;
+  // TODO(bjh): Too tightly coupled to the sensor types. We can delegate a
+  // function getReadingPeriodMs() to AbstractSensor and then call that function.
```
matt001k: Remove comment.

src/apps/bridge/sensorController.cpp
```diff
@@ -49,6 +85,47 @@ static void runController(void *param);
 static bool node_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
                                const char *service, size_t reply_len, uint8_t *reply_data);
 static void abstractSensorAddSensorSub(AbstractSensor *sensor);
+
+/**
+ * @brief Load a sensor's configuration from the config store.
+ * @param config_def The definition of the sensor's configuration.
+ * @param save Reference to a boolean that indicates if a save is needed.
+ */
+static void load_sensor_config(SensorConfigDef &config_def, bool &save) {
```
matt001k: Load is a bit confusing, lets use `get` in this case.
Function parameter `save` can be updated in the API comment to indicate that this value is an output.

src/apps/bridge/sensorController.cpp
```diff
@@ -49,6 +85,47 @@ static void runController(void *param);
...
+ * @brief Helper function to create and configure a sensor subscription
+ * @param subscription_config The sensor subscription configuration
+ * @param reply The decoded system info reply
+ * @param sample_duration_ms Sample duration in milliseconds
+ * @return true if sensor was processed (subscribed or already exists), false otherwise
+ */
+static bool create_and_configure_sensor_subscription(const SensorSubscriptionConfig& subscription_config,
+    const SysInfoReplyData& reply, uint32_t sample_duration_ms) {
+  if (strncmp(reply.app_name, subscription_config.app_name,
+              MIN(reply.app_name_strlen, strlen(subscription_config.app_name))) == 0) {
```
matt001k: Concerns around app_name if the replies app_name is a sub-string of the subscription_config's app_name. Can leverage `bm_wildcard_match`.

src/apps/bridge/sensorController.cpp
```diff
@@ -258,7 +358,23 @@ static bool node_info_reply_cb(bool ack, uint32_t msg_id, size_t service_strlen,
...
+      // Define Aanderaa conductivity subscription config
+      static const SensorSubscriptionConfig aanderaa_conductivity_config = {
+        .app_name = "aanderaa_conductivity",
+        .sensor_type = SENSOR_TYPE_AANDERAA_CONDUCTIVITY,
+        .reading_period_ms = &_ctx.aanderaa_conductivity_reading_period_ms,
+        .samples_pad = AanderaaConductivity_t::N_SAMPLES_PAD,
+        .create_fn = [](uint64_t node_id, uint32_t duration, uint32_t samples) -> AbstractSensor* {
+          return createAanderaaConductivitySub(node_id, duration, samples);
+        }
+        // .create_fn = &createAanderaaConductivitySub//(node_id, duration, samples)
```
matt001k: Does this have to be a lambda function? It seems like it could just be assigned directly?

src/apps/bridge/sensorController.cpp
matt001k, line 1: Let's remove the unnecessary comments in the struct definitions. These add clutter and a bit of confusion.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.h
```diff
@@ -0,0 +1,291 @@
...
+    Depth_m,
+    Count
+  };
+
+  
+  // Container for all samplers, indexed by SamplerType.
+  // Provides operator[] overloads so we can write samplers[SamplerType::X]
+  // instead of manually casting enum values to array indices.
+  struct SamplerSet {
+      std::array<AveragingSampler, static_cast<size_t>(SamplerType::Count)> data{};
```
matt001k: Concerns around what `std::array` might be doing under the hood (in terms of memory allocation). Potentially find an alternative using a c-style array.

```
AveragingSampler data[SamplerType::Count];
```

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  reading_count = 0;
+  last_timestamp = 0;
+  _mutex = xSemaphoreCreateMutex();
+  configASSERT(_mutex);
+
+    // Create mutex for thread-safe access
+  _mutex = xSemaphoreCreateMutex();
+  configASSERT(_mutex);
```
matt001k: Double creation by accident.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+        static_cast<double>(m.depth_m));
+
+    if (err != BmOK) {
+        bm_debug("ERROR: Failed to send Aanderaa Conductivity individual log to spotter, err: %d\n", err);
+    }
+
+    return err;
+}
+
+BmErr AanderaaConductivitySensor::send_spotter_log_aggregate(const AanderaaConductivityAggregations& agg) {
```
matt001k: Maybe not in its own function?

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+/// @details This function is called when the report builder is setting up the sensor data pointers
+void AanderaaConductivitySensor::setup_sensor_pointers(report_builder_element_t *element,
+                                                              const void **nan_sample,
+                                                              void **dst) {
+  *nan_sample = &AanderaaConductivitySensor::aanderaa_conductivity_NAN_AGG;
+  *dst = &(static_cast<AanderaaConductivityAggregations *>(
+      element->sensor_data))[element->sample_counter];
+}
+
+BmErr AanderaaConductivitySensor::send_spotter_log_individual(const AanderaaConductivityMsg::Data& m) {
```
matt001k: Maybe not in its own function?

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+    rval = bm_sub_wl(sub, topic_strlen, sub_callback) == BmOK;
+  }
+  bm_free(sub);
+  return rval;
+}
+
+/// @details This function is called when the report builder is aggregating sensor data
+void AanderaaConductivitySensor::aggregate(void) {
+  char *log_buf = static_cast<char *>(bm_malloc(SENSOR_LOG_BUF_SIZE));
+  configASSERT(log_buf);
```
matt001k: `log_buf` is not used here anymore.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+    reading_count = 0;
+    xSemaphoreGive(_mutex);
+  } else {
+    bm_debug("Failed to get the Aanderaa Conductivity mutex while trying to aggregate\n");
+  }
+  bm_free(log_buf);
+}
+
+/// @details This function is called when the report builder is setting up the sample member parameters
+void AanderaaConductivitySensor::get_sample_member_params(void *sensor_data, uint32_t sample_index, ReportParams &report_params) {
```
matt001k: Rename for clarity, `get_sample_member_params` confusing, ensure that it relates to `reportBuilder` if we continue to go down this path.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+    .sample_index = sample_index,
+    // TODO: fail text should probably be on sample member
+    .fail_text = "Failed to open aanderaa_conductivity sample in addSamplesToReport\n",
+    .sample_type = "bm_aanderaa_conductivity_v0",
+    .num_sample_members = AANDERAA_CONDUCTIVITY_NUM_SAMPLE_MEMBERS,
+    .sample_members = sample_members
+  };
+
+  // Fill the sample members array using the existing function
+  get_sample_member_params(sensor_data, sample_index, params);
```
matt001k: Maybe place function contents here, a lot of inner calls.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+      .sample_member_encoder_cb = encode_double_sample_member, .size = 0};
+
+    configASSERT_EXTRA(sample_member_count == report_params.num_sample_members,
+        "ERROR: Incorrect number of sample members for Aanderaa Conductivity\n");
+}
+
+/// @details This function is called when the report builder is setting up the report parameters
+ReportParams AanderaaConductivitySensor::get_report_params(sensor_report_encoder_context_t &context, void *sensor_data, uint32_t sample_index) {
+
+  // TODO(bjh): This is problematic if we have more than one sensor of this type!!
```
matt001k: This is not problematic in regards to how the `reportBuilder` is packaging CBOR data.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,281 @@
...
+          // Each value is cast to `double` before being passed into the sampler, so all samplers
+          // consistently operate on `double` input
+          const double v = std::visit([&](auto ptr) {
+            return static_cast<double>(composite_cbor_msg.*ptr);  // depth_m (float) becomes double here
+          }, f.cbor_ptr);
+
+          // Skip invalid values -- can may be unset, uninitialized, or explicitly encoded as “NaN” (not-a-number) to mean “no reading”.
+          if (!std::isnan(v)) {
+            conductivity_sensor->samplers[f.type].addSample(v);
+          }
```
matt001k, line 247: Use util function

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -0,0 +1,362 @@
...
+ */
+BmErr AanderaaConductivitySensor::send_spotter_log_aggregate(
+    const AanderaaConductivityAggregations &agg) {
+
+  BmErr err = AbstractSensor::send_spotter_log_aggregate(
+      "aanderaa_conductivity", agg.reading_count,
+      "%.4f,"   // conductivity_mean_ms_cm
+      "%.3f,"   // temperature_mean_deg_c
+      "%.3f,"   // salinity_mean_psu
+      "%.3f,"   // salinity_std_dev
```
victorsowa12: We bitpack/send 4 decimal places so lets also log that many

victorsowa12 (approved):
Looks good to me!

PR #342, opened 2025-09-18
Bug/rx live tags

evanShap:
## What changed?
This is a bugfix to the Rx-LIVE demo app that we fixed and patched back in Jan, and I just realized it was never merged in.
They _just_ re-deployed the system, without this code that makes the system work with more tag versions. 
🤦 🤦 🤦 

I don't know what the changes to the CmakePresets.json - @victorsowa12 did you make those?

## Checklist


victorsowa12 (approved):
LGTM. I believe Zac added the cmake-presets. Those allow for an easier cmake set up flow.

PR #352, opened 2025-10-23
Feature/remote bm commands while bus is off

matt001k:
## What changed?
Adds the ability to reply to power status requests from `bm_serial`!
Refactors the `powerInfoStatsCb` API in `bridgePowerController.cpp` be more readable as it shares a lot of logic with the newly added API `getPowerStats`!

NCP devices connected to the Bridge can now be informed of the Bristlemouth networks power! (Whether it is on or off!)

Largest changes are in `bridgePowerController.cpp`.

## Checklist


src/apps/bridge/bridgePowerController.cpp
```diff
@@ -441,53 +441,190 @@ void BridgePowerController::powerControllerRun(void *arg) {
...
+
+ @return Power status data
+ */
+bm_serial_power_status_reply_data_t BridgePowerController::getPowerStats(void) {
+  bm_serial_power_status_reply_data_t d = {0, 0};
+
+  PowerTimings timings = calculatePowerTimings();
+
+  if (timings.remaining_on_s == POWER_SERVICE_UNDEFINED) {
+    d.remaining_on_ms = S_IN_A_DAY;
```
victorsowa12: Why not set the `d.remaining_on_ms` to `POWER_SERVICE_UNDEFINED`? Seems very unlikely, but I think seconds in a day could be a valid option.

victorsowa12 (approved):
Looks good to me! Just had a question about using `S_IN_A_DAY` v.s. `POWER_SERVICE_UNDEFINED` (aka uint32_max).

src/apps/bridge/bridgePowerController.cpp
matt001k: I believe this can change. Before the finalized design, there was something about this that made sense at the time (that I cannot remember). But, now whenever the bridge boots, a power stats request is sent to the Bridge from the connected device, making this redundant I will update and test!

matt001k: For visibility, this can be changed as was done so in the code before merge.

PR #354, opened 2025-10-29
Improve Command Sending To Aanderaa 5990 Sensor

matt001k:
To avoid opening a large PR I am going to be breaking up the work on the Aanderaa Salinity New Calibration parameters into a few separate PRs.
The other work depends on this new `sendCommand` API which creates a robust way of sending data when configuring the device and reading values from the device.

## What changed?
Addition of the `sendCommand` API in the conductivity application.

This improves the handling of commands sent to the Aanderaa 5990 sensor from the mote application.
Each command waits until an acknowledge is received by the 5990 before moving on, improving the speed at which the device can be configured. Get commands are able to assign values generic values based on their outputs through the checkTypeAndAssign overloaded functions.

This improves the robustness of sending commands and reading values. But I am curious if there should be a retry mechanism as well?
Thoughts welcome.

## Checklist

### DIFF
--- a/src/apps/rs232_expander_apps/aanderaa_salinity/user_code/aanderaa_conductivity_sensor.cpp
+++ b/src/apps/rs232_expander_apps/aanderaa_salinity/user_code/aanderaa_conductivity_sensor.cpp
@@ -41,7 +41,7 @@ void AanderaaConductivitySensor::init() {
   // Baud set to 9600, which is expected by the Aanderaa conductivity sensor
   PLUART::setBaud(BAUD_RATE);
   // Disable passing raw bytes to user app.
-  PLUART::setUseByteStreamBuffer(false);
+  PLUART::setUseByteStreamBuffer(true);
   PLUART::setUseLineBuffer(true);
   // Set a line termination character per protocol of the sensor.
   PLUART::setTerminationCharacter(LINE_TERM);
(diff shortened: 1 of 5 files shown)


towynlin (approved):

src/apps/rs232_expander_apps/aanderaa_salinity/user_code/aanderaa_conductivity_sensor.cpp
```diff
@@ -41,7 +41,7 @@ void AanderaaConductivitySensor::init() {
   // Baud set to 9600, which is expected by the Aanderaa conductivity sensor
   PLUART::setBaud(BAUD_RATE);
   // Disable passing raw bytes to user app.
-  PLUART::setUseByteStreamBuffer(false);
+  PLUART::setUseByteStreamBuffer(true);
   PLUART::setUseLineBuffer(true);
   // Set a line termination character per protocol of the sensor.
   PLUART::setTerminationCharacter(LINE_TERM);
```
towynlin, line 47: One comment — if we're not using readLine anymore, should we turn off the internal line buffering? Anecdotally, I've found issues when trying to use both lines and bytes in the past.

matt001k: I was curious about this is well and did some extensive testing around making sure that there is no funky business afoot with both the line buffer and directly accessing the stream buffer.
Because there are only two instances during run time where the stream buffer is accessed:

1. During configuration at initialization
2. When calibration is being performed

There is not a chance the line buffer and stream buffer will encounter strange results (the line buffer is used when polling the device for newly read data).

The mote app first:

1. Stops the sensor from outputting readings
2. Configures the sensor
3. Starts the sensor's reading output
4. Listens for new incoming data and waits to see if calibration should be performed.

To note, this all happens sequentially in the same task and eliminates the chance of multiple threads attempting to access the PLUART peripheral.

One thing that makes the communication a bit more robust between the 5990 and the mote is the use of the `Start` and `Stop` command.
When sending `Stop` to the 5990, it will not output sensor readings.
Making `sendCommand` very isolated.
The `sendCommand` API will also handle if there are non-expected readings being reported from the 5990, which is actually very common when the sensor first boots or when waking up the sensor to perform calibration!
And whenever `sendCommand` is called, the PLUART is flushed (which flushes both the stream buffer and line buffer outputs)!

victorsowa12 (approved):
LGTM

uma-kat (approved):

PR #356, opened 2025-10-31
Add Legacy Bring Up Apps To Release

matt001k:
## What changed?
Adds legacy bringup apps to release.

Useful for the Sofar EE team on some of their production test fixtures!

Rubber stamp

## Checklist


victorsowa12 (approved):

tools/scripts/release/auto_build.py
```diff
@@ -104,6 +104,9 @@ def setup_cmake_dir(self, config, ext_dir):
         if config["name"].startswith("aanderaa"):
             cmd += ["-DCMAKE_AANDERAA_TYPE=4830"]
 
+        if config["legacy"]:
```
mbella-sofar: I don't know too much about these release scripts but it looks to me like this will raise an exception if the config file being parsed doesn't have any sections with the `legacy` key.

I noticed line 346-349 in this script makes sure the `sign` and `encrypt` keys are in config. Is something similar needed for `legacy` to prevent issues in the future?

matt001k: That is a real valid point, this will raise a `KeyError`. I opted to use the `get` API to assign a default value of None if this key does not exist in the config.

tools/scripts/release/auto_build.py
```diff
@@ -104,6 +104,9 @@ def setup_cmake_dir(self, config, ext_dir):
         if config["name"].startswith("aanderaa"):
             cmd += ["-DCMAKE_AANDERAA_TYPE=4830"]
 
+        if config.get("legacy", None) is not None:
```
mbella-sofar: I think this evaluation will end up adding the LEGACY flag even if the yml section contains `legacy: false`

I _think_ this will do what we want `if config.get("legacy", None) and config.get("legacy", None) is not None:`

mbella-sofar (approved):

PR #357, opened 2025-11-04
Fix/aanderaa salinity low power activation time

matt001k:
## What changed?
Change the timeout for the salinity sensor to go into low power mode from 60 seconds, to 10 seconds.
Also changes the bridge configuration name for salinity reading period from `conductivityReadingPeriodMs` to `salinityReadingPeriodMs`.

When the Aanderaa 5990 is not in low power mode, it consumes an excess of ~100mW, meaning this reduces the amount of time the device will be in a high powered state!
Below is a graph showing the Bristlemouth network power draw of a system with a single Salinity sensor attached with this change implemented:
image

The following is a graph showing the Bristlemouth network power draw of a system with a single salinty sensor attached prior to this change:

image

Note the X-Axis time. The large spikes in power draw are when the salinity sensor is collecting its readings.

Questions or further validations necessary?
Not a large amount of change, but quite important!

## Checklist


victorsowa12 (approved):
🎉

PR #358, opened 2025-11-05
Add In Read Validate Write Logic To aanderaa_conductivity_sensor.cpp

matt001k:
## What changed?
Aanderaa 5990 configuration values are only updated if they do not match what is expected using the new API `readValidateWriteValue`.

The sensor that the salinity app communicates to, the Aanderaa 5990, currently is configured and saved to the sensor on every boot of the mote application. When saving this configuration to the 5990, the internal flash of the sensor is written to in order to store these values permanently. This is an action that does not have to be performed every time that the sensor boots, but only when actual values change (such as the first time the device is configured, when new values are added in a firmware update for the mote or when the Bristlemouth configuration values `sensorDepthM` or `readingPeriodS` are changed).
This update reduces the amount of times the flash is written to and saves precious erase/write cycles.
Also, the save action takes ~8 seconds to complete, so when a save is not required, it allows the sensor to begin collecting data sooner.

The following console outputs represent when a parameter is not configured properly on the 5990:

`Enable Rawdata`
```
command: Get Enable Rawdata
command response: Enable Rawdata        5990    75      No
command response: #
sendCommand err: 0
command: Set Enable Rawdata(Yes)
command response: #
sendCommand err: 0
```

Sensor's sample interval (`Interval`)
```
command: Get Interval
command response: Interval      5990    75      60.000
command response: #
sendCommand err: 0
command: Set Interval(10)
command response: #
sendCommand err: 0
```

System pressure (`Pressure`)
```
command: Get Pressure
command response: Pressure      5990    75      0.000000E+00
command response: #
sendCommand err: 0
command: Set Pressure(155.000000)
command response: #
sendCommand err: 0
````

The configuration is then saved if a value has to change on the 5990 (through the `save` command). This is based on the `_sensorConfigDirty` variable.

## Checklist

### DIFF
--- a/src/apps/rs232_expander_apps/aanderaa_salinity/user_code/aanderaa_conductivity_sensor.cpp
+++ b/src/apps/rs232_expander_apps/aanderaa_salinity/user_code/aanderaa_conductivity_sensor.cpp
@@ -621,6 +621,20 @@ BmErr AanderaaConductivitySensor::sendCommand(const char *command, T *value,
   return err;
 }
 
+/*!
+ @brief Compares UINT Values And Populates Set Buffer To Send Values
+
+ @details If the values for read and expected do not match, the buffer used to
+          set the parameter is populated.
+
+ @param parameter parameter to set if values do not match
+ @param buf buffer to populate to send the set command
+ @param read read value from 5990
+ @param expected expected value
+
+ @return true if the read value matches the expected value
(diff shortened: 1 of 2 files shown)


victorsowa12 (commented):
LGTM!

victorsowa12 (approved):

Issue #16, opened 2023-10-21
DevKit Payload UART wrapper (PLUART) should implement a queue for sensor lines, rather than a single buffer


evanShap:
### Issue / request
- PLUART maintains a single static `_user_line` buffer for thread-safe access of serial line data from sensors.
- If the client app does not service data in this buffer with a call to `readLine` before another line comes from the sensor, the new line overwrites the buffer and data is lost.
- Implement a thread-safe queue/buffer that can store multiple lines.
- Alert client if data if queue overflowed and has been overwritten/lost.

### Background / why
The PLUART namespace (defined [here](https://github.com/bristlemouth/bm_protocol/blob/3c10ebdedcb73a3b012b9545f5171ab0ba3a9ca8/src/lib/common/payload_uart.cpp)) adds some very convenient wrappers to the DevKit's Payload UART interface, which is the data interface for many of the serial connections the DevKit supports (UART, RS232, RS485, RS422, SDI-12).

There are (2) thread-safe methods exposed for users to access Rx data from their apps - `PLUART::readLine` and `PLUART::readByte`.

`readLine` can be used when the attached device is generating frames of serial data with reserved delimiting characters (a common example is ASCII text data with \<CR> and/or \<LF> being the frame delimiter). This is a _very_ common pattern for legacy marine sensors so this is a _very_ convenient wrapper to have.

**The issue** is that most of these sensors will have some multi-line data or command responses, the the potential for overwriting with the existing implementation results hairy-edge brittleness and a poor developer experience - the line buffer works most of the time. Works often enough to make it non-obvious what exactly is going wrong, but fails more than often enough to make it something a developer can just ignore. It's also susceptible to small timing changes in the app causing things to break, which can result in pernicious Heisenbugs.

related PR from legacy private repo for reference: https://github.com/wavespotter/bristlemouth/pull/465

SimenKH:
As of the FreeRTOSConfig.h files I have read, it does seem like the protocol uses the CMSIS API for FreeRTOS.

So it could feasibly be solved by redefining the buffer in https://github.com/bristlemouth/bm_protocol/blob/3c10ebdedcb73a3b012b9545f5171ab0ba3a9ca8/src/lib/common/payload_uart.cpp to be a CMSIS message queue. 

https://www.keil.com/pack/doc/CMSIS/RTOS/html/group__CMSIS__RTOS__Message.html
[...]

PR #361, opened 2025-11-07
Convert PLUART to queue for sensor lines (#16)

keenanjohnson:
## What changed?

Implements queue-based line buffering for PLUART to prevent data loss when multiple lines arrive before being read.

Issue #16 raised by the devs seems to indicate a need and makes sense to me! As mentioned in the issue, many sensors issue multiple lines and this change seems to only make the project more robust to timing jitter without changing the user API.

src/lib/common/payload_uart.cpp contains the important bits. I have added a set of basic unit tests as well.

## Checklist

### DIFF
--- a/src/lib/common/payload_uart.cpp
+++ b/src/lib/common/payload_uart.cpp
@@ -179,14 +179,9 @@ static void processLine(void *serialHandle, uint8_t *line, size_t len) {
   memcpy(queued_line.buffer, line, len);
   queued_line.len = len;
 
-  // Try to enqueue the line without blocking (we're in an ISR context)
-  BaseType_t xHigherPriorityTaskWoken = pdFALSE;
-  if (xQueueSendFromISR(_line_queue, &queued_line, &xHigherPriorityTaskWoken) != pdTRUE) {
-    // Queue is full - increment dropped line counter
-    // Use assignment to avoid C++20 deprecated volatile increment
+  if (xQueueSend(_line_queue, &queued_line, 0) != pdTRUE) {
     _dropped_lines = _dropped_lines + 1;
   }
-  portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
 }
 
 char getTerminationCharacter() { return terminationCharacter; }
--- a/src/lib/common/payload_uart.cpp
+++ b/src/lib/common/payload_uart.cpp
@@ -6,6 +6,7 @@
 
 // FreeRTOS Includes
 #include "FreeRTOS.h"
+#include "queue.h"
 #include "semphr.h"
 
 namespace PLUART {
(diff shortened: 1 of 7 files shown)


keenanjohnson:
Just ran across this project in open source and seems like a great effort! I've done a few ocean robotics projects in the past and always looking for ways to give back. I saw that this issue was open and seemed straightforward, but let me know if it's not helpful or I am headed in the wrong direction.

towynlin:
Thanks @keenanjohnson! Please work on fixing the CI failure, and we'll help with some hardware validation.

keenanjohnson:
Great! The latest push fixes the tests I think

towynlin:
Hey @keenanjohnson CI is still failing — you can see the ❌ next to the latest commit and the "All checks have failed" notice at the bottom fo the PR. Here's a screenshot of the build failure:
Screenshot 2025-11-18 at 4 31 35 PM

matt001k (commented):
Nice, I like this change a lot!
I will give it a shot on actual hardware here soon (give me at most a week)!

src/lib/common/payload_uart.cpp
```diff
@@ -168,14 +175,18 @@ static void processLine(void *serialHandle, uint8_t *line, size_t len) {
...
 
-  configASSERT(xSemaphoreTake(_user_line.mutex, portMAX_DELAY) == pdTRUE);
-  // if the last line has not been read, lets reset the buffer
-  memset(_user_line.buffer, 0, LPUART1_LINE_BUFF_LEN);
-  // lets copy over the line to the user mutex protected buffer
-  memcpy(_user_line.buffer, line, len);
-  _user_line.len = len;
-  _user_line.ready = true;
-  xSemaphoreGive(_user_line.mutex);
+  QueuedLine_t queued_line;
```
matt001k, line 178: Ah this may be a problem here. Every time a line ends we create a `queue_line` item, this struct has a buffer of 2048 bytes in it. Meaning, 2048+ bytes are allocated to the `serialGenericRxTask. Looking at that task the stack is currently 2048 words (32bit) wide which should be enough for this operation.

Note: I tested this on an application and it seems to be running ok.

matt001k: Also to note, this is not occurring from an ISR. We do not need the `xQueueSendFromISR` function here.
The `xQueueSend` function will also copy the above struct (`queue_line`) and heap allocate it. Meaning every time a line is created >4096 bytes of allocation happen (until `queued_line` goes out of scope).

matt001k: I would recommend creating some memory pool that the queue items could point to.
There is a module I created exactly for something like this called [q.c](https://github.com/bristlemouth/bm_core/blob/main/common/q.c) located in the `bm_core` submodule of this project (the submodule might have to be updated). This could be a good alternative as:

- There is no max queue size, it is all dependent of how much memory in the pool has been used
- Elements in the queue can be of different sizes
- It is much faster in terms of operation than the FreeRTOS queue

The only downfall is:

- It is not thread safe and will need mutex protection

Let me know your thoughts on that solution!

src/lib/common/payload_uart.cpp
```diff
@@ -168,14 +175,18 @@ static void processLine(void *serialHandle, uint8_t *line, size_t len) {
...
+  memcpy(queued_line.buffer, line, len);
+  queued_line.len = len;
+
+  // Try to enqueue the line without blocking (we're in an ISR context)
+  BaseType_t xHigherPriorityTaskWoken = pdFALSE;
+  if (xQueueSendFromISR(_line_queue, &queued_line, &xHigherPriorityTaskWoken) != pdTRUE) {
+    // Queue is full - increment dropped line counter
+    // Use assignment to avoid C++20 deprecated volatile increment
```
matt001k: Wow did not know about this one... that is a crazy deprecation

src/lib/common/payload_uart.cpp
towynlin: @keenanjohnson thanks again for the contribution! We should get this one merged! I really appreciate the new unit tests. 🙏 

This comment thread from @matt001k is last follow-up item, since you fixed the build.
- We should change from `xQueueSendFromISR` to `xQueueSend`.
- What do you think about using the queue module in `q.c`? The main benefit for this use case is we could more efficiently use memory, allocating only what we need instead of the max for every queue item, and moving some large memory usage from the stack to the heap.

keenanjohnson: Just changed the  xQueueSendFromISR to xQueueSend.

Looking at the queue module now.

PR #363, opened 2025-11-11
Feature/2025 lc stdev

NicholasIanMerritt:
## What changed?
Restructured and mostly rewrote the loadCellSample.cpp app. 

-Added running variance calculation and standard deviation reporting to remote messages. 
-Added missed/successful/attempted readings counting. N for stats is now successful readings, not reading attempts. 
-Created new format for remote LC message that fits with ME team needs. 

Also some smaller things: Fixed some out of order checks vs measurements, changed some misleading names, deleted unused defines and variables. 

Code has been tested on a loadcell and the remote-reported standard deviation matches one calculated after the fact with SD data. 

-new features (stdev, reading counts)
-much more intelligible

## Checklist

- [ ] Add or update unit tests for changed code -- No relevant unit tests I think. 
- [X] Ensure all submodules up to date. If this PR relies on changes in submodules, merge those PRs first, then point this PR at/after the merge commit
- [ ?] Ensure code is formatted correctly with clang-format. If there are large formatting changes, they should happen in a separate whitespace-only commit on this PR after all approvals.

### DIFF
--- a/src/lib/sensor_sampler/loadCellSampler.cpp
+++ b/src/lib/sensor_sampler/loadCellSampler.cpp
@@ -1,14 +1,3 @@
-/*!
-  Load cell sampling functions
-TO DO :
- Check the running var logic
-
-
- QUESTIONS:
- - USE_TIMESTAMP ? 
- - INA_STR_LEN
-*/
-
 #include "loadCellSampler.h"
 #include "spotter.h"
 #include "pubsub.h"
--- a/src/lib/sensor_sampler/loadCellSampler.cpp
+++ b/src/lib/sensor_sampler/loadCellSampler.cpp
@@ -143,25 +143,25 @@ static bool loadCellSample() {
       stdev      = 0.0f;
     }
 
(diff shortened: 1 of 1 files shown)


src/lib/sensor_sampler/loadCellSampler.cpp
```diff
@@ -59,65 +65,110 @@ static bool loadCellSample() {
...
+              uptimeGetMicroSeconds() / 1000, rtcTimeBuffer, reading);
+    spotter_log_console(0, "loadcell | tick: %llu, rtc: %s, weight: %f\n",
+              uptimeGetMicroSeconds() / 1000, rtcTimeBuffer, weight);
+
+  } 
+  
+  //If load cell read fails 
+  else {
+    missed_reading_counter ++; // to track at the end
+    reading_attempts_counter++; //This still updates so we send our message on time. We do not interate successful_reading_count.
```
victorsowa12, line 118: we could increment this outside of the if/else since it is incremented in both cases.

victorsowa12 (approved):
Looks good to me! Had a small optimization suggestion

PR #368, opened 2026-01-16
Fixes Missing RBR Readings

matt001k:
## What changed?
Ensures that probing the RBR cannot happen at the same time that an expected reading is incoming.
This is due to the fact that the RBR cannot transmit and receive over UART at the same time!
This is called "Data Blanking" in RBR's terminology.

About 1/1,000 RBR streaming data readings were getting dropped.
This improves the data integrity by preventing the [blanking](https://docs.rbr-global.com/l3-command-reference/N/External/timeouts-output-blanking-and-power-saving#id-(N)Timeouts,outputblankingandpowersaving-Outputblanking) timing condition from occurring.

Small PR. Ask any questions you might have!

## Checklist


towynlin (approved):

victorsowa12 (approved):
👏

PR #371, opened 2026-02-26
Example Bristlemouth Ardupilot MAVLink Application

matt001k:
## What changed?

Adds example MAVLink bridge application compatible with Ardupilot flight controllers.
The following [documentation](https://ardupilot.org/dev/docs/learning-ardupilot-uarts-and-the-console.html) describes the UART protocol, by default serial1 and serial2 provide a MAVLink 2 stream.
This code ingests a few MAVlink messages of interest from that stream and forwards it onto the Bristlemouth network.

First example application of a MAVLink integration!
Adds the ability to easily add Bristlemouth communications to an Ardupilot flight controller.

Take a look at the release config, should this app be signed and a release build?

## Checklist


towynlin (approved):
Looks great!

Yes, I think a signed release from CI sounds correct. If people want to install this on their dev kits, that will still work fine.

One small request — in the presets, would you please sort them alphabetically? Right now in the output of `cmake --list-presets` they are, except 2 have been recently added to the bottom: mavlink-bridge and aanderaa-salinity-rs232.

victorsowa12 (approved):
👏

PR #373, opened 2026-03-09
Increase Buffer Size For Bridge Config Topology And Chunk Over UART

matt001k:
## What changed?

Pretty prints cbor configuration map to NULL buffer on bridge to get the size needed for the NCP log buffer, then proceeds to pretty print to the allocated buffer.
The log statement is broken up into chunks so that it may be sent over the UART interface properly.

Allows for larger buffer size on Bridge when sending configuration maps to the NCP over UART to support an increased number of nodes on the Bristlemouth network.

I have looked for alternative solutions that will not potentially use a large amount of RAM but was not able to find anything great in tinycbor's library. Let me know if you can think of some better solutions!

## Testing
Here is the config map output from a Sofar Spotter:
image

```
2025-03-09T18:02:22.113Z [BRIDGE_CFG] [INFO] Bridge network config: [[18051847170415161419, "bridge", 2794679863, 3944824111, {"sampleIntervalMs": 1800000, "sampleDurationMs": 840000, "subsampleIntervalMs": 60000, "subsampleDurationMs": 30000, "subsample
Enabled": 0, "bridgePowerControllerEnabled": 1, "alignmentInterval5Min": 1, "samplesPerReport": 2, "transmitAggregations": 1, "currentReadingPeriodMs": 60000, "softReadingPeriodMs": 500, "rbrCodaReadingPeriodMs": 500, "disableUnusedPortsTimeMs": 0, "tick
sSamplingEnabled": 0, "turbidityReadingPeriodMs": 1000, "salinityReadingPeriodMs": 30000}], [11316520146590703508, "bm_rbr", 3114525382, 1280529602, {"9d0c5b1da7171f94": 0, "3ab84acda350c349": 0, "45694087da99b923": 0, "dfu_confirm": 1, "rbrCodaType": 2,
 "disableUnusedPortsTimeMs": 0}], [15307738215748708816, "serial_bridge", 838057655, 1030244400, {"disableUnusedPortsTimeMs": 0}], [17338441281132032165, "borealis", 3114525382, 245190499, {"disableUnusedPortsTimeMs": 0, "enable_sbc": 0, "sbc_command": "
/usr/local/bin/borealis_hydrotwin.sh", "report_interval": 810.f, "hydrotwin_ldr": 12, "sample_rate": 48000, "enable_daq_recording": 1}], [7371431991783118678, "aanderaa", 3114525382, 1941523237, {"sensorsPollIntervalMs": 5000, "sensorsCheckIntervalS": 60
, "currentAggPeriodMin": 0.f, "nSkipReadings": 0, "readingIntervalMs": 60000, "payloadWdToS": 70, "plUartBaudRate": 9600, "dfu_confirm": 1, "sensorBmLogEnable": 1, "disableUnusedPortsTimeMs": 0}], [4187221732029418504, "serial_bridge", 1876941642, 103024
4400, {"disableUnusedPortsTimeMs": 0}], [3028553066222167635, "bm_rbr", 3114525382, 2842535829, {"dfu_confirm": 1, "rbrCodaType": 2, "sensorsPollIntervalMs": 10000, "sensorBmLogEnable": 1, "disableUnusedPortsTimeMs": 0}], [17099771185193130725, "serial_b
ridge", 3634098711, 1030244400, {"disableUnusedPortsTimeMs": 0}], [16084703961411914896, "aanderaa", 3114525382, 4096890594, {"sensorsPollIntervalMs": 10000, "sensorsCheckIntervalS": 60, "currentAggPeriodMin": 0.f, "nSkipReadings": 0, "readingIntervalMs"
: 60000, "payloadWdToS": 70, "plUartBaudRate": 9600, "dfu_confirm": 1, "sensorBmLogEnable": 1, "disableUnusedPortsTimeMs": 0}], [16089316954901563807, "serial_bridge", 1876941642, 772967734, {"disableUnusedPortsTimeMs": 15000}]
```

## Checklist

### DIFF
--- a/src/apps/bridge/bridgeLog.cpp
+++ b/src/apps/bridge/bridgeLog.cpp
@@ -99,10 +99,9 @@ static void check_lock_publish_or_enqueue(const BridgeLogPubInfo *info) {
   if (!locked || is_locking_task) {
     bm_serial_pub(getNodeId(), info->topic, info->topic_length, info->data, info->data_length,
                   info->type, info->version);
-    return;
+  } else {
+    bridge_log_enqueue(info);
   }
-
-  bridge_log_enqueue(info);
 
   return;
 }
--- a/src/apps/bridge/bridgeLog.cpp
+++ b/src/apps/bridge/bridgeLog.cpp
@@ -1,11 +1,151 @@
 #include "bridgeLog.h"
 #include "FreeRTOS.h"
 #include "app_pub_sub.h"
+#include "bm_os.h"
 #include "bm_serial.h"
 #include "device_info.h"
+#include "q.h"
 #include "stm32_rtc.h"
+#include "task.h"
 #include <stdio.h>
 #include <string.h>
+
+#define SHARE_MESSAGE_BUF_SIZE 2048
+
+// Message sizes rarely should come close to the shared buffer size, this should adequate
+#define QUEUE_BUFFER_SIZE (2 * SHARE_MESSAGE_BUF_SIZE + 2 * sizeof(QItem))
+
+typedef struct {
+  uint8_t type;
+  uint8_t version;
+  uint16_t topic_length;
+  uint16_t data_length;
+  const char *topic;
+  const uint8_t *data;
+} BridgeLogPubInfo;
+
+typedef struct {
+  bool lock;              // If locked, enqueues items to send upon unlocking
+  BmTaskHandle lock_task; // Task performing the locking operation
+  BmSemaphore mut;
+  Q message_queue;
+  uint8_t message_queue_buf[QUEUE_BUFFER_SIZE];
+  uint8_t shared_message_buf[SHARE_MESSAGE_BUF_SIZE];
+} BridgeLogCtx;
+
+static BridgeLogCtx ctx = {};
+
+static BmErr bridge_log_enqueue(const BridgeLogPubInfo *info) {
+  uint32_t topic_offset = sizeof(BridgeLogPubInfo);
+  uint32_t data_offset = topic_offset + info->topic_length;
+  uint32_t total_size = data_offset + info->data_length;
+
+  if (total_size > SHARE_MESSAGE_BUF_SIZE) {
+    return BmENOMEM;
+  }
+
+  uint8_t *buf = ctx.shared_message_buf;
+
+  // Copy over the static parts of the struct
+  memcpy(buf, info, sizeof(BridgeLogPubInfo));
+
+  // Copy over the topic and data
+  memcpy(&buf[topic_offset], info->topic, info->topic_length);
+  memcpy(&buf[data_offset], info->data, info->data_length);
+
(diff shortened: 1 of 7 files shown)


towynlin:
I haven't looked at the code yet, but a quickthought — from the description:

> log statement is broken up into chunks

Does that reintroduce the past buggy behavior of other log lines getting interpolated in the middle of this printed configuration? We need to prevent that. If we have to break it up for memory reasons, then maybe we need a mutex on the serial port to keep lines together?

matt001k:
@towynlin was just going to tag @evanShap in on this to ask. Yes that is a potential problem that can happen.
The chunking is necessary because the serial buffers on the Bridge and Spotter are 2048 bytes, they would have to be increased to support more nodes on the network. Bridge can take the memory hit, but spotter cannot.
I think adding a locking mechanism to `bridgeLog.cpp` is a good move to prevent this behavior.

evanShap:
> @towynlin was just going to tag @evanShap in on this to ask. Yes that is a potential problem that can happen. The chunking is necessary because the serial buffers on the Bridge and Spotter are 2048 bytes, they would have to be increased to support more nodes on the network. Bridge can take the memory hit, but spotter cannot. I think adding a locking mechanism to `bridgeLog.cpp` is a good move to prevent this behavior.

Not sure if you need any more input from me - but I can affirm that interleaved log lines are a big annoyance for QTEAs and Spotter users. We've made a lot of progress in robustifying/reducing corruption in long-running logging performance, want to avoid regressions there.

matt001k:
Ok cool thanks! I will think on this for a bit, determine how the bridge can not drop packets while keeping this message clean and prevent any latency when handling these specific log statements on the bridge.

matt001k:
@towynlin I added a locking mechanism to `bridgeLog` to prevent interleaved log statements. To keep latency down in the system when `bridgeLog` is locked, these log statements are added to a queue. Upon unlocking `bridgeLog`, these log statements are dequeued.
This allows the Bridge to continue chugging along without dropping packets to the NCP (Spotter).
Also, we have about ~100k static RAM available, so all buffers are static here!

matt001k:
I want to continue doing some testing on the 10 node mega-system before merging.

src/apps/bridge/bridgeLog.cpp
```diff
@@ -1,12 +1,152 @@
...
+  bool locked = ctx.lock;
+  bool is_locking_task = xTaskGetCurrentTaskHandle() == ctx.lock_task;
+
+  if (!locked || is_locking_task) {
+    bm_serial_pub(getNodeId(), info->topic, info->topic_length, info->data, info->data_length,
+                  info->type, info->version);
+    return;
+  }
+
+  bridge_log_enqueue(info);
```
towynlin: This confused me a bit because my eye skipped over the return. I think an else clause would be clearer than the early return.

towynlin (approved):
Looks good to me other than one small clarity improvement. 👏

victorsowa12 (approved):
👏

PR #375, opened 2026-03-25
Initialize Bridge Log Earlier

matt001k:
There are bridge log portions of the code that can be reached before `bridgeLogInit` was called causing the bridge to crash

## What changed?
Initialize bridge log very early before anything else can call `bridgeLog` functions.

The HIL tests remove configuration values and `sensorControllerInit` will `bridgeLogPrint` to Spotter if those values are missing. This was causing the bridge to crash. This fixes that by initializing the bridge log module very early on.
The `bridgeLogInit` function creates a mutex and queue, the mutex could not be taken because it was NULL.
Here is the crash backtrace:
image

Should I also add protections in the bridgeLog module to only allow the mutex to be taken only if it has been created?
I feel fairly confident with this change, but extra defensive code would help just in-case `bridgeLogInit` were to change initialization location in the future.

## Checklist

### DIFF
--- a/src/apps/bridge/app_main.cpp
+++ b/src/apps/bridge/app_main.cpp
@@ -343,6 +343,8 @@ static void defaultTask(void *parameters) {
   // Inhibit low power mode during boot process
   lpmPeripheralActive(LPM_BOOT);
 
+  bridgeLogInit();
+
   startSerial();
   startSerialConsole(&usbCLI);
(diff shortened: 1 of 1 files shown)


victorsowa12 (approved):
I'm also pretty confident your change fixes things. I think adding the defensive code could be beneficial, but not critical. I think go for it if you have the bandwidth!

PR #376, opened 2026-03-26
feat: exo3 sonde data viewer

towynlin:
## What changed?

I added a quick AI-generated html file for visualizing exo3 sonde data in charts.

The original poster of this topic on the Bristlemouth forum was looking for alternatives to the existing python file.
https://bristlemouth.discourse.group/t/decoding-data-exo3-that-is-sent-from-the-dev-kit-to-the-sofar-spotter-dashboard/561

Screenshot 2026-03-26 at 2 39 30 PM

Screenshot 2026-03-26 at 2 39 36 PM

Help make EXO3 sonde data more accessible.

Is this the right place for tools like this? Often I'd suggest no, but for this dev kit integration specifically, the python script and other content are in the same directory as the app code.

## Checklist

### DIFF
--- a/src/apps/bm_devkit/sdi12_example/exo3-sonde-data-viewer.html
+++ b/src/apps/bm_devkit/sdi12_example/exo3-sonde-data-viewer.html
@@ -1,0 +1,461 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


victorsowa12 (approved):
The plots look really nice! HTML is pretty foreign to me so can't provide much feedback there. As far as location, I think this is a good spot for now! It helps keep it all in one location to reference.

matt001k (approved):

PR #378, opened 2026-03-29
Feat/uart sbc

towynlin:
## What changed?

Added `native_serial_bridge` dev kit app, which treats the PLUART as a 3rd Bristlemouth port. If you run a Raspberry Pi on the other end of that UART with an app built from the `bm_sbc` repo, then this new dev kit app bridges the Bristlemouth traffice between the hardware motes with ADINs and processes running on the pi.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,7 @@ else()
     # Setup testing
     include(CTest)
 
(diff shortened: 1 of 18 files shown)


campbellaos:
Note that Borealis will require this functionality on usart3 rather than pluart.

PR #381, opened 2026-04-14
Allow For Increased Baudrate On LPUART

matt001k:
## What changed?
If the desired baud rate is above 115200, set the LPUART clock prescaler to 1 before setting the baud rate register.
Disables and potentially re-enabled the peripheral before setting the prescaler and baud rate registers.

This updates the LPUART peripheral clock prescaler, which is necessary for increasing the baudrate above 666kBd.

The math for calculating the baud is (ref: 67.4.8 in [rm0456](https://www.st.com/resource/en/reference_manual/rm0456-stm32u5-series-armbased-32bit-mcus-stmicroelectronics.pdf)):

$$
\text{baud} = \frac{256f_{clk}}{BRR}
$$

Where $BRR$ is the baud rate register and $f_{clk}$ is the prescaled peripheral clock.
Before this change the clock prescaler is defaulted to $8$.
The clock used for this peripheral is the HSI (High Speed Internal) oscillator, which operates at 16MHz.
With a prescaler of 1, the baud rate can max out at 5.333MBd.

In order to write to the prescale and baud rate register, the UE bit (responsible for enabling the peripheral) has to be set to 0. This is explained in sections 67.6.7 of rm0456 (note auto baudrate detection mode is not supported by LPUART and is a typo):

image

And section 67.6.13:

image

Ask questions if the flow does not make sense! A few key items to note:

1. When the peripheral is disabled, the interrupts should also be disabled, a fault will occur if the peripheral is disabled and an interrupt occurs after disabling the peripheral (this was examined when attempting to set the baud rate during runtime after two devices have started communicating to one another).
2. Prior to re-enabling the peripheral, the data is flushed to avoid handling malformed data.

## Checklist

### DIFF
--- a/src/lib/common/payload_uart.cpp
+++ b/src/lib/common/payload_uart.cpp
@@ -328,16 +328,13 @@ void configTxPinAlternate(void) {
 void enable(void) {
   USART_TypeDef *dev = static_cast<USART_TypeDef *>(uart_handle.device);
 
-  // Clear data in read register
-  (void)LL_LPUART_ReceiveData8(dev);
-
   LL_LPUART_Enable(dev);
+
+  serialEnable(&uart_handle);
 
   // Clear any pending NVIC bit from the disabled period, then re-enable
   NVIC_ClearPendingIRQ(LPUART1_IRQn);
   NVIC_EnableIRQ(LPUART1_IRQn);
-
-  serialEnable(&uart_handle);
 }
 
 void disable(void) {
--- a/src/lib/common/payload_uart.cpp
+++ b/src/lib/common/payload_uart.cpp
@@ -1,4 +1,5 @@
 #include "payload_uart.h"
+#include "bm_usart.h"
 #include "spotter.h"
 #include "stm32_io.h"
 #include "stm32_rtc.h"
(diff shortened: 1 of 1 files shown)


src/lib/common/payload_uart.cpp
```diff
@@ -307,9 +325,31 @@ void configTxPinAlternate(void) {
   }
 }
 
-void enable(void) { serialEnable(&uart_handle); }
+void enable(void) {
+  USART_TypeDef *dev = static_cast<USART_TypeDef *>(uart_handle.device);
+
+  // Clear data in read register
+  (void)LL_LPUART_ReceiveData8(dev);
```
victorsowa12: I believe a recieveData8 gets called in serialEnable, so might be able to remove?

victorsowa12 (approved):
Looks good to me! Just had a comment on a potential duplicated receiveData8

src/lib/common/payload_uart.cpp
matt001k: Thanks! Nice catch, I updated it and tested with a runtime baud rate change and it still looks good!

PR #384, opened 2026-04-28
Add Break Callback To Serial Module

matt001k:
## What changed?
Adds serial break character handler.
Sets `EIE` (Error Interrupt Enable) bit to be able to handle framing errors to recognize break events.
Statement within section 66.5.7/67.4.7 of the STM32U5x Reference Manual explains the usage of the framing error:
image

Allows serial peripherals to recognize break characters and handle them accordingly, allowing the mote to synchronize events based on this new handler.

For Borealis V2 this is being used strictly in PLUART, should this be implemented only in the `payload_uart.cpp` file ? I do not see an immediate reason for needing this feature in any other serial peripherals at the moment, but this does provide flexibility for future initiatives (syncing NCP UART device with the bridge for example).

## Checklist

### DIFF
--- a/src/lib/common/serial.c
+++ b/src/lib/common/serial.c
@@ -303,35 +303,6 @@ void serialGenericUartIRQHandler(SerialHandle_t *handle) {
 
   configASSERT(handle != NULL);
 
-  // If error drain the Receive Data Register, data is invalid
-  uint32_t isr = USART2->ISR;
-  if (isr & (USART_ISR_FE | USART_ISR_NE | USART_ISR_ORE)) {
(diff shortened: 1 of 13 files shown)


src/lib/common/serial.c
```diff
@@ -105,6 +105,24 @@ uint32_t serialGetBaudRate(SerialHandle_t *handle) {
   return ret;
 }
 
+// Handle break conditions
+typedef void (*HWBreakCb_t)(void);
```
victorsowa12: nit: isn't this already defined in serial.h?

victorsowa12 (approved):
Looks good to me! I don't see an issue adding this to all of them since it totally allows more flexibility for future use cases!

PR #386, opened 2026-05-12
LSM6DSV Driver

matt001k:
## What changed?
Adds driver for LSM6DSV.

Configures and enables LSM6DSV to operate using the FIFO functionality allowing the processor to sleep while samples are being collected.
Readings can then be collected all at once through the FIFO's buffer and triggered to do so when a configured reading count has occurred on the LSM6DSV (for the gyro and accelerometer).
Each reading is stored within a queue and can be read asynchronously in another task.

Allows for the sensor hub functionality to work by leveraging an abstract ST sensor class that can be used to access and initialize other devices on an I2C bus connected to the LSM6DSV. For the motion sensing board, this will allow for the LISM2MDL to be accessed through the LSM6DSV. Data for sensor hub devices are read from the FIFO and then assigned to these sensors each through their `set_data` API.

The following resources explain the functionality of the LSM6DSV:

- [Datasheet](https://www.st.com/resource/en/datasheet/lsm6dsv.pdf)
- [Application Node](https://www.st.com/resource/en/application_note/an5922-lsm6dsv-6axis-imu-with-embedded-sensor-fusion-i3c-oiseis-for-smart-applications-stmicroelectronics.pdf)

It might be worth reviewing the FIFO and sensor hub sections to get an idea of how this is configured and expected to work!

## Checklist

### DIFF
--- a/src/lib/drivers/lsm6dsv.cpp
+++ b/src/lib/drivers/lsm6dsv.cpp
@@ -384,6 +384,15 @@ BmErr LSM6DSV::stream_handle(void) {
   return BmOK;
 }
 
+/*!
+ @brief Determine if data is ready to be read from FIFO
+
+ @param timeout_ms
+
+ @return BmOK if data is available in FIFO to read
+         BmETIMEDOUT if data is not ready
+         BmENODEV if LSM6DSV was not initialized
+ */
 BmErr LSM6DSV::reading_ready(uint32_t timeout_ms) {
   if (!m_reading_sem) {
     return BmENODEV;
@@ -404,6 +413,7 @@ BmErr LSM6DSV::reading_ready(uint32_t timeout_ms) {
          BmENODATA if there are no elements in the queue to dequeue
          BmENOMEM if size of data is smaller than the size of the element being
                   dequeued
+         BmENODEV if LSM6DSV was not initialized
  */
 BmErr LSM6DSV::get_reading(LSM6DSVReading *reading) {
   if (!reading) {
--- a/src/lib/drivers/lsm6dsv.cpp
+++ b/src/lib/drivers/lsm6dsv.cpp
@@ -26,60 +26,6 @@
     status.bit ? BmOK : BmETIMEDOUT;                                                           \
   })
 
-void LSM6DSVSensor::set_driver_ctx(void) {
-  // Set the read and write functions for the driver
-  m_ctx.write_reg = driver_write;
-  m_ctx.read_reg = driver_read;
-  m_ctx.mdelay = driver_delay_ms;
-  m_ctx.handle = this;
-}
-
-LSM6DSVSensor::LSM6DSVSensor(uint8_t address) : m_address(address) { set_driver_ctx(); }
-
-LSM6DSVSensor::LSM6DSVSensor(SensorInterfaceBus *bus, uint8_t address)
-    : m_bus(bus), m_address(address) {
-  set_driver_ctx();
-}
-
-LSM6DSVSensor::LSM6DSVSensor(SensorInterfaceBus *bus) : m_bus(bus) { set_driver_ctx(); }
-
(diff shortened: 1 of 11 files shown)


src/lib/drivers/lsm6dsv.cpp
```diff
@@ -0,0 +1,494 @@
...
+          readings to buffer on the imu before offloading them to be handled
+          by the processor.
+
+ @return BmOK on success
+         BmENOMEM if there is insufficient memory to create mutexes/semaphore
+         BmENODEV if the imu is not found
+         BmEACCES if bus operations fail when attempting communication to imu
+ */
+BmErr LSM6DSV::init(void) {
+  BmErr err = q_create_static(&m_reading_queue, m_readings_buf, sizeof(m_readings_buf));
```
mbella-sofar, line 62: The app note says there is a 10ms period after power on where the LSM6 is reading cal data from its built in flash.

I suspect this function will be getting called after this at the very soonest but it might be good to add a delay to the start to avoid strange edge cases later since 10ms is not very long.

src/lib/drivers/lsm6dsv.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+  // Validate the LSM6DSV device ID
+  uint8_t whoami = 0;
+  lsm6dsv_return_on_err(lsm6dsv_device_id_get(&m_ctx, &whoami));
+  if (whoami != LSM6DSV_ID) {
+    return BmENODEV;
+  }
+
+  // Reset the LSM6DSV
+  lsm6dsv_return_on_err(lsm6dsv_sw_por(&m_ctx));
```
mbella-sofar, line 79: The coefficient loading process I mentioned above appears to only happen at power on (at least I can't find any detail on any other situations like reset where it should also happen).

We might need to tweak things here during bringup if we hit issues.

src/lib/drivers/lsm6dsv.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+  // Ensures 16 bit registers can't be updated until high and low registers are read
+  lsm6dsv_return_on_err(lsm6dsv_block_data_update_set(&m_ctx, PROPERTY_ENABLE));
+
+  // Calculate sampling timestamp resolution t = 1 / (46080 * (1 + 0.0013 * FREQ_FINE), ref: 6.4 AN5922
+  int8_t freq_fine;
+  lsm6dsv_return_on_err(lsm6dsv_odr_cal_reg_get(&m_ctx, &freq_fine));
+  uint64_t denominator = 46080000 + 59904 * static_cast<uint64_t>(freq_fine);
+  uint64_t numerator = 1000000000ULL * 1000;
+  m_timestamp.resolution_ns = numerator / denominator;
```
mbella-sofar, line 89: I think I understand how you got here from the equation in the comment above. You scaled everything up so fixed point math works and then you distributed the 4608 across the paren enclosed equation?

mbella-sofar (approved):
Looks good to me. Just have a couple minor things I noticed noted in the comments.

src/lib/drivers/lsm6dsv.cpp
matt001k: There is a 20ms delay built into the `lsm6dsv_sw_por` function call. This function does a software reset of the IMU!

src/lib/drivers/lsm6dsv.cpp
matt001k: Ahh ok!

src/lib/drivers/lsm6dsv.cpp
matt001k: Correct:

```math
\begin{align*}
46080 * 0.0013 &= 59.904 \\
59.904 * 1000 &= 59904
\end{align*}
```

Then multiplied the numerator by 1000 as well.

PR #388, opened 2026-05-26
Allow RBR To Recognize Device When Bootstraping And Hotswaping

matt001k:
## What changed?
Check the last time data has been received from the RBR agnostic of the value being properly parsed.
Always probe for attached RBR if `_type` is `UNKNOWN`.

This allows the RBR application to set the RBR if one has not been set yet and reinstates hot swapping RBR modules as this bug also broke that functionality.

This happened due to the fact that `_last_data_time` would never get set because `handleDataString` would always return `false` (the function could not parse the string because the application did not know the `_type` of RBR attached).
`_last_data_time` should be set no matter if `handleDataString` returns `true` or `false` as this variable is meant to be used to determine when **valid** data was last received, not if the data was **parsed** properly. This allows the application to properly probe the RBR sensor.

## Testing

The following setup was used for testing:

- RS232 RBR module on v0.13.11-rc.3
- 1 RBR coda³ T.D
- 1 RBR coda³ T

Initially I deleted the `rbrCodaType` config in the system configs:

Screenshot 2026-05-26 at 10 14 13 AM

I would repeatedly get `attempting FTL recovery` print statements and **the `rbrCodaType` would never get set**.

Screenshot 2026-05-26 at 10 14 02 AM

I then proceeded to flash the fix in this PR.
Immediately the `rbrCodaType` was set:

Screenshot 2026-05-26 at 10 15 58 AM

I then hot swapped from the RBR coda³ T.D to the RBR coda³ T and was able to see the config get set and changed upon the next probing window:

Screenshot 2026-05-26 at 10 34 11 AM

Screenshot 2026-05-26 at 10 34 38 AM

I then hot-swapped back to the RBR coda³ T.D and it was properly recognized and the configuration was set upon the next probing window:

Screenshot 2026-05-26 at 10 35 18 AM

I added an extra safety rail when the `_type` is not set (`UNKNOWN`).
Do you believe this is needed?
It could mess with the probing window, but also data should not be getting transmitted up Bristlemouth if `_type` is not set.

## Checklist

### DIFF
--- a/src/apps/bristleback_apps/bm_rbr/user_code/rbr_sensor.cpp
+++ b/src/apps/bristleback_apps/bm_rbr/user_code/rbr_sensor.cpp
@@ -72,7 +72,7 @@ void RbrSensor::maybeProbeType(uint64_t last_power_on_time) {
       RBR_READING_PERIOD_MS / 2 + PROBE_WINDOW_WIDTH_MS / 2;
   bool lower_window = now > _last_data_time + window_start_time;
   bool upper_window = now < _last_data_time + window_end_time;
-  bool can_transmit = lower_window && upper_window;
+  bool can_transmit = (lower_window && upper_window) || _type == BmRbrDataMsg::UNKNOWN;
 
   if (now - _lastProbeTime >= _minProbePeriodMs &&
       now - last_power_on_time >= _minProbePeriodMs && can_transmit) {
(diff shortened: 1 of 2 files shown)


victorsowa12 (approved):
Nice find! LGTM!

PR #391, opened 2026-06-23
Adds RX Fifo For Serial Interface

matt001k:
## What changed?
Adds RX FIFO functionality and `SerialHandle_t` configuration to choose between using the Fifo or general purpose RX interrupt.

Reduces the number of processor interrupt/wake up events by leveraging the UART FIFO on the STM32U5. This lessens the chance for data to get dropped while the processor is in STOP1 mode. Note it takes about 7-9 microseconds for the processor to wake up from STOP1 mode (we are using the [MSI at 24MHz as our wake up clock](https://github.com/bristlemouth/bm_protocol/blob/1595f804c5012d57619a2645d6661d288a5e96c5/src/lib/common/lpm_u5.c#L41-L48), see the [following](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://www.st.com/resource/en/datasheet/stm32u575ag.pdf&ved=2ahUKEwj3mNii26CVAxWqLzQIHVZeEc8QFnoECBkQAQ&usg=AOvVaw0IN4xj-_DzwFjecdhdTf7W)):
image
The FIFO can hold 8 elements, it is configured to fire an interrupt after it is half full, essentially reducing the number of times the interrupt fires by a factor of 4. It is configured as half full in case there is more data ingested on the RX line while in the interrupt, ensuring bytes do not get dropped.

## Testing

I ran a BorealisV2 unit with 1.5MBd communication between the SCARI DAQ Microcontroller and the Bristlemouth Mote:

```
2026-06-23T22:29:41.953Z | borealis mote firmware: ab539f930e-dirty
2026-06-23T22:29:49.644Z | gpzda_to_sys: adjusted time by 272419657 microseconds
2026-06-23T22:29:49.648Z | $GPZDA,222949.601659,23,06,2026,00,00*6C
2026-06-23T22:29:49.660Z | gpzda_to_sys: adjusted time by 13512 microseconds
2026-06-23T22:29:49.664Z | $GPZDA,222949.631252,23,06,2026,00,00*60
2026-06-23T22:29:49.667Z | defer_work: skipping
2026-06-23T22:29:49.675Z | scari: changing baud rate to 1500000
2026-06-23T22:29:49.726Z | $PCFG,31250.00,rec,usb,gram,spl,replay,baud1500000,gain0*6B
```

The SCARI was set up to transmit SPL and PGRAM messages to the Mote and through Bristlemouth.

Within the SD card contents, there were no indications of dropped packets due to checksum errors as previously added to the Borealis2 code by @victorsowa12.
The FIFO logic was able to keep up with the higher data rate!

When examining the dropped messages ON A SOFAR OCEAN SPOTTER in the `XXXX_SENS_IND.csv` file, there was a total of 21, which is about 0.0157% of the messages not reaching the Spotter's SD card. Note these messages could have been dropped in:

- The Bristlemouth network
- The UART communication between Sofar Ocean's Spotter and the Bridge
- NOT in the communication between SCARI and the mote as there would have been an indication in the `borealis_checksum.log` file which never appears in the `bm` directory.

Here is the output of an analysis of this file:

```
======================================================================
 /Volumes/NO NAME 4/log/0000_SENS_IND.csv
======================================================================

  node_app_name: borealis [type 1] (66821 readings, tick 9624-65712725 ms, median interval: 983.0 ms)
    Gaps (>1.10x median = >1081.3 ms): 11 event(s) (0.02%)
      tick 3222566.000 -> 3228465.000 (6 gaps)
      tick 10822334.000 -> 10827250.000 (5 gaps)

  node_app_name: borealis [type 2] (66823 readings, tick 9618-65713699 ms, median interval: 983.0 ms)
    Gaps (>1.10x median = >1081.3 ms): 10 event(s) (0.01%)
      tick 3222560.000 -> 3228459.000 (6 gaps)
      tick 10823312.000 -> 10827244.000 (4 gaps)
```

Attached below are some pertinent logs from this test(the full zipped directory from the test is larger than Github's allowable upload size of 30MB):

[0000_borealis.log.zip](https://github.com/user-attachments/files/29308850/0000_borealis.log.zip)
[0000_port_monitor.log.zip](https://github.com/user-attachments/files/29308852/0000_port_monitor.log.zip)
[0000_power.log.zip](https://github.com/user-attachments/files/29308853/0000_power.log.zip)
[0000_SENS_AGG.csv.zip](https://github.com/user-attachments/files/29308865/0000_SENS_AGG.csv.zip)
[0000_SENS_IND.csv.zip](https://github.com/user-attachments/files/29308866/0000_SENS_IND.csv.zip)

## Checklist

[0000_SENS_AGG.csv](https://github.com/user-attachments/files/29308856/0000_SENS_AGG.csv)

### DIFF
--- a/src/lib/common/payload_uart.cpp
+++ b/src/lib/common/payload_uart.cpp
@@ -328,9 +328,9 @@ void configTxPinAlternate(void) {
 void enable(void) {
   USART_TypeDef *dev = static_cast<USART_TypeDef *>(uart_handle.device);
 
-  LL_LPUART_Enable(dev);
-
   serialEnable(&uart_handle);
+
+  LL_LPUART_Enable(dev);
 
   // Clear any pending NVIC bit from the disabled period, then re-enable
   NVIC_ClearPendingIRQ(LPUART1_IRQn);
@@ -378,7 +378,8 @@ SerialHandle_t uart_handle = {
     .preTxCb = NULL,
     .postTxCb = NULL,
     .breakISR = NULL,
-};
+    .fifoEnabled = true,
+}; // namespace PLUART
 
 BaseType_t init(uint8_t task_priority) {
   // Create the stream buffer for the user bytes to be buffered into and read from
--- a/src/lib/common/serial.c
+++ b/src/lib/common/serial.c
@@ -261,8 +261,16 @@ void serialEnable(SerialHandle_t *handle) {
 
     if (handle->rxStreamBuffer) {
       xStreamBufferReset(handle->rxStreamBuffer);
-      // Enable Uart RX interrupt
-      usart_EnableIT_RXNE((USART_TypeDef *)handle->device);
+      if (handle->fifoEnabled) {
+        LL_USART_EnableIT_IDLE(handle->device);
+        LL_USART_EnableIT_RXFT(handle->device);
+        LL_USART_EnableIT_RXFF(handle->device);
+        LL_USART_SetRXFIFOThreshold(handle->device, LL_USART_FIFOTHRESHOLD_1_2);
+        LL_USART_EnableFIFO(handle->device);
+      } else {
+        // Enable Uart RX interrupt
+        usart_EnableIT_RXNE((USART_TypeDef *)handle->device);
+      }
     }
   }
 #endif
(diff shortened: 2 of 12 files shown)


src/lib/common/payload_uart.cpp
```diff
@@ -328,10 +328,10 @@ void configTxPinAlternate(void) {
 void enable(void) {
   USART_TypeDef *dev = static_cast<USART_TypeDef *>(uart_handle.device);
 
-  LL_LPUART_Enable(dev);
-
   serialEnable(&uart_handle);
 
+  LL_LPUART_Enable(dev);
```
matt001k, line 333: Modify registers in `serialEnable` before actually enabling the peripheral.

matt001k:
@victorsowa12 this will need some testing on Borealis2 to validate at higher baud rates there are no dropped packets such as we have seen with normal interrupt RX.

matt001k:
@victorsowa12 testing results have been added!

src/lib/common/serial.c
```diff
@@ -303,13 +311,25 @@ void serialGenericUartIRQHandler(SerialHandle_t *handle) {
...
+                usart_IsEnabledIT_RXNE((USART_TypeDef *)handle->device);
   // Process received bytes
-  if( usart_IsActiveFlag_RXNE((USART_TypeDef *)handle->device) &&
-      usart_IsEnabledIT_RXNE((USART_TypeDef *)handle->device)){
+  if (idle_int || fifo_int || rx_int) {
+    while (LL_USART_IsActiveFlag_RXNE_RXFNE(handle->device)) {
+      uint8_t byte = usart_ReceiveData8((USART_TypeDef *)handle->device);
+      configASSERT(handle->rxBytesFromISR);
+      higherPriorityTaskWoken = handle->rxBytesFromISR(handle, &byte, 1);
+    }
```
victorsowa12, line 328: does this while loop impact non-fifo implementations at all? I assume we read it once and it clears the RXNE interrupt, we read the single byte, and then don't loop again. Is that correct?

victorsowa12 (approved):
Looks good! Just had a clarifying question about the while loop in the interrupt.

src/lib/common/serial.c
matt001k: It does not, there will only be one RXFNE event while in this handler in the non-fifo implementation.

PR #393, opened 2026-06-24
Implement chunked network info transfer for large config maps

Davidbick:
## What changed?
The Bridge now sends the network config map in 1024-byte chunks using BM_SERIAL_NETWORK_INFO_CHUNK before sending the final bm_serial_send_network_info as a commit. Also updates bm_topology_last_network_info_cb to use the same chunking logic.

The config map sent from Bridge to Spotter was silently dropped once it exceeded 2048 bytes. As networks grow with more nodes and config keys this becomes a real failure. This fix removes that ceiling while remaining backwards compatible, old Spotter firmware ignores the chunk messages and still receives the final packet.

The chunking logic in check_topology_report and bm_topology_last_network_info_cb in topology_sampler.cpp. Submodule pointers for bm_serial and bm_core point to this branch — merge those PRs first.

``
Tests have been conducted using configs with cbor encoded sizes of less than 1024, 1024 exactly, between 1024 and 2048, and above 2048 bytes. All have been successfully passed to the backend.
``

### DIFF
--- a/src/apps/bridge/topology_sampler.cpp
+++ b/src/apps/bridge/topology_sampler.cpp
@@ -131,7 +131,7 @@ static void topology_sample_cb(NetworkTopology *topology) {
 // Returns the new pointer location of where the cbor_map now exists
 static uint8_t *send_network_info_chunks(uint8_t *cbor_map, size_t cbor_map_size,
                                          uint32_t &chunk_data_to_send_bytes) {
-  static constexpr uint16_t NETWORK_INFO_CHUNK_SIZE = 1024;
+  static constexpr uint16_t NETWORK_INFO_CHUNK_SIZE = 1800;
   if (cbor_map_size <= NETWORK_INFO_CHUNK_SIZE) {
     return cbor_map;
   }
--- a/src/apps/bridge/topology_sampler.cpp
+++ b/src/apps/bridge/topology_sampler.cpp
@@ -85,6 +85,9 @@ static void _update_sensor_type_list(uint64_t node_id, char *app_name, uint32_t 
 
 static void log_network_crc_info(uint32_t network_crc32, SMConfigCRCList &sm_config_crc_list);
 
+static uint8_t *send_network_info_chunks(uint8_t *cbor_map, size_t cbor_map_size,
+                                         uint32_t &chunk_data_to_send_bytes);
+
 static void topology_sample_cb(NetworkTopology *topology) {
   if (!topology) {
     bridgeLogPrint(BRIDGE_CFG, BM_COMMON_LOG_LEVEL_INFO, USE_HEADER,
@@ -123,6 +126,26 @@ static void topology_sample_cb(NetworkTopology *topology) {
   }
   xSemaphoreGive(_node_list.node_list_mutex);
   bm_semaphore_give(network_topology_semaphore);
+}
+
+// Returns the new pointer location of where the cbor_map now exists
+static uint8_t *send_network_info_chunks(uint8_t *cbor_map, size_t cbor_map_size,
+                                         uint32_t &chunk_data_to_send_bytes) {
+  static constexpr uint16_t NETWORK_INFO_CHUNK_SIZE = 1024;
+  if (cbor_map_size <= NETWORK_INFO_CHUNK_SIZE) {
+    return cbor_map;
+  }
+  uint32_t offset = 0;
+  chunk_data_to_send_bytes = cbor_map_size - NETWORK_INFO_CHUNK_SIZE;
+  while (offset < chunk_data_to_send_bytes) {
+    uint32_t remaining = chunk_data_to_send_bytes - offset;
+    uint16_t chunk_size = remaining > NETWORK_INFO_CHUNK_SIZE
+                              ? NETWORK_INFO_CHUNK_SIZE
+                              : static_cast<uint16_t>(remaining);
+    bm_serial_send_network_info_chunk(cbor_map_size, offset, chunk_size, &cbor_map[offset]);
+    offset += chunk_size;
+  }
+  return &cbor_map[offset];
 }
 
 static void check_topology_report(uint32_t timeout_ms) {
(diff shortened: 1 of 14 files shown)


src/apps/bridge/topology_sampler.cpp
```diff
@@ -207,12 +207,34 @@ static void check_topology_report(uint32_t timeout_ms) {
...
+                                      ? NETWORK_INFO_CHUNK_SIZE
+                                      : static_cast<uint16_t>(remaining);
+            bm_serial_send_network_info_chunk(cbor_bufsize, offset, chunk_size,
+                                              &cbor_buffer[offset]);
+            offset += chunk_size;
+          }
+          bm_serial_send_network_info(network_crc32_calc, &config_crc, &fw_info,
+                                      _node_list.num_nodes, _node_list.nodes,
+                                      cbor_bufsize - chunk_data_to_send_bytes,
+                                      &cbor_buffer[chunk_data_to_send_bytes]);
```
matt001k: This section of code and lines 823-839 seem to be copy paste of one another. In order to improve the maintainability of this, I would recommend turning it into a separate function such as:

```c
// Returns the new pointer location of where the cbor_map now exists
// chunk_data_to_send_bytes is now passed by reference to know the amount of data left over
uint8_t *send_network_chunks(const uint8_t *cbor_map, size_t cbor_map_size, uint32_t &chunk_data_to_send_bytes) {
  static constexpr uint16_t NETWORK_INFO_CHUNK_SIZE = 1024;
   if (cbor_map_size <= NETWORK_INFO_CHUNK_SIZE) {
    return cbor_map
   }

    uint32_t offset = 0;
    chunk_data_to_send_bytes = cbor_map_size - NETWORK_INFO_CHUNK_SIZE;
    while (offset < chunk_data_to_send_bytes) {
      uint32_t remaining = chunk_data_to_send_bytes - offset;
      uint16_t chunk_size = remaining > NETWORK_INFO_CHUNK_SIZE
                                ? NETWORK_INFO_CHUNK_SIZE
                                : static_cast<uint16_t>(remaining);
      bm_serial_send_network_info_chunk(cbor_map_size, offset, chunk_size,
                                        &cbor_map[offset]);
      offset += chunk_size;
    }

  return &cbor_map[offset]
}
```

This will encapsulate the logic into a single location.
I would also recommend taking out the `else` statement and having the code be invoked as the following:

```c
uint32_t chunk_data_to_send_bytes = 0;
uint8_t *cbor_buffer_updated = send_network_chunks(cbor_buffer, cbor_bufsize, chunk_data_to_send_bytes);
bm_serial_send_network_info(network_crc32_calc, &config_crc, &fw_info,
                            _node_list.num_nodes, _node_list.nodes, cbor_bufsize - chunk_data_to_send_bytes,
                            cbor_buffer_updated);
```

Notice how this reduces the amount of code necessary by:

1. Breaking it into a function
2. Having only 1 call site for `bm_serial_send_network_info` and reducing the number of branching (`if` statement) conditions. Branching conditions lead to complexity and fragility in code! Read more about [Cyclomatic Complexity](https://www.codeant.ai/blogs/reduce-code-complexity-guide) here.

And this exact reduced amount code can then be used in both locations!

Happy to talk through this more and the best practices for reducing code complexity.

src/apps/bridge/topology_sampler.cpp
```diff
@@ -775,6 +797,9 @@ uint8_t *topology_sampler_alloc_last_network_config(uint32_t &network_crc32,
 void bm_topology_last_network_info_cb(void) {
   if (xSemaphoreTake(_node_list.node_list_mutex, pdMS_TO_TICKS(NETWORK_CONFIG_TIMEOUT_MS))) {
     do {
+      if (!_node_list.last_network_configuration_info.cbor_config_map) {
+        break;
+      }
```
matt001k: What was the reason for adding these lines?

src/apps/bridge/topology_sampler.cpp
Davidbick: That sounds good, I will make a commit with the new function and apply it to reduce the code complexity like you suggested.

src/apps/bridge/topology_sampler.cpp
Davidbick: On boot, the last configuration is not set so this pointer is null. I further investigated this and it seems like a bug for a different PR so I will remove these lines for now.

Although nothing crashes with or without these lines, the bigger issue is that smsync can be stuck in an "ongoing" state if:
1. It is run immediately after boot and before the first configuration was set
2. The cbor array is invalid. This is especially prevalent for outdated spotters than get a bridge update and don't know how to interpret chunked messages (above 1024 byes)

The reset the smsync ongoing flag, a reboot is needed. I spoke to Victor and I will look into and document this bug for a future PR.

matt001k: I vote to open another PR with this fix in it! That is a very very good catch! For reference, Sofar has seen systems in the field that have lost their sensor data and I assume it is because of your second bullet point!

src/apps/bridge/topology_sampler.cpp
matt001k: Looks real clean 😮‍💨 nice work!

matt001k (approved):

matt001k:
Please merge `bm_serial` before this PR and then update the reference to `bm_serial`'s latest `develop` branch.

src/apps/bridge/topology_sampler.cpp
```diff
@@ -125,6 +128,26 @@ static void topology_sample_cb(NetworkTopology *topology) {
   bm_semaphore_give(network_topology_semaphore);
 }
 
+// Returns the new pointer location of where the cbor_map now exists
+static uint8_t *send_network_info_chunks(uint8_t *cbor_map, size_t cbor_map_size,
+                                         uint32_t &chunk_data_to_send_bytes) {
+  static constexpr uint16_t NETWORK_INFO_CHUNK_SIZE = 1024;
```
victorsowa12: The previous "limit" was 2048. Could we make this limit 2048? This would allow for the most systems to continue to use the regular config map path, and then only larger smart mooring systems have to use the chunking. I'm not sure if this is true, but I can imaging that there are systems in the field right now that have config maps that fall between 1024 < config_map_size < 2048 bytes. Doing an OTA update of the bridge on this type of system would then mean we don't get the config map back after the update until we also update the Spotter, but getting that config map is important for validating successful updates.

victorsowa12 (approved):
Looks good! Just had a question about increasing the chunk size.

src/apps/bridge/topology_sampler.cpp
Davidbick: This is a good catch. The serial limit is 2048 bytes so to have the maximum chunk size while allowing space for headers, I changed the chunk size to be 1800 bytes in this commit: https://github.com/bristlemouth/bm_protocol/pull/393/changes/8f0c5d62d6d97eb3a0ae81b72a2d3358ea7ee8a7

PR #394, opened 2026-06-26
Grab The Node ID From Hydrotwin Topic Rather Than Embedded In The Message

matt001k:
## What changed?
Rather than using the node ID in a Bristlemouth publish message to determine which hydrotwin topic this publication belongs to, use the Node ID embedded in the topic of the publication.

Allows Hydrotwin HDR and LDR messages to be logged onto Sofar's Spotter.

I decided to use `strtoull`, but am open to other options (note I tried `sscanf` but the ARM compiler does not support grabbing 64bit integer values).

## Checklist


victorsowa12 (approved):
LGTM

PR #395, opened 2026-07-01
Fix/chunk num nodes

Davidbick:
## What changed?
The chunked config map message now carries the number of nodes so that the spotter can allocate memory accurately when dealing with the network headers

This helps the spotter code to remain clean and allocation to be precise when receiving chunked config maps

## Checklist


victorsowa12 (approved):
Looks good!

matt001k (approved):

PR #396, opened 2026-07-10
Fixes Invalid Instruction After Waking Up From Sleep

matt001k:
## What changed?
Add a `_NOP()` instruction after exiting sleep mode. 

Prevents a `UsageFault` that could occur when the mote wakes from sleep, where the CPU would attempt to execute an invalid instruction.

## Why is this happening?
There is no documented errata for this issue, but debugging showed the mote was hitting this exception on wake.
Evaluating the [CFSR](https://developer.arm.com/documentation/ddi0403/d/System-Level-Architecture/System-Address-Map/System-Control-Space--SCS-/Configurable-Fault-Status-Register--CFSR?lang=en) register showed a `UsageFault` set at bit 0, indicating the CPU attempted to execute an invalid instruction:

image
image

Indicating that an instruction was attempted to be executed that was invalid.

This was occurring upon the entry of the following [function](https://github.com/bristlemouth/bm_protocol/blob/0989713378142ad27e8f52ebfc66ca9d02ee558d/src/lib/common/lpm_u5.c#L154):

```c
     configPOST_SLEEP_PROCESSING( (const TickType_t)xExpectedIdleTime );
```

I validated this by loading code that reports the Program Counter (PC) and Link Register (LR) to a Sofar Ocean Spotter's SD card. On reset, the captured PC/LR values map to these lines in the built binary:

image

image

An [ST forum report](https://community.st.com/stm32-mcus-embedded-software-32/stm32u5a5-instruction-fault-after-wfi-158950) describes matching behavior on an STM32U5 variant. I'll post our solution there and push to get this into the next errata revision.

## Checklist

### DIFF
--- a/src/lib/common/lptimTick_u5.c
+++ b/src/lib/common/lptimTick_u5.c
@@ -453,6 +453,7 @@ void vPortSuppressTicksAndSleep( TickType_t xExpectedIdleTime )
             //
             __DSB();
             enter_sleep_mode();
+            __NOP();
             __ISB();
          }
          configPOST_SLEEP_PROCESSING( (const TickType_t)xExpectedIdleTime );


campbellaos (commented):
I wonder if the nop shouldn't be in the asm block immediately following wfi, within enter_sleep_mode(), instead of here.

campbellaos (approved):

victorsowa12 (approved):
lgtm!

PR #397, opened 2026-07-10
LIS2MDL Driver

matt001k:
## What changed?
Adds driver for LIS2MDL magnetometer.

Allows the magnetometer to data to be captured by the LSM6DSV's sensor hub feature.

To determine how all of this works you will want to look at the following files:

- `src/lib/drivers/lsm6dsv.cpp`
- `src/lib/drivers/abstract/abstract_st_sensor.cpp`

The following datasheet can also be viewed for further information on functionality.
[lis2mdl.pdf](https://github.com/user-attachments/files/29909213/lis2mdl.pdf)

## Checklist

### DIFF
--- a/src/lib/drivers/lis2mdl.cpp
+++ b/src/lib/drivers/lis2mdl.cpp
@@ -1,0 +1,67 @@
+#include "lis2mdl.h"
+#include "app_util.h"
+#include "bm_config.h"
+
+#define lis2mdl_return_on_err(f)                                                               \
+  ({                                                                                           \
+    int32_t err = (f);                                                                         \
+    if (err != 0) {                                                                            \
+      bm_debug("lis2mdl err %d:%" PRId32 "\n", __LINE__, err);                                 \
+      return BmEACCES;                                                                         \
+    }                                                                                          \
+  })
+
+BmErr LIS2MDL::init(void) {
+  BmErr err = q_create_static(&m_reading_queue, m_readings_buf, sizeof(m_readings_buf));
+  if (err != BmOK) {
+    return err;
+  }
+
+  m_queue_mut = bm_mutex_create();
+  if (!m_queue_mut) {
+    return BmENOMEM;
+  }
+
+  // Validate the LIS2MDL device ID
+  uint8_t whoami = 0;
+  lis2mdl_return_on_err(lis2mdl_device_id_get(&m_ctx, &whoami));
+  if (whoami != LIS2MDL_ID) {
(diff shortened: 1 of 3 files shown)


src/lib/drivers/lis2mdl.cpp
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  int16_t datax = static_cast<int16_t>(le_uint8_to_uint16(&buf[0]));
+  int16_t datay = static_cast<int16_t>(le_uint8_to_uint16(&buf[2]));
+  int16_t dataz = static_cast<int16_t>(le_uint8_to_uint16(&buf[4]));
+
+  //TODO: should we put timestamps here?
+  //
+  LIS2MDLReading reading = {
+      .ns = 0,
```
matt001k, line 86: @mbella-sofar here is where timestamps gathered from the LSM6DSV's FIFO could get injected into the sensor.
I think we should do this, but it would require a change in the `lsm6dsv.cpp` file and I want to keep this PR clean (would open another for this change).

mbella-sofar: This seems like a good approach to me. We should probably check with stanford on this still.

It looks like the [MSM Design and Specification Details notion page](https://app.notion.com/p/sofarocean/MSM-Design-and-Specification-Details-36e8ff95945080e6ad4edd48d5f3bfd7) we put together for the Stanford team lists microseconds since boot. Does that mean the mote uptime or the LSM6?

Also agreed that using a seperate PR for the lsm6dsv.cpp change is best.

mbella-sofar: Maybe we actually don't need to check with Stanford. Using the LSM6 timebase will be better than the mote timestamp overall in any case. So it seems like we should clarify the line in the above notion page instead.

mbella-sofar (approved):
Looks good to me!

src/lib/drivers/lis2mdl.cpp
matt001k: Well to note, if the sensor hub from the LSM polls the LIS2MDL at the same interval as the accelerometer/gyro rather than using the interrupt line to take readings, there would be more alignment with the reported timestamps.

In the document, this uptime from the LSM6's boot (note it does get reset in the init call, meaning it's uptick will be a little later than the mote's). I will update the doc to reflect.

matt001k: I will make that update in another PR then!

PR #398, opened 2026-07-13
Added BNO085 SPI driver for motion module

Davidbick:
## What changed?
Ported BNO085 IMU from I2C (devkit_imu_example) to SPI on the Mote board. Streams accelerometer, gyroscope, and magnetometer at 10 Hz via CEVA sh2 library. LSM6DSV path commented out but preserved for reference.

Previously the only BNO085 integration was I2C on the devkit. This enables the motion sensing module to use the BNO085 over the shared SPI bus.

LSM6DSV path is commented out, not deleted. This is intentional for review visibility; the final app will be separated after driver approval.

## Checklist


src/lib/sensor_sampler/imu_types.h
```diff
@@ -0,0 +1,19 @@
...
+// Neutral IMU types shared by all IMU drivers (LSM6DSV, BNO085).
+// Values are in each driver's native units; the compile-time-selected
+// driver fixes the units for a given build:
+//   LSM6DSV: acc = mg,    gyro = mdps
+//   BNO085:  acc = m/s^2,  gyro = rad/s
+typedef struct {
+  uint64_t ns;
+  struct { float x, y, z; } acc;
+  struct { float x, y, z; } gyro;
+} IMUReading;
```
matt001k, line 13: It might be nice to handle the compass in here too. I can see the `motionSampler` being updated to make this cleaner. I will have to think deeply about this one, not necessary for this PR.

src/lib/sensor_sampler/motionSampler.h
```diff
@@ -7,9 +7,10 @@
 #include "lsm6dsv.h"
 #include "protected_spi.h"
 #include "util.h"
+#include "imu_types.h"
 
-typedef LSM6DSV::LSM6DSVReading IMUReading;
-typedef LIS2MDL::LIS2MDLReading CompassReading;
+//typedef LSM6DSV::LSM6DSVReading IMUReading;
+//typedef LIS2MDL::LIS2MDLReading CompassReading;
```
matt001k: These can be removed.

src/lib/sensor_sampler/bno085Sampler.h
```diff
@@ -0,0 +1,38 @@
...
+    Q _imuQueue;
+    Q _magQueue;
+    uint8_t _imuBuf[(sizeof(QItem) + sizeof(IMUReading)) * IMU_QUEUE_COUNT];
+    uint8_t _magBuf[(sizeof(QItem) + sizeof(CompassReading)) * MAG_QUEUE_COUNT];
+    BmSemaphore _queueMut;   // guards both queues (driver task <-> app task)
+    BmSemaphore _imuSem;     // "IMU reading ready" -- mirrors LSM6DSV m_reading_sem
+
+    IMUReading _latch;       // accel + latest gyro, emitted on each accel report
+    bool _haveGyro;
+};
```
matt001k, line 44: Similarly to the `motionSampler` it might be nice to have a static function in here called `addBno085Sampler` that could invoke `init` directly there. Reference line 209 in `motionSampler` to see the task it creates along with the initialization done in that task.
Here the initialization could just be done in the call.

src/apps/motion_module/sensors.cpp
```diff
@@ -39,19 +52,20 @@ void sensorsInit(void) {
                       &sample_rate)) {
     cfg.sample_rate = static_cast<lsm6dsv_data_rate_t>(sample_rate);
   }
-  motion.set_cfg(cfg);
-  motion_sampler_add(&motion);
+  imu.set_cfg(cfg);
+  motion_sampler_add(&imu);
```
matt001k: Per my comment below, in the bno085Sampler, this would clean up this logic. It also would be nice to use similar configurations for the Bno085, even if not immediately applicable.

src/lib/sensor_sampler/bno085Sampler.h
Davidbick: Sounds good! I made that change in a following commit.

src/apps/motion_module/CMakeLists.txt
```diff
@@ -18,6 +24,19 @@ set(APP_DEFINES
...
+        ${SRC_DIR}/third_party/ceva_sh2/sh2.c
+        ${SRC_DIR}/third_party/ceva_sh2/shtp.c
+        ${SRC_DIR}/third_party/ceva_sh2/sh2_SensorValue.c
+        ${SRC_DIR}/third_party/ceva_sh2/sh2_util.c
+        ${SRC_DIR}/lib/drivers/bno085.cpp
+        ${SRC_DIR}/lib/sensor_sampler/bno085Sampler.cpp
+    )
+    list(APPEND APP_INCLUDES ${SRC_DIR}/third_party/ceva_sh2/)
+    list(APPEND APP_DEFINES IMU_BNO085=1)
+endif()
```
matt001k, line 29: The files should be able to be linked in depending on the IMU set.
We can build these files in:

- `src/third_party/CMakeLists.txt`
- `src/lib/CMakeLists.txt`

This statement could be kept here although:

```
    list(APPEND APP_DEFINES IMU_BNO085=1)
```

src/apps/motion_module/app_main.cpp
```diff
@@ -292,5 +292,6 @@ static void defaultTask(void *parameters) {
 
   while (1) {
     sensorsHandle();
+    vTaskDelay(pdMS_TO_TICKS(10));
```
matt001k: This should not be needed anymore. or should not be necessary in the first place, I will look to see how you are handling the incoming BNO data.

src/bsp/bm_mote_spi_v1_0/Core/Src/spi.c
```diff
@@ -51,8 +51,8 @@ void MX_SPI1_Init(void)
   hspi1.Init.Mode = SPI_MODE_MASTER;
   hspi1.Init.Direction = SPI_DIRECTION_2LINES;
   hspi1.Init.DataSize = SPI_DATASIZE_8BIT;
-  hspi1.Init.CLKPolarity = SPI_POLARITY_LOW;
-  hspi1.Init.CLKPhase = SPI_PHASE_1EDGE;
+  hspi1.Init.CLKPolarity = SPI_POLARITY_HIGH;
+  hspi1.Init.CLKPhase = SPI_PHASE_2EDGE;
```
matt001k: Is this correct? I do not think the LSM6 will work in this configuration.

src/lib/sensor_sampler/bno085Sampler.cpp
```diff
@@ -0,0 +1,126 @@
...
+    Bno085Sampler *self = _instance;
+    if (!self || !event) return;
+    // The driver re-enables its reports on reset; we just drop the stale gyro latch
+    // so a post-reset reading never pairs a fresh accel with a pre-reset gyro.
+    if (event->eventId == SH2_RESET) {
+        self->_haveGyro = false;
+    }
+}
+
+bool Bno085Sampler::data_ready(uint32_t timeout_ms) {
```
matt001k, line 92: Yea because of this line, the vTaskDelay should be removed in app_main.cpp. Maybe follow the pattern in the `motionSampler`file that adds a delay if there is no existing device..

matt001k (commented):
I think after the last few comments are addressed, this is ready to be merged.

src/bsp/bm_mote_spi_v1_0/Core/Src/spi.c
Davidbick: The LSM6DSV datasheet supports SPI mode 0 and mode 3. Tested on hardware, readings are correct. I used Mode 3 because BNO085 only supports mode 3, and both share the SPI bus.

matt001k: Oof ok. We might have to discuss/look into this more, because there are other applications that use this SPI line (the SOFT module for example).
It might have to be something we do in our application code rather than the BSP itself.

matt001k: Oof ok. We might have to discuss/look into this more, because there are other applications that use this SPI line (the SOFT module for example).
It might have to be something we do in our application code rather than the BSP itself.

matt001k: Oof ok. We might have to discuss/look into this more, because there are other applications that use this SPI line (the SOFT module for example).
It might have to be something we do in our application code rather than the BSP itself.

matt001k: Looking at the TSYS01 datasheet for the soft it does state it supports Mode3:

image

matt001k: But I am still a bit wary.

src/lib/drivers/bno085.cpp
matt001k, line 1: Let's open another PR into `develop` with the changes here.

src/lib/CMakeLists.txt
```diff
@@ -150,6 +150,13 @@ if(CMAKE_APP_TYPE STREQUAL "BMDK" OR
     endif()
 endif()
 
+if(IMU STREQUAL "BNO085")
+    list(APPEND LIB_FILES
+        ${LIB_DIR}/drivers/bno085.cpp
+        ${LIB_DIR}/sensor_sampler/bno085Sampler.cpp
+    )
+endif()
```
matt001k: Sorry for the nit-picky. But lets remove the `if(IMU STREQUAL "BNO085")`, it can just be built in with any application in general!

src/third_party/CMakeLists.txt
```diff
@@ -75,6 +75,16 @@ include(${SRC_DIR}/third_party/memfault.cmake)
 set(LWIP_COMPILER_FLAGS ${COMPILE_FLAGS})
 include(${LWIP_DIR}/src/Filelists.cmake)
 
+if(IMU STREQUAL "BNO085")
```
matt001k: Same here, can remove the if statement 😄

matt001k:
Cool, code looks good @Davidbick, let's turn this draft into the ready state and get it merged!

matt001k (approved):

PR #399, opened 2026-07-13
Added BNO085 SPI driver

Davidbick:
## What changed?
Added a SPI driver for the BNO085 IMU and the CEVA sh2 submodule it depends on. It supports configurable sensor reports (accelerometer, gyroscope, magnetometer).

Enables the BNO085 IMU on the Mote board over SPI. Previously the only BNO085 integration was I2C on the devkit. This is the reusable driver; app integration will come in follow-up PRs.

## Checklist

### DIFF
--- a/src/lib/drivers/bno085.cpp
+++ b/src/lib/drivers/bno085.cpp
@@ -16,16 +16,14 @@ Bno085::Bno085(SPIInterface_t *spi, IOPinHandle_t *csPin, IOPinHandle_t *intPin,
       _serviceTaskHandle(NULL),
       _eventCallback(NULL),
       _sensorCallback(NULL),
-      _cookie(nullptr),
       _initialized(false),
       _resetSeen(false) {
 }
 
 BmErr Bno085::init(sh2_EventCallback_t *eventCallback,
-                  sh2_SensorCallback_t *sensorCallback, void *cookie) {
+                  sh2_SensorCallback_t *sensorCallback) {
     _eventCallback = eventCallback;
     _sensorCallback = sensorCallback;
-    _cookie = cookie;
     printf("BNO085: Creating service task\n");
 
     return xTaskCreate(serviceTask, "BNO085", configMINIMAL_STACK_SIZE * 8,
@@ -92,7 +90,7 @@ void Bno085::serviceTask(void *arg) {
 
     // Set sensor callback
     if (instance->_sensorCallback) {
-        sh2_setSensorCallback(instance->_sensorCallback, instance->_cookie);
+        sh2_setSensorCallback(instance->_sensorCallback, instance);
     }
 
     instance->_initialized = true;
@@ -168,7 +166,7 @@ void Bno085::eventHandler(void *cookie, sh2_AsyncEvent_t *event) {
         self->_resetSeen = true;   // serviceTask re-applies sensor config
     }
     if (self->_eventCallback) {
-        self->_eventCallback(self->_cookie, event);   // forward to the app callback
+        self->_eventCallback(self, event);   // forward to the app callback
     }
 }
 
--- a/src/lib/drivers/bno085.cpp
+++ b/src/lib/drivers/bno085.cpp
@@ -5,9 +5,6 @@
 #include "string.h"
 #include "watchdog.h"
 
-// Initialize static instance pointer
-Bno085* Bno085::_instance = nullptr;
-
 Bno085::Bno085(SPIInterface_t *spi, IOPinHandle_t *csPin, IOPinHandle_t *intPin,
                IOPinHandle_t *rstPin, IOPinHandle_t *bootPin, IOPinHandle_t *wakePin)
     : _interface(spi),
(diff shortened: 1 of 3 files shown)


src/lib/drivers/bno085.cpp
```diff
@@ -0,0 +1,233 @@
...
+}
+
+bool Bno085::init(sh2_EventCallback_t *eventCallback,
+                  sh2_SensorCallback_t *sensorCallback) {
+    _eventCallback = eventCallback;
+    _sensorCallback = sensorCallback;
+    printf("BNO085: Creating service task\n");
+
+    return xTaskCreate(serviceTask, "BNO085", configMINIMAL_STACK_SIZE * 8,
+                       this, 21, &_serviceTaskHandle) == pdPASS;
```
victorsowa12: lets define the task priority in the header with something like:
```
#ifndef BNO085_TASK_PRIORITY
#define BNO085_TASK_PRIORITY 21
#endif 
```

This will reduce a magic number and allow someone to configure the priority in `task_priorities.h` if they want.

victorsowa12 (approved):
Looks good! Only had a small suggestion for the task priority

src/lib/drivers/bno085.cpp
Davidbick: Sounds good! I just made that change.

src/lib/drivers/bno085.cpp
```diff
@@ -0,0 +1,233 @@
...
+    IOWrite(_bootPin, 1);
+    IOWrite(_wakePin, 1);
+    IOWrite(_csPin, 1);
+    IOWrite(_rstPin, 0);
+    vTaskDelay(pdMS_TO_TICKS(10));
+    IOWrite(_rstPin, 1);
+    if (!waitForIntLow(pdMS_TO_TICKS(500))) {
+        printf("BNO085: reset - NO INT within 500ms\n");
+        return false;
+    }
```
matt001k: This is something guaranteed to happen whenever the sensor resets?
Is it worth checking if this was asserted at the beginning of this function and if so making sure that the sequence of `IOWrite` calls bring it up then back down?

src/lib/drivers/bno085.cpp
```diff
@@ -0,0 +1,233 @@
...
+    // Use the static instance pointer
+    if (!_instance) {
+        return 0;
+    }
+    return _instance->writeBytes(pBuffer, len);
+}
+
+uint32_t Bno085::getTimeUs(sh2_Hal_t *self) {
+    (void)self;
+    return uptimeGetMs() * 1000;
```
matt001k: YOu can use `uptimeGetMicroSeconds` here.

src/lib/drivers/bno085.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+        if (instance->_resetSeen) {
+            instance->_resetSeen = false;
+            printf("BNO085: reset detected - re-enabling sensors\n");
+            instance->enableSensors();
+        }
+    }
+}
+
+bool Bno085::configureSensor(sh2_SensorId_t sensorId, uint32_t reportInterval_us) {
```
matt001k: Again might be good to return `BmErr` here on public facing API.

src/lib/drivers/bno085.h
```diff
@@ -0,0 +1,62 @@
...
+              sh2_SensorCallback_t *sensorCallback);
+
+    bool configureSensor(sh2_SensorId_t sensorId, uint32_t reportInterval_us);
+
+    // HAL functions (called by sh2 library)
+    static int open(sh2_Hal_t *self);
+    static void close(sh2_Hal_t *self);
+    static int read(sh2_Hal_t *self, uint8_t *pBuffer, unsigned len, uint32_t *t_us);
+    static int write(sh2_Hal_t *self, uint8_t *pBuffer, unsigned len);
+    static uint32_t getTimeUs(sh2_Hal_t *self);
```
matt001k, line 34: Do these have to be public?

src/lib/drivers/bno085.h
```diff
@@ -0,0 +1,62 @@
...
+    TaskHandle_t _serviceTaskHandle;
+
+    sh2_EventCallback_t *_eventCallback;
+    sh2_SensorCallback_t *_sensorCallback;
+
+    bool _initialized;
+    bool _resetSeen;
+
+    // Static instance pointer for HAL callbacks
+    static Bno085* _instance;
```
matt001k: Does this make the class singleton? Is there any reason that has to be static?

src/lib/drivers/bno085.h
```diff
@@ -0,0 +1,62 @@
...
+class Bno085 {
+public:
+    // SPI bus + chip-select + INT + the three control pins (RST/BOOTN/WAKE)
+    Bno085(SPIInterface_t *spi, IOPinHandle_t *csPin, IOPinHandle_t *intPin,
+           IOPinHandle_t *rstPin, IOPinHandle_t *bootPin, IOPinHandle_t *wakePin);
+
+    bool init(sh2_EventCallback_t *eventCallback,
+              sh2_SensorCallback_t *sensorCallback);
+
+    bool configureSensor(sh2_SensorId_t sensorId, uint32_t reportInterval_us);
```
matt001k: For public API lets try and return `BmErr` to make this easier to be compatible with the LSM6DSV driver and other potential for abstractions with drivers in the future.

src/lib/drivers/bno085.h
Davidbick: No, they're only referenced when building the sh2_Hal_t inside serviceTask, so I will move them to private!

src/lib/drivers/bno085.h
Davidbick: Yes, it's effectively a singleton right now but it shouldn't be. I will move the sh2_Hal_t into the class as the first member and recover the instance from self via container-of in the callbacks. That drops the static and the single-instance assert.

src/lib/drivers/bno085.cpp
Davidbick: Good question. I changed it so the helper now takes the level, so we can check both directions: waitForInt(1, …) for deasserted and waitForInt(0, …) for asserted. That way we can be sure the `IOWrite` calls funnction as they should.

matt001k (approved):
Nice! Looks clean

src/lib/drivers/bno085.cpp
```diff
@@ -0,0 +1,222 @@
...
+      _serviceTaskHandle(NULL),
+      _eventCallback(NULL),
+      _sensorCallback(NULL),
+      _cookie(nullptr),
+      _initialized(false),
+      _resetSeen(false) {
+}
+
+BmErr Bno085::init(sh2_EventCallback_t *eventCallback,
+                  sh2_SensorCallback_t *sensorCallback, void *cookie) {
```
matt001k: I think if `this` is going to be referenced for every implementation of this driver's cookie, we can just remove this variable and use `this` in every reference of the cookie.

matt001k:
Cool this looks good to squash and merge now!

PR #400, opened 2026-07-14
Update LSM6DSV With Some Quality Of Life Improvements

matt001k:
## What changed?
Improved configuration order handling (sensor hub) Default to polled mode with sensor hub devices
Improve function comments
Use 1 data rate to sample from the input configuration
Additional protection against NULL semaphore/mutices

Reduces the number of configurations needed for the LSM6DSV.
Improves robustness for sampling checks.

@mbella-sofar this PR addresses the changes made to the code that utilizes 1 unified polling rate to sample all the gyro, acc, and mag.

## Checklist

### DIFF
--- a/src/lib/drivers/lsm6dsv.cpp
+++ b/src/lib/drivers/lsm6dsv.cpp
@@ -263,10 +263,6 @@ BmErr LSM6DSV::start_stream(std::span<LSM6DSVSensorHub> sensor_hub_items, size_t
 }
 
 lsm6dsv_sh_data_rate_t LSM6DSV::sensor_hub_data_rate_convert(lsm6dsv_data_rate_t data_rate) {
-  if (data_rate <= LSM6DSV_ODR_AT_15Hz) {
-    return LSM6DSV_SH_15Hz;
-  }
-
   lsm6dsv_sh_data_rate_t sh_data_rate = LSM6DSV_SH_15Hz;
   switch (data_rate) {
   case LSM6DSV_ODR_AT_30Hz:
--- a/src/lib/drivers/lsm6dsv.cpp
+++ b/src/lib/drivers/lsm6dsv.cpp
@@ -54,8 +54,8 @@ BmErr LSM6DSV::init(const Cfg *cfg) {
   }
 
   // Do not support high accuracy data rates ref: 6.5 DS13476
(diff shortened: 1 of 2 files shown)


src/lib/drivers/lsm6dsv.cpp
```diff
@@ -191,13 +192,24 @@ BmErr LSM6DSV::start_stream(std::span<LSM6DSVSensorHub> sensor_hub_items, size_t
...
+    lsm6dsv_sh_data_rate_t sh_data_rate = sensor_hub_data_rate_convert(m_cfg.sample_rate);
+    lsm6dsv_return_on_err(lsm6dsv_sh_data_rate_set(&m_ctx, sh_data_rate));
 
     // This must be set to enable reading from device 0
     lsm6dsv_return_on_err(lsm6dsv_sh_write_mode_set(&m_ctx, LSM6DSV_ONLY_FIRST_CYCLE));
 
     lsm6dsv_return_on_err(lsm6dsv_sh_slave_connected_set(
         &m_ctx, static_cast<lsm6dsv_sh_slave_connected_t>(sensor_hub_count - 1)));
+
+    // Configure sensor hub to trigger from interrupt INT2, by default this is active low
```
victorsowa12, line 204: there is a comment on line 116 that says the int pin is configured as active high and here we are configuring and commenting that it is active low. I see this pin is called INT2, are there actually 2 INT pins and are they different polarities?

victorsowa12 (approved):
Changes look good to me! Only had a question about code comments regarding the active high/low of the interrupt.

src/lib/drivers/lsm6dsv.cpp
matt001k: Yes, there is an input interrupt to the LSM6DSV which can be used to sample external sensors on an I2C bus (called Sensor Hub in ST's lingo), this is called INT2.
INT1 is the interrupt hooked up to the U5, this one is active high.

Let me know if it is worthwhile to clear that up on line 116!

matt001k: Added an indication on line 116 that it is in reference to INT1!

victorsowa12: Nice! Thanks for the clarification!

mbella-sofar (approved):
Looks good to me!

PR #403, opened 2026-07-20
Metrics service MVP

Davidbick:
## What changed?
Adds a network metrics sampler to the bridge. A new background task (metrics_sampler.cpp) polls every discovered node's metrics service on a timer, and appends each CBOR reply to network_metrics.log via spotter_log(). The bridge polls its own metrics service too.

Depends on: [bm_common_messages PR](https://github.com/bristlemouth/bm_common_messages/pull/65) and [bm_core PR](https://github.com/bristlemouth/bm_core/pull/150).

This gives us observability into the mesh's physical link health. Each node's ADIN2111 PHY exposes signal quality (MSE, SQI, link-quality) and error counters (RX CRC, symbol errors) that were previously unreadable off-device. Now the bridge collects them fleet-wide into one timestamped log, so we can spot a degrading cable or a noisy port over time instead of after a link drops.

`metrics_sampler.cpp` is the substance. 

## Checklist


src/apps/bridge/metrics_sampler.cpp
```diff
@@ -0,0 +1,97 @@
+#include "metrics_sampler.h"
+
+#include "FreeRTOS.h"
+#include "task.h"
```
matt001k: Lets use `bm_os` calls here to try and reduce as much FreeRTOS to bristlemouth OS noise as possible.

src/apps/bridge/app_main.cpp
```diff
@@ -441,9 +443,11 @@ static void defaultTask(void *parameters) {
   sensorControllerInit(&bridge_power_controller);
   ncpInit(&usart3, &dfu_partition, &bridge_power_controller);
   topology_sampler_init(&bridge_power_controller);
+  metrics_sampler_init();
   debug_ncp_init();
   debugBmServiceInit();
   sys_info_service_init();
+  metrics_service_init();
```
matt001k: Maybe the metrics service can be a default service hosted on every Bristlemouth device. Thoughts? @Davidbick @victorsowa12

src/apps/bm_soft_module/app_main.cpp
```diff
@@ -406,6 +408,8 @@ static void defaultTask(void *parameters) {
   bm_sub(APP_PUB_SUB_UTC_TOPIC, handle_bm_subscriptions);
   echo_service_init();
   sys_info_service_init();
+  metrics_service_init();
+  metrics_network_component_init();
```
matt001k: I might be a bigger fan of this being initialized in `bm_core`. I think it should be something that is configurable in the `bm_config.h` file. This might look something like

```

#define METRICS_SERVICE_ENABLE 0

```

And then in `bristlemouth_init` there would be a line as follows:

```
...
#if METRICS_SERVICE_ENABLE
metrics_service_init();
#endif
...
```

The same would go in the adin initialization function for the network component:

```
#if METRICS_SERVICE_ENABLE
adin_network_metrics_init();
#endif
```

This will also reduce a bunch of code in every `app_main.cpp`.

src/apps/bridge/metrics_sampler.cpp
```diff
@@ -0,0 +1,109 @@
...
+  // Raw CBOR travels on the wire; base64 is applied only here
+  size_t b64_len = 0;
+  mbedtls_base64_encode(NULL, 0, &b64_len, reply_data, reply_len);
+  if (b64_len == 0) {
+    bridgeLogPrint(BRIDGE_SYS, BM_COMMON_LOG_LEVEL_WARNING, USE_HEADER,
+                   "Empty metrics reply from %.*s\n", (int)service_strlen, service);
+    return true;
+  }
+
+  uint8_t *b64 = (uint8_t *)bm_malloc(b64_len);
```
matt001k: Can this be a static buffer?
Because we know this won't be larger than 1 MTU of Bristlemouth data, can we calculate the max size of of Base64'd data on 1500 bytes of Bristlemouth data (or 1000 bytes whatever is truly being used for services).
This way we do not have to continuously `malloc` buffers here, saving time and reducing potential fragmentation in the heap.
Also note that this `cb` is called from a single thread, so no need to `mutex` protect the variable.

src/apps/bridge/metrics_sampler.cpp
```diff
@@ -0,0 +1,109 @@
...
+                   (int)service_strlen, service);
+  }
+  bm_free(b64);
+  return true;
+}
+
+static void metrics_sampler_task(void *parameters) {
+  (void)parameters;
+  for (;;) {
+    vTaskDelay(pdMS_TO_TICKS(_poll_interval_ms));
```
matt001k: Lets try as much as we can to use the `bm_os.h` function calls (`bm_delay` here) to standardize abstractions in our calls!

src/apps/bridge/metrics_sampler.cpp
```diff
@@ -0,0 +1,109 @@
...
+                   _poll_interval_ms);
+    set_config_uint(BM_CFG_PARTITION_SYSTEM, AppConfig::METRICS_POLL_INTERVAL_MS,
+                    strlen(AppConfig::METRICS_POLL_INTERVAL_MS), _poll_interval_ms);
+    save_config(BM_CFG_PARTITION_SYSTEM, false);
+  }
+
+  BaseType_t rval =
+      xTaskCreate(metrics_sampler_task, "METRICS_SAMPLER", METRICS_SAMPLER_TASK_STACK_SIZE,
+                  NULL, METRICS_SAMPLER_TASK_PRIORITY, NULL);
+  configASSERT(rval == pdPASS);
```
matt001k: I think we should only create this task if `metricsPollIntervalMs` is not zero. Meaning we will only poll from the network if there is a non-zero configuration set. The default should also be zero.

src/apps/bridge/app_config.h
```diff
@@ -24,6 +24,7 @@ constexpr const char *RBR_CODA_READING_PERIOD_MS = "rbrCodaReadingPeriodMs";
 constexpr const char *TICKS_SAMPLING_ENABLED = "ticksSamplingEnabled";
 constexpr const char *TURBIDITY_READING_PERIOD_MS = "turbidityReadingPeriodMs";
 constexpr const char *AANDERAA_SALINITY_READING_PERIOD_MS = "salinityReadingPeriodMs";
+constexpr const char *METRICS_POLL_INTERVAL_MS = "metricsPollIntervalMs";
```
matt001k: Lets make this `seconds` instead of `milliseconds`. I think `ms` is too granular and could create chaos in a system in the value was set really low like `10ms`, flooding the network with messages.

tools/scripts/misc/decode_metrics.py
```diff
@@ -0,0 +1,77 @@
...
+            m = LINE_RE.match(line)
+            if not m:
+                continue
+            try:
+                env = cbor2.loads(base64.b64decode(m.group("b64")))
+            except Exception:
+                continue  # skip malformed lines
+            net = env.get("data", {}).get("network_port_stats")
+            if not net:
+                continue  # only network stats for now; other components ignored
```
matt001k: Maybe place a print statement here to indicate the data type is not supported.
This way folks can know they should update this script if they add new components.

matt001k (commented):
Nice work! A couple of comments here, let me know your thoughts!

src/apps/bridge/app_main.cpp
Davidbick: Great idea! This is now implemented in bm_core.

src/apps/bm_soft_module/app_main.cpp
Davidbick: I moved the initialization to bm_core and cleaned up the app code!

src/apps/bridge/metrics_sampler.cpp
Davidbick: Yes, it can be static. That cleans up the sizing calls and removes the malloc!

tools/scripts/misc/decode_metrics.py
Davidbick: Instead of warning per unknown component, I will make the decoder generic so new types are captured with no code changes. ADIN keeps its port/lq_label enrichment as a special case.

matt001k: Nice! If it is generic enough, maybe one day we can place it in `bm_core` to be used!

matt001k (approved):
Looks clean 😮‍💨 LGTM!

PR #404, opened 2026-07-20
Add Keller Depth Sensor Driver For XLD Sensors

matt001k:
## What changed?
Adds driver for Keller XLD sensors.

This prepares a driver necessary for the Motion Depth Module.

I tried to write this to match the same flow that the LSM6DSV driver has.
Let me know your thoughts, or if I need to explain in detail any other information!

## Checklist

### DIFF
--- a/src/lib/drivers/xld.cpp
+++ b/src/lib/drivers/xld.cpp
@@ -1,4 +1,6 @@
 #include "xld.h"
+#include "FreeRTOS.h"
+#include "semphr.h"
 #include <string.h>
 
 XLD::XLD(SensorInterfaceBus *bus, void *arg) : m_bus(bus), m_arg(arg) {}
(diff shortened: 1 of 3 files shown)


src/lib/drivers/xld.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+/*!
+ @brief Handle EOC interrupt
+
+ @details Must be called in the interrupt handler tied to the EOC pin.
+          Indicates when a reading has been performed. Pin is active low.
+ */
+void XLD::handle_interrupt(void) {
+  if (m_data_ready_sem) {
+    bm_semaphore_give(m_data_ready_sem);
```
victorsowa12: should this be `bm_semaphore_give_from_isr`? I'm also not sure if we have that api.

victorsowa12 (approved):
Looks good!

src/lib/drivers/xld.cpp
matt001k: Ah yes good call out. I will use the direct FreeRTOS call.

matt001k: Updated!

PR #405, opened 2026-07-21
BNO085 driver: configure SPI mode 3 and sample rate

Davidbick:
## What changed?
SPI mode is now set by the driver itself: configureSpiMode() runs at startup and puts the bus in mode 3, which the BNO085 requires. 
Report rate is now configurable: added setReportInterval() and a _reportInterval_us member.

The driver is now more self-contained as it no longer depends on a board-wide SPI-mode change to work. Consumers can now pick a sample rate, making the driver reusable across applications.

## Checklist


victorsowa12 (approved):

matt001k (approved):
Awesome job!

PR #408, opened 2026-07-29
Feature/network metrics cli

Davidbick:
## What changed?

- New `src/lib/bm_ncp/ncp_metrics.{cpp,h}`: requests metrics from the target node (or all nodes via topology when `node_id == 0`), pretty-prints each CBOR reply, and returns it to the console via `bm_serial_send_metrics_reply()`.

- Completes the on-demand `bm_metrics` path: the bridge serves host requests and returns readable per-node metrics, rendering on the bridge so the host just prints.

- The PR relies on [bm_serial PR 33](https://github.com/bristlemouth/bm_serial/pull/33) which should be merged first.

## Checklist


victorsowa12 (approved):
Lookin good to me!

matt001k (approved):

PR #413, opened 2026-08-04
Add Sensor Sampler To Sample Keller Sensor

matt001k:
## What changed?
Adds sampler to read from Keller XLD sensor devices.

Allows applications to run a task that handles reading from Keller sensors.

I am not leveraging the `sensorSampler` here, but rather creating my own separate task.
This is mainly to avoid inducing latency when accessing other sensors as each read in the `sensorSampler` task is blocking and it takes up to 8ms to read from the Keller sensor.
Let me know your thoughts on this approach.

## Checklist

### DIFF
--- a/src/lib/drivers/protected/protected_i2c.h
+++ b/src/lib/drivers/protected/protected_i2c.h
@@ -31,8 +31,8 @@ typedef struct {
 
 bool i2cInit(I2CInterface_t *interface);
 I2CResponse_t i2cTxRx(I2CInterface_t *interface, uint8_t address, uint8_t *txBuff, size_t txLen, uint8_t *rxBuff, size_t rxLen, uint32_t timeoutMs);
-#define i2cTx(interface, address, buff, len, timeout) i2cTxRx(interface, address, buff, len, NULL, 0, timeout);
-#define i2cRx(interface, address, buff, len, timeout) i2cTxRx(interface, address, NULL, 0, buff, len, timeout);
+#define i2cTx(interface, address, buff, len, timeout) i2cTxRx(interface, address, buff, len, NULL, 0, timeout)
+#define i2cRx(interface, address, buff, len, timeout) i2cTxRx(interface, address, NULL, 0, buff, len, timeout)
 I2CResponse_t i2cProbe(I2CInterface_t *interface, uint8_t address, uint32_t timeoutMs);
 void i2cLoadLogCfg();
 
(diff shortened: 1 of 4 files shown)


src/lib/sensor_sampler/kellerSampler.cpp
```diff
@@ -0,0 +1,138 @@
...
+ */
+BmErr keller_sampler_add(KellerSampler *sampler) {
+  constexpr uint8_t max_sample_rate = 125;
+  if (!sampler || sampler->m_cfg.sample_rate > max_sample_rate) {
+    return BmEINVAL;
+  }
+
+  IORegisterCallback(sampler->m_ctx.isr, xld_isr_handle, sampler);
+
+  static constexpr uint32_t stack_size = 512;
```
victorsowa12, line 136: does it need a stack size this large?

src/lib/drivers/protected/protected_i2c.h
```diff
@@ -31,8 +31,8 @@ typedef struct {
 
 bool i2cInit(I2CInterface_t *interface);
 I2CResponse_t i2cTxRx(I2CInterface_t *interface, uint8_t address, uint8_t *txBuff, size_t txLen, uint8_t *rxBuff, size_t rxLen, uint32_t timeoutMs);
-#define i2cTx(interface, address, buff, len, timeout) i2cTxRx(interface, address, buff, len, NULL, 0, timeout);
-#define i2cRx(interface, address, buff, len, timeout) i2cTxRx(interface, address, NULL, 0, buff, len, timeout);
+#define i2cTx(interface, address, buff, len, timeout) i2cTxRx(interface, address, buff, len, NULL, 0, timeout)
+#define i2cRx(interface, address, buff, len, timeout) i2cTxRx(interface, address, NULL, 0, buff, len, timeout)
```
victorsowa12, line 35: curious, this doesn't impact other usages of these defines?

victorsowa12 (approved):
Looks good to me! Only had a couple minor questions.

src/lib/sensor_sampler/kellerSampler.cpp
matt001k: Nope! This seems to be more than sufficient for the task!

src/lib/drivers/protected/protected_i2c.h
matt001k: No, all other usage of this macro adds an extra semicolon at the end.
I wrapped the i2cTx and i2cRx calls in a `static_cast<BmErr>()` which caused the build to break due to the already added semicolon in this macro.
You can have multiple semicolons finishing a statement, but you cannot have a semicolon before other characters in a statement sadly.
This just makes the macros a bit more versatile throughout the codebase!

victorsowa12: Awesome!

PR #414, opened 2026-08-04
Add Unbuffered average samplers 

Davidbick:
## What changed?

Replaced the bridge's buffered `AveragingSampler` (per-node malloc'd buffer) with bufferless streaming samplers on every bridge sensor:
- `AveragingSamplerUnbuffered` (running mean) - soft, seapoint, PME oxygen
- `AveragingSamplerStats` (Welford mean/std + min/max) - rbrCoda, aanderaaConductivity, aanderaa
- `AveragingSamplerCircular` (sin/cos sums) - aanderaa direction

Also removed the now-dead buffer-sizing code (averager_max_samples, per-sensor N_SAMPLES_PAD, and stale avgSampler.h includes) and added unit tests for all three classes.

Per-node memory drops from ~7 KB to ~20 bytes, fixing the heap-exhaustion crash at 7 nodes and unblocking 20+ node networks.

## Testing
- Unit tests: one suite per class, reusing the buffered class's datasets and numpy-verified expected values, so they double as bit-for-bit equivalence checks vs the old code.

### Overnight Bay Soak
**Goal**

- Validate the streaming (unbuffered) samplers against the existing buffered code on live hardware over a long run.
- All three new classes run live: Stats (current speed mean/std), Circular (current direction), Unbuffered/Stats (water temp)

**Setup**

- **Soak unit:** `SPOT-SUNFLOWER17` with unbuffered bridge and v2.16.8 spotter.
    - Sensors: 1× Aanderaa, 1× RBR
- **Baseline:** `SPOT-31299C` with existing buffered firmware, already in the bay.
    - Sensors: aanderaa, aanderaa_salinity, pme_do_sensor, pro_oceanus_minico2.
- Both deployed in the bay overnight (11 Aug to 12 Aug).

**Results** 

- **Subsurface water temperature**
    
Screenshot 2026-08-12 at 9 51 40 AM

    
    - Avg: 17.63 °C (soak) vs 17.7 °C (baseline)
    - Max: 18.65 vs 18.67
    - Min: 16.91 vs 16.93
    - Curves track each other closely across the full window.
- **Currents speed & direction**
    
Screenshot 2026-08-12 at 9 51 59 AM

    
    - Avg direction: 17° vs 13°.
    - Avg speed: 16.6 cm/s vs 12.8 cm/s; Max 34.96 vs 18.68; Min 6.52 vs 2.65.
    - Trends/shape track. Absolute speed differs more, as expected, 
    since the two Spotters are at different points in the bay and currents will vary.

**Conclusion**

- New-code aggregates are physically sensible and track a known-good baseline overnight.
- Temp is near-identical; currents are consistent given location differences.

- Streaming stats match buffered semantics (variance ÷ n, circular mean in [0, 2π)).
- Each field uses the right sampler for the stats it reports.

**Follow-up:** `avgSampler` can be fully removed once the node/`user_code` apps (aanderaa, devkit_monitor, rbr_coda_example) migrate. They only need mean/std/min/max, which `AveragingSamplerStats` now covers.

## Checklist


src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
```diff
@@ -26,9 +26,7 @@ AanderaaConductivitySensor::AanderaaConductivitySensor(uint64_t conductivity_nod
                                                        uint32_t agg_period_ms,
                                                        uint32_t averager_max_samples) {
 
-  for (auto &sampler : samplers.data) {
-    sampler.initBuffer(averager_max_samples);
-  }
+  (void)averager_max_samples;
```
matt001k: Can this variable be removed?

src/apps/bridge/sensor_drivers/aanderaaSensor.cpp
```diff
@@ -228,11 +227,7 @@ Aanderaa_t *createAanderaaSub(uint64_t node_id, uint32_t current_agg_period_ms,
   new_sub->type = SENSOR_TYPE_AANDERAA;
   new_sub->next = NULL;
   new_sub->current_agg_period_ms = current_agg_period_ms;
-  new_sub->abs_speed_cm_s.initBuffer(averager_max_samples);
-  new_sub->direction_rad.initBuffer(averager_max_samples);
-  new_sub->temp_deg_c.initBuffer(averager_max_samples);
-  new_sub->abs_tilt_rad.initBuffer(averager_max_samples);
-  new_sub->std_tilt_rad.initBuffer(averager_max_samples);
+  (void)averager_max_samples;
```
matt001k: Can this variable be removed?

src/lib/common/avgSamplerStats.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include <math.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+//Bufferless mean + population variance/std plus running min/max. O(1) memory, single pass.
```
matt001k: It might be nice to provide a brief comment in here describing Welford's algorithm here.

src/lib/common/avgSamplerCircular.h
matt001k, line 1: Prior to reading this I initially was under the guise it was using a _circular buffer_ and not necessary performing a circular mean/std. A brief explanation describing what this is calculating would be very beneficial.

matt001k (commented):
The code looks good! A few comments to clean up. I really think it would be good to remove all of the unused variables rather than placing `(void)` in front of them IF and only IF it does not add a bunch of excess code.

Other than unit tests I would like to see a real world comparison of the old averaging techniques to this new one! It would help provide some evidence that can be tied to this PR when folks want to review it in the future.

src/apps/bridge/sensor_drivers/aanderaaConductivitySensor.cpp
Davidbick: Yes! I went through and removed all `averager_max_samples`, its computations and `N_SAMPLES_PAD` which should all be stale after the unbuffered averager switch.

matt001k:
@Davidbick, that link in association with the soak is not viewable by folks who are not from Sofar Ocean. Would you please place the results from your testing here in this PR so anyone inquiring about this feature can see your test results.

victorsowa12 (approved):
The code changes look good and appreciate all of the testing with hardware (second to Matt's comment, posting results here will make sure everyone in the open source repo can see them!). I appreciate the unit tests matching the pre-existing ones! That helps validate that the new implementation does not stray from the old one. I wonder if these bufferless implementations will help reduce code/the need for getting/setting sample cadences for each sensor! Something that can be explored in the future.

PR #415, opened 2026-08-05
Reserve NUL byte in network config log chunk buffer

Davidbick:
## What changed?

`print_stream_handler` (network_config_logger.cpp) now reserves a byte for the NUL terminator in its chunk-flush test:

`vsnprintf` reserves a byte for the NUL, so the old code dropped one character at every 1024-byte boundary of the `Bridge network config:` log line.

The branch guarantees `space_left >= n + 1`, so every fragment is written in full and `state->pos` stays correct.


matt001k (approved):
Nice find!

victorsowa12 (approved):

PR #417, opened 2026-08-07
Use canonical metrics log decoder from bm_common_messages

Davidbick:
## What changed?
- tools/scripts/misc/decode_metrics.py drops its own decode loop and imports the shared decoder; keeps only its pandas/CSV + ADIN-label layer.

- Removes a duplicate decoder and the debug tool now can't misalign with the schema.

- This PR depends on [bm_core bump](https://github.com/bristlemouth/bm_core/pull/153) and the core decoder [PR for bm_common_messages](https://github.com/bristlemouth/bm_common_messages/pull/66)

## Checklist

### DIFF
--- a/tools/scripts/misc/decode_metrics.py
+++ b/tools/scripts/misc/decode_metrics.py
@@ -1,79 +1,59 @@
+"""
+Decode a Bristlemouth network_metrics.log to CSV for inspection.
+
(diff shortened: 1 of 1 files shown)


matt001k (approved):

victorsowa12 (approved):

Davidbick:
I added the change that makes each component into its own CSV for readability.

PR #419, opened 2026-08-15
Feature/ftp

victorsowa12:
## Checklist


src/lib/common/nvmPartition.cpp
```diff
@@ -32,6 +32,6 @@ uint32_t NvmPartition::alignment(void) {
 }
 
 bool NvmPartition::crc16(uint32_t offset, size_t len, uint16_t &crc, uint32_t timeoutMs) {
-    configASSERT(offset + len + (len % _storage_driver.getAlignmentBytes()) < _partition.fa_size);
+    configASSERT(offset + len < _partition.fa_size);
```
victorsowa12, line 35: not sure if this is an actual bug, but it prevented 2560 from being read/written to in the nvm.

